PR #14, opened 2017-02-23
Split AOI polygon into squares

LindaAlblas:


client/app/admin/create-project/create-project.controller.js
```diff
@@ -30,7 +30,7 @@
          */
         vm.setStep = function(step){
```
hunt3ri: Add comment to define what step is, and why we need to define it.

client/app/services/project.service.js
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        /**
+         * Returns a task feature with a polygon geometry defining the task area
+         * @param step
+         * @param x
+         * @param y
+         * @returns {ol.Feature}
+         * @private
+         */
```
hunt3ri: Define what each param is in the comment

client/app/services/project.service.js
```diff
@@ -0,0 +1,108 @@
...
+        return service;
+
+        /**
+         * Creates a task grid with features for a polygon feature.
+         * It snaps to the OSM grid
+         * @param areaOfInterest (ol.Feature) - this should be a polygon
+         * @param zoom
+         * @returns {Array} of Features
+         */
+        function getTaskGrid(areaOfInterest, zoom) {
```
hunt3ri: As discussed, we really should have some test here to ensure no-one breaks this algo, so that would be v. BAD

PR #20, opened 2017-02-28
Ht 21 task size

LindaAlblas:
Main feature:
- added ability to make the task sizes smaller or larger and give the user warning message if going above number of task threshold or task size threshold

Also:
- added scale line control
- rename variables for calculating grid to make it easier to read
- store task grid geometry in project service


client/app/admin/create-project/create-project.html
```diff
@@ -57,63 +59,67 @@ <h1 class="section__title">
...
+                                    ng-click="createProjectCtrl.changeSizeTaskGrid(1)">Smaller
+                            </button>
                         </div>
-                    </form>
+                        <hr/>
+                        <p>A new project will be created with {{ createProjectCtrl.numberOfTasks }} tasks.</p>
+                        <p>The size of each task is roughly {{ createProjectCtrl.sizeOfTasks | number : 2 }} km<sup>2</sup>.</p>
+                        <div ng-show="createProjectCtrl.numberOfTasks > createProjectCtrl.maxNumberOfTasks"
+                             class="alert alert--warning" role="alert">
+                          <p><strong>Warning:</strong> You can't create this project with smaller tasks.
```
feenster: Suggests reviewing this message, eg "The tasks sizes cannot be made smaller.  If smaller task sizes are require, consider splitting nthe project into smaller projects using smaller AOI's"

client/app/admin/create-project/create-project.controller.js
```diff
@@ -11,9 +11,27 @@
 
     function createProjectController(mapService, drawService, projectService) {
         var vm = this;
+
+        // Wizard variables
         vm.currentStep = '';
+        vm.taskGrid = false;
```
hunt3ri: I'd usually prefix bools with is or are, eg

vm.isTaskGrid
vm.isTaskArbitrary

client/app/admin/create-project/create-project.controller.js
```diff
@@ -11,9 +11,27 @@
...
+        vm.taskArbitrary = false;
+
+        // AOI variables
         vm.AOIValid = true;
         vm.AOIValidationMessage = '';
+        vm.AOI = null;
+
+        // Creating grid variables
+        vm.sizeOfTasks = 0;
```
hunt3ri: Should these variables be Constants (this might not make sense in Angular land) so ignore if so.

client/app/admin/create-project/create-project.controller.js
```diff
@@ -92,18 +131,35 @@
...
-            var zoomLevel = mapService.getOSMMap().getView().getZoom();
+            
+            vm.taskGrid = true;
+            
+            // Remove existing task grid
+            projectService.removeTaskGrid();
 
              // Get the AOI
-            var areaOfInterest = drawService.getFeatures();
+            var areaOfInterest = vm.AOI;
```
hunt3ri: Why map to a local var, and not just use the class level one?

client/app/admin/create-project/create-project.html
hunt3ri: I might suggest a message saying something like:

Maximum task limit breached.  If smaller tasks required please make AOI smaller.

client/app/admin/create-project/create-project.controller.js
LindaAlblas: I was thinking too much ahead (when we'll also have to handle uploaded AOIs). Changed it back to getting it from the service for now.

client/app/admin/create-project/create-project.html
LindaAlblas: " Maximum number of tasks reached. If smaller tasks are required please make the area of interest smaller."

client/app/admin/create-project/create-project.controller.js
LindaAlblas: Agree. Done.

client/app/admin/create-project/create-project.controller.js
LindaAlblas: constants do exist in modern JS land but it is not supported by all browsers. So I've used the convention to put them in uppercase.

client/app/admin/create-project/create-project.html
feenster: Sounds good to me

client/app/admin/create-project/create-project.controller.js
feenster: If we can solve how to do constants in anglar land, I would use them for the AOI validation fail reasons

LindaAlblas:
Made all the requested changes.

PR #25, opened 2017-03-02
Ht 24 split tasks

LindaAlblas:
Split specific area into smaller tasks.

Also:
- AOI and task grid are now MultiPolygon instead of Polygon
- AOI validation support for MultiPolygon + tests added

### DIFF
--- a/client/app/admin/create-project/create-project.controller.js
+++ b/client/app/admin/create-project/create-project.controller.js
@@ -12,17 +12,15 @@
     function createProjectController($scope, mapService, drawService, projectService) {
         var vm = this;
 
-        // Wizard variables
+        // Wizard 
         vm.currentStep = '';
         vm.isTaskGrid = false;
         vm.isTaskArbitrary = false;
 
-        // AOI variables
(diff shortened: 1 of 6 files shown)


client/app/admin/create-project/create-project.controller.js
```diff
@@ -7,22 +7,20 @@
...
 
-        // AOI variables
-        vm.AOIValid = true;
-        vm.AOIValidationMessage = '';
+        // AOI 
         vm.AOI = null;
 
-        // Grid variables
+        // Grid v
         vm.sizeOfTasks = 0;
```
hunt3ri: Typo

client/app/services/project.service.js
```diff
@@ -224,20 +252,112 @@
             }
 
             // check for self-intersections
+            for (var j = 0; j < features.length; j++) {
```
hunt3ri: j and k counters would be better named something more readable:

taskCount, polyCount or whatever

LindaAlblas:
@hunt3ri - all requested changes done

PR #42, opened 2017-03-09
Add Task Lock APIs

hunt3ri:


bgirardot:
Just double checking: Usually we do not allow a user to lock more than one task at a time, however, in tm3 a major goal is to allow validators to lock/checkout more than one task at time for validation. We need to make sure these two scenarios are possible and where/how we use them. I can not ever see mapping checking out more than one task at a time, however, validation it is very important.

hunt3ri:
Thanks for the notes @bgirardot.  We currently don't have the user models in the db, but I'll add a TODO re making sure users, can't lock more than one task.

Re validating multiple tasks - one issue that we haven't given thought to is navigating to those tasks in the editor.  eg if you select 2 tasks at either side of an AOI how can you easily view them??

One solution would be to select tasks for validation by user, so you could check one or two manually and if they looked good, then mark them all as valid.  WDYT?

Nick-Tallguy:
Hi,

Rather than marking some areas as validated unseen, I would rather have
them briefly viewed. I think we've all seen an expert mapper accidentally
click on the 'complete, button and just a cursory view would check for this.

Nick (OSM=Tallguy)
my phone is responsible for any spelling mistakes!

PR #52, opened 2017-03-13
create and edit project enhancements

LindaAlblas:
Create project enhancements:
- Split a task by point (in addition to polygon)
- Ability to modify an AOI after drawing it

Edit project enhancements:
- Added fields for beginner/intermediate/advanced in the 'allowed users' section - just placeholders for now


feenster:
@LindaAlblas - a couple of comments from a usability perspective:
- on the set task sizes step, make text "Split a specific area into smaller tasks by drawing an area:" respond to current split mode point/area
- Also on that page, could the buttons be styled to indicate which mode is currently on
- On the draw page, if you step back to that page in the wizard, you need to click the draw button to allow aoi modify, but this may no be obvious.  Could this button remain enable, if the aoi was drawn in the first place?

LindaAlblas:
@feenster , requested changes made. Please review again.

LindaAlblas:
@feenster , I have added a few fields (in the permissions and tags bit of the project edit) reflecting what we discussed earlier

PR #62, opened 2017-03-16
refactor, new geospatial service, ability to add temp layer

LindaAlblas:
- Added ability to add 'temporary' layer on the fly during project creation. At the moment it only supports XYZ layers such as MapBox. To test, use this URL: https://api.mapbox.com/v4/openstreetmap.map-inh7ifmo/{z}/{x}/{y}.png?access_token=pk.eyJ1Ijoib3BlbnN0cmVldG1hcCIsImEiOiJhNVlHd29ZIn0.ti6wATGDWOmCnCYen-Ip7Q
The layer is then also added to the layer switcher and can be switched off.
- Refactor of conversion between GeoJSON and ol features. A new geospatial service handles these conversions.

### DIFF
--- a/client/app/admin/create-project/create-project.controller.js
+++ b/client/app/admin/create-project/create-project.controller.js
@@ -7,17 +7,15 @@
      */
     angular
         .module('taskingManager')
-        .controller('createProjectController', ['$scope', '$location', 'mapService', 'drawService', 'projectService', createProjectController]);
+        .controller('createProjectController', ['$scope', '$location', 'mapService', 'drawService', 'projectService','geospatialService', createProjectController]);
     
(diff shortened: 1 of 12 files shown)


feenster:
- Might have found a bug:  Add the wms at step 1, draw AOI, then got to step 2.  Unable to add task squares.

- Also, consider usability improvements to UI widget for adding the WMS url and selecting the service format.

feenster:
would it be worth including a refactor of client\app\project\project.controller.js>addProjectTasksToMap to use the new geospatial service in this PR?

LindaAlblas:
@feenster , made suggested changes and fixed the bug you found. I thought the geospatial service would be useful for conversions etc. Not sure if addProjectTasksToMap would fit in this picture..

PR #101, opened 2017-03-24
include images in gulp build

feenster:


client/gulpfile.js
```diff
@@ -90,35 +91,48 @@ gulp.task('sass:watch', function () {
...
+gulp.task('run', function (callback) {
     runSequence('create-dev-config',
-                'compile-sass',
-                'browser-sync',
-                'sass:watch',
-                callback);
-});
\ No newline at end of file
+        'compile-sass',
+        'copy_images_to_dist',
```
LindaAlblas: we don't need to copy images to dist when running the task, only when building it.

PR #104, opened 2017-03-27
add overview map

LindaAlblas:
Add overview map on /project/project_id page. It is hidden initially (click on the button on the left hand corner of the map to see it)


feenster:
I think the layout isn't working as intended in firefox

PR #139, opened 2017-04-03
Potlatch field papers editors

LindaAlblas:
Launch Potlatch2 and Field Papers. Walking Papers does not seem to exist anymore.


feenster:
One comment for you to consider, but looks fine

PR #166, opened 2017-04-10
Ht 62 validate by user with api

feenster:


client/app/project/project.controller.js
```diff
@@ -472,6 +427,7 @@
                 vm.resetStatusFlags();
                 vm.resetTaskData();
                 refreshProject(projectId);
+                populateMappedTaskByUserTable(projectId);
```
LindaAlblas: Calling the reset functions happens 4x exactly the same. Would it be useful to create a function to avoid the repetition?

PR #189, opened 2017-04-13
can download currents tasks as gpx for JOSM

feenster:
If you select JOSM as the editor you see a link to download the current tasks as GPX file.  Download then Open in Josm with File > Open

### DIFF
--- a/client/app/project/project.controller.js
+++ b/client/app/project/project.controller.js
@@ -7,9 +7,9 @@
      */
     angular
         .module('taskingManager')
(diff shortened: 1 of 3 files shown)


client/app/project/project.controller.js
```diff
@@ -930,6 +930,16 @@
             vm.taskUnLockError = false;
             vm.taskUnLockErrorMessage = '';
         }
+
+        /**
+         * Create the url for downloading the currently selected tasks as a gpx file
+         * @returns {string}
```
LindaAlblas: maybe we can put this function in the editor service?

PR #190, opened 2017-04-13
feedback changes

LindaAlblas:
Feedback changes and bugfixes:
- edit project: default due date not to 1970
- edit project: warnings populate fields only when status is published
- contribute: add search 'all' mapping level button (not working in API yet)
- contribute + users list: pagination max size
- comments on tasks: add characters remaining text


LindaAlblas:
@feenster , I realised this PR is mainly tiny changes that are easier for Zlata to review, so merging!

PR #200, opened 2017-04-14
Fix josm launch and gpx download bugs

feenster:


client/app/project/project.controller.js
```diff
@@ -936,7 +936,11 @@
          * @returns {string}
          */
         vm.getGpxDownloadURL = function(){
-            return editorService.getGPXUrl(vm.projectData.projectId, vm.getSelectTaskIds(), true);
+            //return editorService.getGPXUrl(vm.projectData.projectId, vm.getSelectTaskIds(), true);
```
LindaAlblas: can we still use the editor service or does the functionality need to be moved here? If it needs to be here, then remove line 939 please

PR #210, opened 2017-04-18
Project dashboard admin

LindaAlblas:
Project dashboard - work in progress

- project**s** dashboard for admins (/admin/dashboard) where users can see the projects that they have created. Data comes from the API. The lastUpdated date is hardcoded in the API still.
- individual project dashboard with stats on the project (project/528/dashboard). Data is mocked except the comments.  We could potentially use this page to display what is in the activity tab now? And link to it from the project page so any user can see this page? Just an idea.


feenster:
Just the comments about screen headings and button text, but happy for you to merge.

PR #215, opened 2017-04-19
Update dependencies

hunt3ri:


hunt3ri:
@feenster I just updated a bunch of our dependencies after noticing on your install that a few were out of date.  If u do a pip install you should get them all 👍

PR #218, opened 2017-04-20
Ht 111 enforce restrictions

feenster:


feenster:
Good spot.  Fixed now.

PR #265, opened 2017-04-30
Get consumer key from env variables

pgiraud:
This lets people create and use their own oauth key.
By the way, since I don't have the `consumer_secret` for the currently provided key, I'm not able to login in my locally installed instance. Did I miss something?


ethan-nelson:
Pierre, for your last question, I think so far everyone has been generating their own key/secret combos. It may be worthwhile for someone to register a set that devs can use on testing instances so we don't have to make everyone register an application just to develop locally.

PR #266, opened 2017-04-30
Ht 112 arbitary tasks

feenster:
- Can create arbitrary tasks from imported features
- Tidy up and refinement of project creation wizard
- Cannot save projects with feature count greater than limit


client/app/admin/create-project/create-project.controller.js
```diff
@@ -89,6 +93,21 @@
         }
 
         /**
+         *
```
LindaAlblas: Please add some docs

server/models/dtos/project_dto.py
```diff
@@ -42,7 +42,8 @@ class DraftProjectDTO(Model):
     """ Describes JSON model used for creating draft project """
     project_name = StringType(required=True, serialized_name='projectName')
     area_of_interest = BaseType(required=True, serialized_name='areaOfInterest')
-    tasks = BaseType(required=True)
+    tasks = BaseType(required=False)
+    arbitrary_tasks = BooleanType(required=True, serialized_name='arbitraryTasks')
```
hunt3ri: Suggest renaming to has_abitrary_tasks

server/services/project_admin_service.py
```diff
@@ -40,7 +42,11 @@ def create_draft_project(draft_project_dto: DraftProjectDTO) -> int:
         draft_project = Project()
         draft_project.create_draft_project(draft_project_dto, area_of_interest)
 
-        ProjectAdminService._attach_tasks_to_project(draft_project, draft_project_dto.tasks)
+        # if arbitrary_tasks requested, create tasks from aoi otherwise use tasks in DTO
+        tasks = GridService.tasks_from_aoi_features(draft_project_dto.area_of_interest) \
+            if draft_project_dto.arbitrary_tasks \
```
hunt3ri: What is the point of this if.else it doesn't seem to do anything?

PR #269, opened 2017-05-01
added db migration for tasks table

feenster:


server/models/postgis/task.py
```diff
@@ -115,9 +115,10 @@ class Task(db.Model):
     # Table has composite PK on (id and project_id)
     id = db.Column(db.Integer, primary_key=True)
     project_id = db.Column(db.Integer, db.ForeignKey('projects.id'), index=True, primary_key=True)
-    x = db.Column(db.Integer, nullable=False)
-    y = db.Column(db.Integer, nullable=False)
-    zoom = db.Column(db.Integer, nullable=False)
+    x = db.Column(db.Integer)
+    y = db.Column(db.Integer)
+    zoom = db.Column(db.Integer)
+    splittable = db.Column(db.Boolean, default=True)
```
pgiraud: Do you need a specific column for this? If there's a value in x, y and zoom already means that the task can be split, in my opinion.

hunt3ri: Hi @pgiraud I think deriving a conclusion from an absence of data (eg these columns are Null) is asking for trouble down the line, not to mention makes the code harder to grok.  Much better to always be explicit IMHO 😄

PR #275, opened 2017-05-02
handle aoi with 3d coords

feenster:


server/services/grid_service.py
```diff
@@ -152,11 +152,22 @@ def _convert_to_multipolygon(features: list) -> MultiPolygon:
...
 
         # Downsample 3D -> 2D
+        if geometry.has_z:
+            geometry = shapely.ops.transform(GridService._to_2d, geometry)
         wkt2d = geometry.wkt
         geom2d = shapely.wkt.loads(wkt2d)
 
         return geom2d
 
+    def _to_2d(x, y, z=None):
```
hunt3ri: Add type hints for these, and either describe what the X and Y are in the docs or delete.  As presumably it's fairly obvious.

PR #290, opened 2017-05-04
Priority Areas

hunt3ri:
Add code to persist priority areas


server/models/postgis/priority_area.py
```diff
@@ -0,0 +1,37 @@
+import geojson
+import json
+from server import db
+from geoalchemy2 import Geometry
+from server.models.postgis.utils import InvalidGeoJson, ST_SetSRID, ST_GeomFromGeoJSON
+
+# Priority areas aren't shared, however, this arch was taken from TM2 to ease data migration
```
pgiraud: Indeed, this many-to-many relationship isn't used actually. I did this in case several projects share the same AOI and priority areas.
Do you want to simplify this?
This should be easy to handle during data migration.

hunt3ri: Thanks for the note @pgiraud.  I think we *could* merge the areas_of_interest table and priority_areas tables into the projects table.  Which would simplify some of the queries, however, it is quite an intrusive refactor, that might be worth doing post migrating the data.  WDYT?

PR #308, opened 2017-05-08
GET comments API no authentication

LindaAlblas:


server/api/project_admin_api.py
```diff
@@ -309,8 +309,6 @@ def delete(self, project_id):
 
 class ProjectCommentsAPI(Resource):
 
-    @tm.pm_only()
-    @token_auth.login_required
```
feenster: Need to remove the  Authorization request param in the docs for swagger

LindaAlblas: Thanks. Done!

PR #316, opened 2017-05-11
Cannot validate own tasks

feenster:


server/services/validator_service.py
```diff
@@ -35,7 +36,11 @@ def lock_tasks_for_validation(validation_dto: LockForValidationDTO) -> TaskDTOs:
             if TaskStatus(task.task_status) not in [TaskStatus.MAPPED, TaskStatus.VALIDATED]:
                 raise ValidatatorServiceError(f'Task {task_id} is not MAPPED or VALIDATED')
 
-            # TODO can't validate tasks you own
+            # can't validate tasks you own unless you are a project manager (admin counts as project manager too)
+            is_project_manager = UserService.is_user_a_project_manager(validation_dto.user_id)
```
hunt3ri: I'd be tempted to put this code in it's own method to make the code more readable something like

is_mapped_by_validator()

would make it easier to test also.

PR #320, opened 2017-05-15
Mobile message

LindaAlblas:


LindaAlblas:
@feenster , I have updated the z indices

PR #338, opened 2017-05-19
Validate trim aoi

feenster:


server/api/grid_apis.py
```diff
@@ -72,6 +73,8 @@ def put(self):
         try:
             grid = GridService.trim_grid_to_aoi(grid_dto)
             return grid, 200
+        except (InvalidGeoJson) as e:
```
hunt3ri: No brackets need around the exception

server/api/grid_apis.py
```diff
@@ -72,6 +73,8 @@ def put(self):
...
+            return {"error": f'{str(e)}'},
```
hunt3ri: You need to return a 400 status here.

server/services/grid_service.py
```diff
@@ -32,6 +33,15 @@ def trim_grid_to_aoi(grid_dto: GridDTO) -> geojson.FeatureCollection:
 
         # create a shapely shape from the aoi
         aoi_multi_polygon_geojson = GridService.merge_to_multi_polygon(aoi, dissolve=True)
+
```
hunt3ri: As discussed, testing we have a multipolygon should be responsibility of the merge_to_multi_polygon function

server/api/grid_apis.py
feenster: done

server/api/grid_apis.py
feenster: done

server/services/grid_service.py
feenster: done

PR #341, opened 2017-05-19
check response status and use message

feenster:


client/app/admin/create-project/create-project.controller.js
```diff
@@ -236,8 +239,21 @@
...
+                switch (reason.status) {
+                    case 400:
+                        vm.trimErrorReason = reason.data.error;
+                        break;
+                    case 500:
+                        vm.trimErrorReason = reason.data.error;
+                        break;
+                    default:
+                        vm.trimErrorReason = 'The grid could not be trimmed. This may be because the combined data size of the AOI and the ' +
+                        'grid is too large. Try again with a smaller AOI, fewer tasks and/or larger tasks.'
```
LindaAlblas: We can't have any text in the controller. This would make translation difficult...We also can't use any error messages coming from the server.

client/app/admin/create-project/create-project.controller.js
```diff
@@ -433,11 +449,25 @@
...
+                    switch (reason.status) {
+                        case 400:
+                            vm.createProjectFailReason = reason.data.error
+                            break;
+                        case 500:
+                            vm.createProjectFailReason = reason.data.error
+                            break;
+                        default:
+                            vm.createProjectFailReason = 'The project could be be created. This may be because the combined data size of the AOI and the ' +
+                            'grid is too large. Try again with a smaller AOI, fewer tasks and/or larger tasks.'
```
LindaAlblas: Again, we can't have any text in the controller for translation reasons. We also can't use any error messages coming from the server.

client/app/admin/create-project/create-project.html
```diff
@@ -228,13 +238,22 @@ <h1 class="section__title">
...
+                                 class="alert alert--danger" role="alert">
+                                <p><strong>Error:</strong> AOI geometry must be a valid multipolygon
+                                </p>
+                            </div>
+                            <div ng-show="createProjectCtrl.createProjectFailReason == 500"
+                                 class="alert alert--danger" role="alert">
+                                <p><strong>Error:</strong> Project could not be created.
+                                </p>
+                            </div>
+                            <div ng-show="createProjectCtrl.createProjectFail == -1"
```
LindaAlblas: Should this be createProjectCtrl.createProjectFailReason instead of createProjectCtrl.createProjectFail?

feenster: done

PR #343, opened 2017-05-20
Updates home.html for translation.

bgirardot-np:
Not sure how to handle these pr's. The file home.html.en.json has the strings that need to be added to the main en.json file. Note that in this case, with home.html, a few of the entries in the current en.json are also in the home.html.en.json file.


bgirardot-np:
Looks like this file is part of learn.html pr.

PR #344, opened 2017-05-20
Adds translate syntax index.html

bgirardot-np:
index.html.en.json contains that strings that need to be added to the main en.json file.

The gettext file has no use and is just a leftover from the automatic string extraction process.

And the formatting of the long strings broken across multiple lines had to be changed for the automatic string extraction to work.


LindaAlblas:
If the gettext files have no use, could you put them in a separate folder? Maybe client/locale/gettext? If you add that folder to the .gitignore file, then it won't add it to git.

bgirardot-np:
I can just remove the .gettext files, i do not think there is a reason to keep them.

the *.en.json files are what I do not know how to handle. I realize that angular-translate will only look at the one, "real" en.json but....

doing this one .html file at a time, if I add the strings to the "real" en.json there will be conflicts with each .html file that I add the translation syntax to. Somehow we have to get the *.html.en.json text into the real en.json file as you merge in the updated files. 

It is hard to explain really, maybe we should speak about it on skype. I am happy to do whatever you suggest, i just need a suggestion for how you would like to handle those strings that need to be added to en.json.

LindaAlblas:
I'll have a look and see what we can do

LindaAlblas:
@bgirardot could you assign any PRs that a ready for review to me please?

bgirardot-np:
found an error with single quotes in some of the text. will fix and make a new pr

PR #346, opened 2017-05-21
Translate syntax for create-project.html

bgirardot-np:


bgirardot-np:
@LindaAlblas can you help me understand what check failed with this pr?

LindaAlblas:
@bgirardot it looks like it has nothing to do with your changes. Some server side code tests are failing. If you merge the latest develop into your branch and push it up then it will hopefully build again.

bgirardot-np:
@LindaAlblas  Can you start merging in these PR's for the translation?

bgirardot-np:
Yes, en.json should get updated later today. seems like it is no problem to do the .html files independently.

I have about 15-20 more pr's to make for the .html files as I did each file as a feature branch.

then one more pr to update en.json which will have all the strings from the individual pr's

PR #352, opened 2017-05-22
Ht 145 split tasks ui

feenster:


client/app/project/project.controller.js
```diff
@@ -1099,27 +1148,6 @@
 
         };
 
-        vm.resetTaskData = function () {
```
feenster: These functions were duplicated elswhere, so removing

PR #353, opened 2017-05-22
angular translate extract

LindaAlblas:
Please do a "npm install" because a new package was added

The Gulp task 'gulp translate-extract' which will extract any strings you have decorated for Angular Translate and put it in the locale folder


LindaAlblas:
@bgirardot Closing this PR for now. I'll leave the branch intact so a PR can be created if needed

PR #367, opened 2017-05-24
Adds full en.json

bgirardot-np:


bgirardot-np:
@LindaAlblas This is already uploaded to the Transifex website.

PR #382, opened 2017-05-24
Ht 149 status lost on stop mapping validating

feenster:
API's to stop mapping and validating without loosing status or incrementing stats


server/services/validator_service.py
```diff
@@ -129,6 +129,48 @@ def unlock_tasks_after_validation(validated_dto: UnlockAfterValidationDTO) -> Ta
         return task_dtos
 
     @staticmethod
+    def stop_validating_tasks(stop_validating_dto: StopValidationDTO) -> TaskDTOs:
```
hunt3ri: I think we should really refactor this, as it's a 90% cut and paste job.  So it means bugs will inevitably creep in as Unlock changes but Stop doesn't etc.  So common code can be called by both Unlock and Stop

server/services/mapping_service.py
```diff
@@ -96,6 +96,27 @@ def unlock_task_after_mapping(mapped_task: MappedTaskDTO) -> TaskDTO:
         return task.as_dto()
 
     @staticmethod
+    def stop_mapping_task(stop_task: StopMappingTaskDTO) -> TaskDTO:
```
hunt3ri: Common code between stop and unlock should be refactored into a single method shared by both, to prevent bugs and ease testing.

server/services/mapping_service.py
```diff
@@ -96,6 +89,29 @@ def unlock_task_after_mapping(mapped_task: MappedTaskDTO) -> TaskDTO:
...
+            # Parses comment to see if any users have been @'d
+            MessageService.send_message_after_comment(stop_task.user_id, stop_task.comment, task.id,
+                                                      stop_task.project_id)
+
+        task.reset_lock(stop_task.user_id, stop_task.comment)
+        return task.as_dto()
+
+    @staticmethod
+    def get_task_to_unlock(project_id, task_id, user_id):
+        task = MappingService.get_task(task_id, project_id)
```
hunt3ri: Nice
Suggest renaming to get_task_locked_by_user
Add a docstring here describing function and good to go

server/services/validator_service.py
```diff
@@ -129,6 +116,62 @@ def unlock_tasks_after_validation(validated_dto: UnlockAfterValidationDTO) -> Ta
...
+            dtos.append(task.as_dto())
+
+        task_dtos = TaskDTOs()
+        task_dtos.tasks = dtos
+
+        return task_dtos
+
+    @staticmethod
+    def get_unlockable_tasks(project_id, unlock_tasks, user_id):
+        tasks_to_unlock = []
```
hunt3ri: Suggest renaming this to get_tasks_locked_by_user
Also missing docstring

PR #385, opened 2017-05-24
revert legend translation

LindaAlblas:
@bgirardot the legend on the contribute page did not translate well and was generating Angular errors. I have reverted to non translatable strings for now. We'll probably have the same issue on the project page legend. I'll have a look at what is causing it when I get a moment.

### DIFF
--- a/client/app/contribute/contribute.html
+++ b/client/app/contribute/contribute.html
@@ -253,30 +253,30 @@
                     <div>
                         <div id="map" class="map-container"></div>
                         <div id="legend-container" class="ol-control legend-control">
-                            <h3>{{ 'Legend' | translate }}</h3>
+                            <h3>Legend</h3>
                             <div>
                                 <svg height="25" width="25">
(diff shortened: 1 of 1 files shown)


LindaAlblas:
@bgirardot are you happy for me to merge this?

bgirardot:
@LindaAlblas Oh I thought you had already removed this. I did not see any issues with the legend when I was looking at the site yesterday on staging.

If you think it needs to be reverted, please revert it.

LindaAlblas:
Ok will do. This is the issue:

legend-translate-issue

bgirardot-np:
that is strange, shouldn't I have seen that on staging 6 hours ago when I was looking for other reasons? It looked fine to me and in the screenshots I posted on another issue.

bgirardot-np:
Oh I see, different legend. It does seem to work on the contribute map legend as see here:

https://github.com/hotosm/tasking-manager/issues/388

PR #389, opened 2017-05-25
verify email address

LindaAlblas:


client/app/profile/profile.controller.js
```diff
@@ -22,6 +22,8 @@
         vm.errorSetRole = false;
         vm.errorSetLevel = false;
         vm.errorSetContactDetails = false;
+        vm.verificationEmailSent = false;
+        vm.errorVerificationEmailSent = false;
```
hunt3ri: Is this var needed?  Or is it required to keep state locally??  If it's required maybe add a comment to explain it's purpose.

LindaAlblas: It is needed because when the user updates their Facebook, Twitter or LinkedIn details, but not their email address then no email is sent but it was not an error either so can't use the same variable for both cases.

LindaAlblas: I have put a few comments in to make it a bit clearer

PR #403, opened 2017-05-25
Translate transifex updates

bgirardot-np:
Limited testing (can't get my local copy to let me login at the moment) but pages I could browse not logged in all looked good.


bgirardot-np:
Note: Italian is the only language that is 100% translated, so if you see non translated strings in Italian the need either the decorators or to have en.json updated.

PR #421, opened 2017-05-29
Ht 169 get projects v1

feenster:


server/api/project_apis.py
```diff
@@ -59,6 +59,60 @@ def get(self, project_id):
                 current_app.logger.critical(str(e))
 
 
+class ProjectSearchBBoxAPI(Resource):
+    def get(self):
```
hunt3ri: Should be admin only API?

PR #467, opened 2017-05-31
Ht 169 get projects v2

feenster:


server/api/project_apis.py
```diff
@@ -107,15 +115,18 @@ def get(self):
             search_dto.bbox = map(float, request.args.get('bbox').split(','))
             search_dto.input_srid = request.args.get('srid')
             search_dto.preferred_locale = request.environ.get('HTTP_ACCEPT_LANGUAGE')
+            search_dto.project_author = request.args.get('projectAuthor')
```
hunt3ri: Apologies, I thought this feature was for the user-profile not the project-dashboard.  So you were correct you can use a flag to limit by author, rather than provide the username.

In that case it's a lot simpler to just access the tm.authenticated_user param

server/models/dtos/project_dto.py
```diff
@@ -104,6 +104,7 @@ class ProjectSearchBBoxDTO(Model):
     bbox = ListType(FloatType, required=True, min_size=4, max_size=4)
     input_srid = IntType(required=True, choices=[4326])
     preferred_locale = StringType(required=True, default='en')
+    project_author = StringType(required=False, serialized_name='projectAuthor')
```
hunt3ri: This can be int as you'll have ID

server/services/project_search_service.py
```diff
@@ -1,12 +1,21 @@
...
 from flask import current_app
+from geoalchemy2 import shape
+import math
+from server.services.users.user_service import UserService
+
+# max area allowed for passed in bbox
+# 243375 is arbitrarily chosen map maximum map width in meters
+# multiply by 1.5 to give a buffer beyond edge of map
+# raise to power 2 to get the maxim allowed area
+MAX_AREA = math.pow(243375 * 1.5, 2)
```
hunt3ri: Is it not simpler to just just set the MAX area to the result of that calculation??

server/services/project_search_service.py
```diff
@@ -97,52 +114,88 @@ def _filter_projects(search_dto: ProjectSearchDTO):
...
+        # make a polygon from provided bounding box
+        polygon = ProjectSearchService._make_4326_polygon_from_bbox(search_bbox_dto.bbox, search_bbox_dto.input_srid)
+
+        # validate the bbox area is less than or equal to the max area allowed to prevent
+        # abuse of the api or performance issues from large requests
+        if not ProjectSearchService.validate_bbox_area(polygon):
+            raise BBoxTooBigError('Requested bounding box is too large')
+
+        # try to get a user id for the project_author
+        author_id = None
```
hunt3ri: Can remove this look up, and just use the ID (Sorry again)

PR #468, opened 2017-05-31
Per task instructions

hunt3ri:
Allow users to add instructions to each task


feenster:
Just a couple of comments for consideration, but otherwise looks fine

hunt3ri:
Thanks @feenster but the comments don't seem to be saved with the PR 😢

server/api/mapping_apis.py
```diff
@@ -42,7 +48,8 @@ def get(self, project_id, task_id):
                 description: Internal Server Error
         """
         try:
-            task = MappingService.get_task_as_dto(task_id, project_id)
+            preferred_locale = request.environ.get('HTTP_ACCEPT_LANGUAGE')
```
hunt3ri: No, it shouldn't - we also have a default value just in case :)

server/models/postgis/task.py
```diff
@@ -439,3 +435,37 @@ def as_dto(self):
...
+        for info in project_info:
+            if info.locale == search_locale:
+                return self.format_per_task_instructions(info.per_task_instructions)
+
+    def format_per_task_instructions(self, instructions) -> str:
+        """ Format instructions by looking for X, Y, Z tokens and replacing them with the task values """
+        if not instructions:
+            return ''  # No instructions so return empty string
+
+        if not self.splittable and ('{x}/{y}/{z}' in instructions):
```
hunt3ri: No it's more performant to write AND statements like this.  All interpreters will evaluate the first part of the statement and if it's false not bother evaluating the second

pgiraud: Sorry to bother but I'd be interested by more information about this performance issue.
I hope that the interpreter evaluates the first part and does not bother evaluating the second even without the brackets.
Please correct me if I'm wrong of point us to documentation about this.

hunt3ri: You are correct @pgiraud 👍

pgiraud:
I just wanted to mention that in TM2 we don't only support x, y, z in the per task instructions.
There was an addition to support any arbitrary property along with arbitrary geometries.
https://github.com/hotosm/osm-tasking-manager2/commit/5487acbbf3f9ffaeb25ad745dab4955af3fb2c00

hunt3ri:
Thanks @pgiraud yeah I'd seen that in the code.  However, looking at the TM2 data no one had ever used that feature.  So it didn't seem worth the effort to port it.  

So consider this PR a version 1 of the functionality, if anyone starts screaming for arbitrary grid properties then we could implement it.  

If you feel like porting it - a PR is always welcome :)

pgiraud:
Fair enough. ;-)
Please remember that not only HOT uses the tasking manager.
I'll see what I can do. Thanks for your efforts.
cc @batpad

althio:
> However, looking at the TM2 data no one had ever used that feature.

Not often, but you should not bet on 'never, ever'. At least on HOT
instance for Missing Maps projects focused on spread out villages, or
projects alongside roads, I think I can remember this functionality being
used.

> So consider this PR a version 1 of the functionality, if anyone starts
screaming for arbitrary grid properties then we could implement it.

Let's see ;)

> If you feel like porting it - a PR is always welcome :)

maning:
We've used arbitrary grids a lot for imports (see [LABuildings](https://github.com/osmlab/labuildings)) and I see this very useful in other situations.  Can we maintain this feature?  Happy for our team work on getting this ported to the new version.

hunt3ri:
Hi @althio @maning TMv3 supports arbitrary grids.  We're talking here about adding a Dynamic URL to per task instructions for attributes attached to the geoJSON when it's imported. 

Looking at the 2500 projects in my dataset this feature has never been used.

You will still be able to attach per task instructions for arbitrary grids.

pgiraud:
@maning @hunt3ri FYI I'll give the per task arbitrary attributes port a try today.

Issue #463, opened 2017-05-30
Fails to save project - error on saving


bgirardot:
If I make any changes to this project and hit "save changes" it just returns an error:

http://tasking-manager-staging.eu-west-1.elasticbeanstalk.com/admin/edit-project/2608

I do not see anything obviously wrong with it. I have also tried clearing it out to the most bare bones project, just all saves result in an error. 

Not a major issue, but might be informative to look at to see what is going on.

LindaAlblas:
@bgirardot this issue should now be fixed

PR #472, opened 2017-06-01
Other projects aoi

LindaAlblas:
- shows project AOIs on the project dashboard page and create project page
- bugfix for https://github.com/hotosm/tasking-manager/issues/463 (removing areaOfInterest before posting back)


feenster:
Only comment is would it be possible to make it a bit more obvious when the toggle control is inactive.  It's a very subtle change between active and inactive.  Happy to merge otherwise.

LindaAlblas:
It is now completely greyed out when not active

PR #477, opened 2017-06-01
change upload value on osmxml to false

feenster:
it seems that upload=no is not a valid value, should be upload = false


hunt3ri:
This PR should resolve issue raised in #477

PR #497, opened 2017-06-05
Bugfix - Ensure counters set for validate, invalidate all

hunt3ri:
Counters now set appropriately for validate, invalidate all.

### DIFF
--- a/server/services/validator_service.py
+++ b/server/services/validator_service.py
@@ -189,9 +189,31 @@ class ValidatorService:
     @staticmethod
     def invalidate_all_tasks(project_id: int, user_id: int):
         """ Invalidates all mapped tasks on a project"""
-        Task.invalidate_all(project_id, user_id)
+        mapped_tasks = Task.query.filter(Task.project_id == project_id,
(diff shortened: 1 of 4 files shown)


server/services/validator_service.py
```diff
@@ -189,9 +189,32 @@ def get_mapped_tasks_by_user(project_id: int) -> MappedTasks:
...
+                                              Task.task_status != TaskStatus.BADIMAGERY.value).all()
+
+        for task in tasks_to_validate:
+            task.mapped_by = user_id  # Ensure we set mapped by value
+            task.lock_task_for_validating(user_id)
+            task.unlock_task(user_id, new_state=TaskStatus.VALIDATED)
+
+        # Set counters to fully mapped and validated
+        project = ProjectService.get_project_by_id(project_id)
+        project.tasks_mapped = project.total_tasks
```
feenster: What about badimagery tasks?  You filter out badimagery tasks above, and elsewhere when we do stats we subtract and add the badimagery count?

hunt3ri:
Thanks @feenster well spotted.

Issue #502, opened 2017-06-05
Translatable strings visible for about a second after logging in


TylerRadford:
After logging in, I see the placeholders for translatable strings in brackets, this lasts for about a second before the strings are replaced with the menu items in English below

![image](https://cloud.githubusercontent.com/assets/9853903/26806631/ed358686-4a1f-11e7-8b0a-9ed618e1b473.png)

tastrax:
strings also seem to show permanently on Samsung Tablet  GT_P5100 Android 4.2.2

PR #519, opened 2017-06-07
fix for loading header translations

LindaAlblas:
This should fix https://github.com/hotosm/tasking-manager/issues/502


feenster:
Looks fine, seems to work.  We also see the issue on the tab text.  Could this technique be used there?

LindaAlblas:
ng-cloak only works on elements in the body so it won't work on the title. The title doesn't seem to get translated, also not in TM2 so I have removed translations from the title.

PR #541, opened 2017-06-09
Project search v2

hunt3ri:
Enhance search API so it returns all matches as points so we can draw them on the map.


server/models/dtos/project_dto.py
```diff
@@ -5,6 +5,7 @@
 from server.models.dtos.user_dto import is_known_mapping_level
 from server.models.dtos.stats_dto import Pagination
 from server.models.postgis.statuses import ProjectStatus, ProjectPriority, MappingTypes
+import geojson
```
hunt3ri: Remove uneeded import

PR #554, opened 2017-06-12
chat enhancements

LindaAlblas:
- Display time with Moment.js (3 days ago etc)
- Don't scroll to bottom again (only once)
- Add fix for being able to exceed max character limit
- Replaced the library for typeahead Ment.io with the latest version for a bugfix (https://github.com/jeff-collins/ment.io/releases/tag/0.9.24). The latest version is not available via npm. Maybe we should consider replacing the library with another one at some point.

### DIFF
--- a/client/app/components/project-chat/project-chat.directive.js
+++ b/client/app/components/project-chat/project-chat.directive.js
@@ -43,7 +43,7 @@
         vm.messages = [];
         vm.maxlengthComment = 250;
 
-        vm.hasNotScrolledYet = true;
+        vm.hasScrolled = false;
         
         // Errors
         vm.successMessageAdded = false;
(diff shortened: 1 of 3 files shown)


client/app/components/project-chat/project-chat.directive.js
```diff
@@ -42,11 +42,14 @@
         vm.message = '';
         vm.messages = [];
         vm.maxlengthComment = 250;
+
+        vm.hasNotScrolledYet = true;
```
feenster: Perhaps consider switching logic to hasScrolled rather than hasNotScrolled to avoid double negative.

feenster:
just a minor suggestion, otherwise happy to merge

PR #593, opened 2017-06-21
Add support for extra_properties in per task instructions

pgiraud:
See https://github.com/hotosm/tasking-manager/pull/468#issuecomment-305471518

With this pull request any property found in the file imported to create arbitrary task when creating a project will be recorded as `extra_property` in the database.
Then the project administrator can use this use additional information in the per task instructions.

I wanted to add a specific help for the per instructions textarea in the project editing form, but I couldn't find a way to determine whether a project has been created in arbitrary mode or not.

Any comments welcome.


feenster:
@pgiraud Regarding finding a way to determine if a project was created in arbitrary mode.  Currently that is not stored against the project.  However, it should be easy enough to start storing this.  The PUT admin/project api which is used to create draft the project takes a json property "arbitraryTasks": true|false in the request.  This is currently only used as a logic control to determine whether the tasks get derived from the AOI, or whether the gridded tasks squares created and passed in by the client application get used.  The 'mode' as you call it, is not stored.

One option for doing this could be:
- add a new task_mode attribute to the Project table, models etc.
- At this place in the service code, set the attribute on the project appropriately https://github.com/hotosm/tasking-manager/blob/develop/server/services/project_admin_service.py#L49

pgiraud:
@feenster Adding a `task_creation_mode` attribute would make sense.
Have you seen my [comment](https://github.com/hotosm/tasking-manager/issues/594#issue-237433816) about infinite zoom? Readding a zoom attribute at the project level would also do the job. No zoom would mean "not gridded".
And at the same time this would be easier to re-import data from TM2.

hunt3ri:
@david-hotosm can you confirm if your CI system automatically runs the migrations, as this PR from @pgiraud contains a new migrations script that will likely break the build if migration not run

pgiraud:
@hunt3ri @feenster I added a new commit to add a new `task_creation_mode` in the project table. This allows me to change the per task instructions help message according to whether the project mode is grid or arbitrary.

I couldn't test the following code: https://github.com/pgiraud/tasking-manager/blob/f611d9e29e44625373709a18232bbaf31ef7bc8e/devops/tm2-pg-migration/migrationscripts.sql#L76-L81
Can someone please confirm that it works correctly?

hunt3ri:
@pgiraud I had to revert this PR as it broke the app, we were no longer able to create projects.  I can see where the problem is, if you can resubmit it, I'll let you know.

pgiraud:
@hunt3ri The revert probably didn't happen as expected. I just found this file in the migration scripts: https://github.com/hotosm/tasking-manager/blob/develop/migrations/versions/7d55a089b5bc_.py
This includes the creation of the `extra_properties` column as well as the usage of `TaskCreationMode` which doesn't exist. This migration script is invalid in my opinion.
I don't know what to do in this case. The branch cannot be rebased without breaking other things.

hunt3ri:
Well spotted @pgiraud , I actually did this deliberately.  The revert was complicated as there was another migration in play.  So I decided to merge the two together, on the view that when you resubmitted the patch the db work would already be there (hope that makes sense)

The main problem with your original PR was that you were setting the arbitrary grid flag on the incorrect API request.  It should have been on the PUT project rather than the POST.  Happy to discuss with you if you like

pgiraud:
Thanks for the explanation. I hope I can work on this in the next days.

Issue #579, opened 2017-06-16
Project Creation - Max task squares limit problem


bgirardot:
@feenster 

Given that the current functionality grids the bounding box of the aoi and not just the aoi itself, it makes it impossible to get the needed size of task squares because the task square count of the total bounding box is used, not what will end up being the trimmed number of task squares.

I think the quick fix it to just make the maximum number of task square pretty high or remove the hard limit and just give a warning.

![2017-06-16_16h27_29](https://user-images.githubusercontent.com/265831/27230840-ec3bf24e-52b0-11e7-9f7b-d31a50f43868.png)

feenster:
The real hard limit here is that there is a 1GB limit on posting data to the server (this is an aws limit rather than anything configured within theTM3 application). With the current functionality, when gridding an AOI, the client app creates the task squares, then posts the AOI + tasks to the server for trimming, and for creating the project.  If the the size of the AOI + the tasks is greater than 1GB the server will respond with a 413 request entity too large.  The task square limit was imposed to mitigate this by making it less likely that the 1GB limit will be reached.  While increasing the task square limit will help in some cases, it will make it more likely that the 1GB limit is reached in other cases.  There is no hard and fast rule, it depends on the complexity of the AOI + the number of tasks created.

I see 3 options, but there may be others:

1. Increase the task limit and see how it goes..  Easy fix, but will make it more likely that the 1GB limit will be reached in certain scenarios.
2. Increase the 1GB limit on the server.  Not recommended as we should be keeping projects to a manageable size.  Risk that this cannot be configured in amazon anyway.
3. Move grid creation to the server.  This is the recommended solution.  Although the 1gb limit would still apply, it would be easier for users to work with as it would apply to the AOI only, since only the AOI needs posted to the server.  This will involve a non trivial refactor of existing client and server code.

bgirardot:
Could we just grid the actual aoi instead of the bbox of the aoi?

feenster:
We've tried that.  With bigger AOI's involving lots of task squares, Turf.js grinds to a halt as it tests each task  square to see if it intersects the aoi.

bgirardot:
Does it really get anywhere near a 1GB even for lots of task squares?

feenster:
Aggh, MB, not GB!  The limit is 1MB.  Sorry.

The attached file would do it if you went to a task size of 0.36 km2.  This creates a POST data size of 1.6MB.

[...]

PR #643, opened 2017-06-29
remove project task count limit and add error message for trim timout…

feenster:
… and unexpected error


feenster:
client side fix for addresses #579

PR #655, opened 2017-07-07
undo button

LindaAlblas:
Undo button which shows when you are logged in and when you have mapped/validated that task.

### DIFF
--- a/client/app/project/project.controller.js
+++ b/client/app/project/project.controller.js
@@ -339,6 +339,9 @@
             var taskId = vm.selectedTaskData.taskId;
             var resultsPromise = taskService.undo(projectId, taskId);
             resultsPromise.then(function (data) {
+                vm.resetErrors();
+                vm.resetStatusFlags();
+                vm.resetTaskData();
                 setUpSelectedTask(data);
                 refreshProject(vm.projectData.projectId);
             }, function (error) {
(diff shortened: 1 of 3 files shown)


feenster:
Haven't tested, but looks fine.  I can see that there is now an undoable flag on the task when it comes back from the getTask api; that is much better than using the history client side to work out if the current user did the last status change.  Looks like the buttons are in the right place.  More a question for the API, but I assume the undo goes in to the history just like a normal status change?

client/app/project/project.controller.js
```diff
@@ -330,6 +331,23 @@
...
         /**
+         * Undo task status change
+         */
+        vm.undo = function(){
+            vm.taskUndoError = false;
+            var projectId = vm.projectData.projectId;
+            var taskId = vm.selectedTaskData.taskId;
+            var resultsPromise = taskService.undo(projectId, taskId);
+            resultsPromise.then(function (data) {
+                setUpSelectedTask(data);
```
feenster: Should maybe put these lines in before line 342.  Not sure if it's necessary, but thinki the pattern is used elsewhere:
 
vm.resetErrors();
vm.resetStatusFlags();
vm.resetTaskData();

LindaAlblas:
@feenster thanks for reviewing. I have added the reset pattern to make sure it all gets reset properly. The undo does show in the history (see image). I'll merge and get @popelnuh to test it
undo

PR #659, opened 2017-07-09
Add support for extra_properties in per task instructions

pgiraud:
Replaces #593


pgiraud:
@hunt3ri I updated my code so that it doesn't include any the migration scripts anymore.
I also fixed obvious typos which were indeed preventing the project creation.
However I didn't really get our point about the PUT API request. https://github.com/hotosm/tasking-manager/pull/593#issuecomment-312585788
After some tests on my local instance, I can confirm that I'm able to create project in the different modes. I you don't please give me a procedure so that I can reproduce.

server/models/dtos/project_dto.py
```diff
@@ -89,6 +98,8 @@ class ProjectDTO(Model):
     priority_areas = BaseType(serialized_name='priorityAreas')
     last_updated = DateTimeType(serialized_name='lastUpdated')
     author = StringType()
+    task_creation_mode = StringType(required=True, serialized_name='taskCreationMode',
```
hunt3ri: Hi @pgiraud this is the source of the bug.  We are successfully creating the project using the ProjectAdminService.create_draft_project method, called from ProjectAdminAPI.put

However, we have a seperate API to allow the user to then modify their new project here ProjectAdminService.update_project(project_dto) called from the ProjectAdminAPI.post.  So the DTO is expecting that param to be there, but the front-end is not supplying it, meaning users can't update their project.

So I think we just need to remove the required=True, and the validators, as we want this property on the GET but we don't expect or need it on the POST.

Hope that makes sense.

pgiraud: I'm so sorry to bother but I'm still unable to reproduce the bug you are talking about. I may be stupid but I can always edit a project after it has been created.
Can you please describe the complete workflow (like if I was 5-years old)?

pgiraud: Of course I can remove the "required" and "validators" if it helps merging this issue. I don't want this to be a blocker. We can then discuss it further. Or I can take a bit more time to analyze it.

LindaAlblas: @pgiraud FYI, @hunt3ri is out of the office until the end of the month so not sure when he'll pick this up

3vivekb:
This feature is pretty important for imports.  It is something that exists in TM2.  I hope somebody can test and get this deployed.

bgirardot:
@3vivekb Hi vivek, I am not sure how to test this really. Any help would be appreciated on this PR. It will not really be able to be dealt with for at least a month or so as pierre is out of town and will not be able to look at it again before he returns.

Think you can help with it in some way? Test or provide a file and steps for testing? It is a feature I am totally not familiar with so I am not sure how to even test it.

bgirardot:
Hi @pgiraud and @ethan-nelson Do you think we could revisit this PR and see about getting it merged in?

I am happy to test it well if we can resolve the conflicts, but I do not have the skills to resolve the conflicts.

This bumped up in priority due to the MS Building footprints. Several folks would like to do the building imports based on arbitrary geometries (census tracts) which is what this enables.

pgiraud:
Hi @bgirardot I just rebased the PR. Please let me know if it works.

pgiraud:
@bgirardot Did you get a chance to test this?

bgirardot:
@pgiraud No, I apologize, I thought @ethan-nelson Was looking into this one Let me see about testing it later this week.

ethan-nelson:
Yes I can test this week. Thanks!

server/models/postgis/task.py
```diff
@@ -142,6 +143,7 @@ class Task(db.Model):
     x = db.Column(db.Integer)
     y = db.Column(db.Integer)
     zoom = db.Column(db.Integer)
+    extra_properties = db.Column(db.Unicode)
```
ethan-nelson: Note: this column already exists in the database ([added through the migrations](https://github.com/hotosm/tasking-manager/blob/56d1d13056bbea342132cf6e78b12f01c73d5dd3/migrations/versions/7d55a089b5bc_.py)), but it wasn't mapped into the DB models.

server/models/postgis/project.py
```diff
@@ -56,6 +56,7 @@ class Project(db.Model):
     license_id = db.Column(db.Integer, db.ForeignKey('licenses.id', name='fk_licenses'))
     geometry = db.Column(Geometry('MULTIPOLYGON', srid=4326))
     centroid = db.Column(Geometry('POINT', srid=4326))
+    task_creation_mode = db.Column(db.Integer, default=TaskCreationMode.GRID.value, nullable=False)
```
ethan-nelson: Note: this column already exists in the database ([added through the migrations](https://github.com/hotosm/tasking-manager/blob/56d1d13056bbea342132cf6e78b12f01c73d5dd3/migrations/versions/7d55a089b5bc_.py)), but it wasn't mapped into the DB models.

ethan-nelson:
Hi Pierre, I added a database revision to handle the case where a database has been migrated prior to your changes. In addition, I also added to the migration script the case where zoom values were not none so that projects would not have a blank value.

I ran the revision file and it worked for a local clone (the ratio was about 9 to 1 for non-arbitrary versus arbitrary). If you are okay with these changes, then we should be set to merge. Thanks for your help getting this implemented!

bgirardot:
I also got a little closer to getting a file to test this with. will update in a few days with the file hopefully.

PR #660, opened 2017-07-10
Add migration test

david-hotosm:
This adds a step to run migrations in the CI environment as part of our test suite.

PR #662, opened 2017-07-11
Fix migration error

hunt3ri:


hunt3ri:
This should fix the issue @david-hotosm was having with #660

PR #663, opened 2017-07-14
Feature/new styles

felskia:
Hi @LindaAlblas,

Here is a PR that is just the start of the styling changes. Let me know if it will be better if I keep working in this branch or if we merge this first and I keep working to spruce it up. 

Let me know what you think.

Thank you!


LindaAlblas:
Hi @felskia, it looks like you've been busy :). I think it is better to keep working on this branch and merge it once it is more complete. 

I think the /contribute page looks really neat (I am hoping that the map view will stay?). On the /project page, I think it would be good to display the map next to the instructions/map/validate/discussion. Or find another way to allow users to see the map and the map/validate 'panels' at the same time. My reason is that actions on the map change the map/validate panel content and vice versa.  I know all is still work in progress so please ignore if it was already on your to-do list. 

I like the design overall and I really like the colour scheme.

felskia:
@LindaAlblas thanks for the comments. Right now I'm a little stuck, when I login from localhost it takes me to the staging site. I have some small things to clean up here but my hope is that we can push this and then I can finish up the cleaning, as well as look at styles that will be affected while logged in. 

Let me know what you think. 

Thank you!

Ali

LindaAlblas:
Hi Ali,

For logging in on localhost you have two options. Running the API locally as well and pointing the app to the locally running API or follow the steps in question 4 in the FAQ: https://github.com/hotosm/tasking-manager/wiki/Dev-Environment-FAQ. Let me know if you need more help.

I think it would be good to do regression testing before pushing this to make sure that all is still working. What are your thoughts @bgirardot?

Thanks,

Linda

bgirardot-np:
Or this was possibly related to the hard coded path in /client/taskingmanager.config.json ?

LindaAlblas:
@bgirardot yes, if you want to **run the server locally** and login locally then replacing

"tmAPI": "http://tasking-manager-staging.eu-west-1.elasticbeanstalk.com/api/v1" 
for
"tmAPI": "localhost:5000/api/v1"

should do the trick. 

If you **don't want to run the API locally** (avoiding having to set all that up), then following the steps in question 4 in the FAQ allows you to login locally as well.

bgirardot-np:
@LindaAlblas I see the difference now, I have always run the whole thing locally, I should try and connect to a remote instance. That would make some things much easier.

LindaAlblas:
@felskia I have added a 'show/hide' for the share button. Could you have a look and see if that is what you were looking for? Thanks!

LindaAlblas:
@felskia @bgirardot I have added the metadata to the project page (below the description). Everything was already returned by the API so no API changes were required.

metadata

LindaAlblas:
@felskia @bgirardot I have added a character limit of 250 to the short description in the last commit. It will add '...' when it is not displaying the full short description. I have also removed the 'overflow-y: hidden' from the cards because this doesn't seem necessary anymore now it gets trimmed. Could you confirm that this is what you were asking for Blake? Thanks!

PR #667, opened 2017-07-28
Catch Master up with Develop for 3.0 tagging

bgirardot:
@david-hotosm Hi David, in looking at things, I saw default branch was changed to master, but in looking it still seemed to be behind develop.

so I switched the default branch back to develop and am creating this pull request because I think it will catch master up to develop and then we need delete and re-tag the master branch at this point for 3.0.

Can you please review this and double check my thinking and if this PR is correct merge it into master?


david-hotosm:
@bgirardot you seem to be right, the commits haven't been merged. According to [this PR](https://github.com/hotosm/tasking-manager/pull/666/commits) they are though. I'm opening a ticket with GitHub.

Issue #668, opened 2017-08-02
Can't "Stop Mapping" a task


bgirardot-np:
If I checkout a task to map, when I click "Stop Mapping" I get an error.

For example on this project:
http://tasking-manager-staging.eu-west-1.elasticbeanstalk.com/project/2651?task=71

"There was a problem unlocking the task. Task Lock API - unhandled error: list index out of range"

A second attempt to "Stop Mapping" gives a different error message:

"There was a problem unlocking the task. Task Lock API - unhandled error: No row was found for one()"

That same project seems to show other task still locked from 23 days ago along the west border.

Also when running locally I see these error messages:

```
--------------------------------------------------------------------------------
CRITICAL in mapping_apis [/home/bg/tasking-manager/server/api/mapping_apis.py:237]:
Task Lock API - unhandled error: No row was found for one()
--------------------------------------------------------------------------------
127.0.0.1 - - [02/Aug/2017 03:35:54] "POST /api/v1/project/1/task/17/stop-mapping HTTP/1.1" 500 -
```

Also not sure if it is related, but I also see these messages locally:

`127.0.0.1 - - [02/Aug/2017 03:36:26] "GET /api/v1/project/1/has-user-locked-tasks HTTP/1.1" 404 -
`

bgirardot-np:
@hunt3ri Hi Iain, any progress on this item? cheers

hunt3ri:
Hi @bgirardot I've merged the fix into the develop branch - you may need to merge up into Ali's branch to test on the demo site.

As I mentioned in the PR you'll need to run the following update statement to clean up the data caused by the bug

update task_history
set action_text = NULL
where id in (select th.id
from tasks t,
task_history th
where t.id = th.task_id
and t.project_id = th.project_id
and t.locked_by = th.user_id
and t.locked_by is not null
and th.action_text is not null
and t.mapped_by is null
and th.action = 'LOCKED_FOR_MAPPING');

Let me know if there's any further problems.  Thanks

bgirardot:
@hunt3ri Thank you very much! I will start with testing on the staging server and then we can catch up new-styles branch.

bgirardot:
Fixed, thank @hunt3ri

bgirardot:
We still get this error with seemingly increasing fequency lately:

"[2018-03-15 11:33:25,179] CRITICAL in mapping_apis: Task Lock API - unhandled error: No row was found for one()"

bgirardot:
[...]

PR #692, opened 2017-08-17
Incorrect null check, should check for empty list. 

hunt3ri:
This should fix error #668 when attempting to stop mapping.

@bgirardot note that any tasks you previously attempted to stop mapping on will now contain a data error that will need to be cleaned up with a database fix.  To fix these you will need to run the following Update on the DB

update task_history
   set action_text = NULL
 where id in (select th.id
                from tasks t,
                     task_history th
               where t.id = th.task_id
                 and t.project_id = th.project_id
                 and t.locked_by = th.user_id
                 and t.locked_by is not null
                 and th.action_text is not null
                 and t.mapped_by is null
                 and th.action = 'LOCKED_FOR_MAPPING');

### DIFF
--- a/server/models/postgis/task.py
+++ b/server/models/postgis/task.py
@@ -111,8 +111,8 @@ class TaskHistory(db.Model):
                     TaskHistory.action == TaskAction.STATE_CHANGE.name) \
             .order_by(TaskHistory.action_date.desc()).all()
 
-        if result is None:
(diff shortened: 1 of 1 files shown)


ronnymartinez:
i have face the same problem on API ''There was a problem unlocking the task. Task Lock API - unhandled error: Multiple rows were found for one()'' what should i do?

PR #708, opened 2017-08-22
What is new page

LindaAlblas:
The new page can be found here: app/about/news.html

A new item called 'news' has been added to the menu bar. I have named it 'news' instead of 'what's new' to save space. I did have to make the mobile menu appear sooner to fit the new item in. 

@david-hotosm can this be merged into develop?


bgirardot-np:
Hi Linda,

I think we should go with "What's New" as my understanding is this page is temporary to introduce the new changes to the tasking manager.

bgirardot:
@LindaAlblas If you could still change the "News" link to "What's New", I just needed to get this content in so we can deploy it to our production server.

PR #728, opened 2017-08-24
Update project-dashboard.html

IknowJoseph:
Added rel="noopener" to target="_blank" href
https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/
https://jakearchibald.com/2016/performance-benefits-of-rel-noopener/


bgirardot:
@IknowJoseph thank you very much for these, finally at a point where I understand enough about the code base to be sure these do not affect anything but their intended purpose.

IknowJoseph:
Thanks Blake, it's a very small change, but I'd like to keep involved where
I can. Hopefully will make some more impactful changes in the future.
Learning from small beginnings!

PR #733, opened 2017-08-24
Update map.service.js

IknowJoseph:
Removed hardcoded "http" as this will cause mixed-content issues on sites running on https. 
Added rel='noopener'

### DIFF
--- a/client/app/services/map.service.js
+++ b/client/app/services/map.service.js
@@ -43,8 +43,8 @@
                 layers: [
                     new ol.layer.Tile({
                         source: new ol.source.OSM({
-                            url: "http://{a-c}.tile.openstreetmap.fr/hot/{z}/{x}/{y}.png",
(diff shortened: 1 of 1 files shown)


IknowJoseph:
I realise that openstreetmap.fr don't have a valid https certificate, so by removing the http from the openstreetmap.fr URL we will cause errors for anyone running the TM over https. However, I believe the TM should be able to be run over HTTPS, so those wanting to do so should either assist osm.fr in gaining a valid certificate, or should use another tile source.

IknowJoseph:
Apologies, I've just checked https://a.tile.openstreetmap.fr/hot/10/1/1.png and there is no https problems here. With that in mind, the change I have made will allow https to work: Without the change in this request tiles will be fetched over http when the pages is running over https, causing a mixed content warning.

ethan-nelson:
Hi IknowJoseph, we've gone ahead and changed the map tiles protocol in #908, so I'm going to go ahead and close this PR--thanks for the submission! Other layers like the population density map, are also in queue. https should be coming soon.

PR #752, opened 2017-08-27
Fix cursor pointer on some elements

### LINKED ISSUE #696: clickable elements don't necessarily change cursor to pointer
A lot of elements in the UI are clickable. A lot of them don't change cursor on hover. Usually clickable elements show a "pointer" cursor on hover.
This is for example the case for tabs (in the front page, or in the project page) or for the dropdown to change the language.

I'd be happy to provide a pull request. But do you guys prefer to add an `<a>` HTML tag or should we try to handle this only with CSS?

pgiraud:
Fixes #696

Changes the way the language switcher displays :
 - no label "language" since it is in my opinion not required,
 - style and behavior consistent with other elements in nav bar.

Show a pointer on hover on tab items.

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -94,10 +94,10 @@
                     </li>
                 </ul>
                 <ul class="global-menu menu-secondary-wrapper">
-                    <li class="global-menu-item global-menu-item--pipe">
-                        <span class="label">Language:</span>
+                    <li class="global-menu-item">
                         <language-switch class="language-dropown-wrapper"></language-switch>
                     </li>
+                    <li class="global-menu-divider"></li>
                     <li class="global-menu-item">
                         <account-nav></account-nav>
                     </li>
--- a/client/app/components/language/language-switch.html
+++ b/client/app/components/language/language-switch.html
@@ -1,16 +1,18 @@
-  <div class="drop drop--down drop--align-left drop--open">
-      <div class="dropdown"
-          ng-click="languageSwitchCtrl.toggleMenu($event)">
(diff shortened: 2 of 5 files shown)


client/index.html
```diff
@@ -94,10 +94,10 @@ <h2 class="toggle-menu">
                     </li>
                 </ul>
                 <ul class="global-menu menu-secondary-wrapper">
-                    <li class="global-menu-item global-menu-item--pipe">
-                        <span class="label">Language:</span>
+                    <li class="global-menu-item">
                         <language-switch class="language-dropown-wrapper"></language-switch>
                     </li>
+                    <li class="global-menu-item--pipe">&nbsp;</li>
```
LindaAlblas: When the 'global-menu-item--pipe' class is applied to the li element, an empty element shouldn't be necessary?

pgiraud: 1. It is unfortunately required otherwise the separator isn't displayed. I changed it a bit so that it's done CSS-side.
2. I prefered to use a single element for the divider. It deserves a separated element in my opinion. I just changed the class name to something more self-explaining "global-menu-divider".

pgiraud:
> I'd personally use the .pointer class to add the pointer cursor instead of adding empty a tags.

What empty tags are you talking about? The `<li>` tag to show the divider? If so, then it could have been done in a separate pull request, sorry. It's not really related to the pointer cursor issue.

In my opinion, it's better to use `<a>` tags instead of adding a `.pointer` class since it helps for accessibility.

PR #759, opened 2017-08-29
Avoid project card footer transparency

pgiraud:
screenshot from 2017-08-29 09-42-30

Fixes #758

PR #760, opened 2017-08-29
Avoid mix of text in project cards

### LINKED ISSUE #758: Mix of text in project cards in contribute page
In some cases, the texts in the cards of the contribute page mix.

screenshot from 2017-08-29 09-37-45

pgiraud:
screenshot from 2017-08-29 09-47-35

An other option to fix #758
See #759 as well.
Fixes #758

### DIFF
--- a/client/assets/styles/sass/_contribute.scss
+++ b/client/assets/styles/sass/_contribute.scss
@@ -74,6 +74,7 @@
     .ng-isolate-scope {
       max-height: 180px;
       margin-bottom: 1em;
+      overflow: hidden;
     }
 
     @include media(large-up) {


pgiraud:
@ethan-nelson I rebased the branch onto develop and extra diff went out. Should be good now.

alekno:
# This is the new (worse) status:

grafik

from: http://tasks-stage.hotosm.org/contribute?difficulty=ALL

Have a look at #826

ethan-nelson:
I guess I don't understand the issue you're referring to @alekno. What I see rendered is what I would expect, where the description text is truncated and information about the organization, campaign and difficulty is shown. Then at the bottom, a summary of statistics is shown regarding percentage mapped and validated.

alekno:
Ethan, when this is what you expect, we can stop this discussion. The picture shows many issues in my understanding. It is not only an unacceptable box overflow solution. There is also still text alignment problems.
 And when you are annoyed by my contributions, tell me here. This is second time I feel ignored.

pgiraud:
@ethan-nelson can you please confirm that the version after the current pull request was merged is not deployed yet?

ethan-nelson:
@alekno I do appreciate your input, I'm just trying to get an understanding of the problem. Frontend dev is not my forte, and I didn't originally design this, so I need pointed out specifically what the problem is. The above was my crude interpretation of what the design was intended to be.

@pgiraud my understanding is changes to develop _are_ deployed to tasks-stage. The issue with the map layers not showing up on deploy was a separate problem and is corrected in #846

alekno:
So do I need to reopen an Issue report? SNAFU. #758 #759 #826

bgirardot:
@alekno Yes, please re-open any closed issues that should be open based on your view.

bgirardot:
@alekno I am having a hard time understanding the differences you are correcting. We are just not seeing what looks quite obvious to you. I think it must be very subtle and maybe even dependent on how many lines are in the project title so can be hard for us to reproduct locally. I only have 2 test projects in my local testing env so I might not see what you see with more real world data, like long titles.

But I will keep looking at it and see if I can tell what the differences I am not seeing are.

I see the BEGINNER vertical alignment issue, but nothing else honestly.

The text is a little too spacy, the red text, it would be good to tighten up those lines via the markdown rendering css.

Otherwise, maybe you just want longer cards for the projects so we can display more text?

cheers
blake

Issue #718, opened 2017-08-24
Messages - Inbox: Cannot delete messages


popelnuh:
Functinality to delete messages is no longer available. The div is still there but not working. 
![bin_msg](https://user-images.githubusercontent.com/18286711/29666643-0f619340-88d1-11e7-8738-777c40a1faf0.PNG)

popelnuh:
The bug has been introduced with the UI rework. thinkWhere will investigate.

PR #761, opened 2017-08-29
fix delete message button

LindaAlblas:
Show delete message button

Fix for https://github.com/hotosm/tasking-manager/issues/718

### DIFF
--- a/client/app/message/inbox.html
+++ b/client/app/message/inbox.html
@@ -37,8 +37,10 @@
                                     <span am-time-ago="message.sentDate | amUtc | amLocal"></span>
                                 </td>
                                 <td>
-                                    <i class="hot-ds-icon-trash-bin pointer"
-                                       ng-click="inboxCtrl.confirmDeleteMessage(message.messageId)"></i>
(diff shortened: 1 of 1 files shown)


client/app/message/inbox.html
```diff
@@ -37,8 +37,10 @@ <h1 class="section__title">
                                     <span am-time-ago="message.sentDate | amUtc | amLocal"></span>
                                 </td>
                                 <td>
-                                    <i class="hot-ds-icon-trash-bin pointer"
-                                       ng-click="inboxCtrl.confirmDeleteMessage(message.messageId)"></i>
+                                    <button type="button ng-binding" class="button"
```
pgiraud: Are you sure of value of type here?
It looks weird to me to have ng-binding here.
Shouldn't this be in `class` instead?

pgiraud: By the way, is it useful?

LindaAlblas: Good spot! Fix in https://github.com/hotosm/tasking-manager/pull/767

Issue #706, opened 2017-08-22
Mapper experience level for imported data


bgirardot:
Hi @hunt3ri @david-hotosm Wasn't the data import process or some part of that workflow going to update the mapper experience levels? I see on the tW staging site, all the imported users have the correct mapper experience level, but not on HOT's staging sites.

And, can we adjust those changeset buckets before we go live?

bgirardot:
@hunt3ri Hi Iain, I see issue #458 says when they visit their profile page. 

Is that the current functionality? How did the thinkWhere stage site get everyone's mapper experience level correct?

Is there any other way to get their experience level updated? as mentioned on 458, it would be better if it happened on login as well.

hunt3ri:
Hi @bgirardot I think I must have written a wee script to do it, I can't remember now.  It must have been a few lines that I had locally to call the appropriate function to do the update.

As you say the experience level will fill in, as user log in or people view the profiles.

bgirardot:
@hunt3ri If you can just confirm it gets updated on login that would be great. I just do not want everyone to have to visit their user profile to get that stat updated.

hunt3ri:
If they are a new user, it will create the stats, otherwise they need to visit their profile.  But to see their experience level they need to visit their profile anyway, right?

bgirardot:
@hunt3ri The main impact is on the validation process. Everyone is a beginner and those will be stuck at whatever level they were at when they did visit their user profile last.

On login would be great, or a wee script that could run on a regular basis, maybe weekly.

It caught me by surprise because on the tW staging server everyone's was updated so I thought it was part of the import process or ran at another time.

hunt3ri:
Script added in PR #764 let me know if you have any questions

bgirardot:
@david-hotosm can you try this on the next database import please?

PR #764, opened 2017-08-29
Refresh mapper level helper

hunt3ri:
Helper for #706   To allow us to set the mapper level for all users, I've added a wee helper that will initialise the level for all users in the db.  You will need to have an environment that can be connected to the live DB to run it.

Once updated the levels are recalculated everytime a users submits mapping so it should be fine.  You can run it from the solution as follows:

python manage.py refresh_levels

It will probably take an hour to run, but it will print out progress as it goes.  I seem to remember I had a clever hack to do this really fast previously, but I can't remember it.

@pgiraud @david-hotosm @bgirardot let me know if you have any questions

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -5,6 +5,7 @@ from flask_script import Manager
 
 from server import create_app
 from server.services.users.authentication_service import AuthenticationService
+from server.services.users.user_service import UserService
 
 # Initialise the flask app object
 application = create_app()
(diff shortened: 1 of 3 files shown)


bgirardot:
@hunt3ri Thank you very much Iain, slower is probably better anyway, it only needs to run once based on the stats being updated on a regular basis and we don't want to abuse the osm api.

Issue #789, opened 2017-09-08
Duplicate URL method of getting geojsons from TM3


bgirardot:
@ethan-nelson This is a good place to point out, in TM3 you can NOT get the aoi and tasks geojson easily now
[5:55 PM] 
it would be good to at least duplicate the below url schemes so it works exactly as tm2, but also add some simple ui to get them somewhere too. url manipution is great, but you know, not friendly ui

/project/nnnn.json = aoi only, not actual task squares

[5:51 PM] 
smit1678 i’ve just been hitting the project.json endpoint to download
[5:52 PM] 
ah ok
[5:52 PM] 
blake_hot /project/nnnn/tasks.json = the actual task boundaries as the project was actually created (larger than aoi)

[5:52 PM] 
I focus on tasks.json

[5:53 PM] 
because I load them in JOSM and trace the edge, the go out to next area so the projects fit like tetris objects

[5:53 PM] 
aoi is less interesting as I know we mapped more area with our boxes stair step.

ethan-nelson:
Fixed in #790 and #934

PR #790, opened 2017-09-10
Add project AOI json endpoint

ethan-nelson:
First half of #789: provide an API endpoint to access only the area of interest of a project as a JSON file.

Specific url endpoint is `/api/v1/project/#####/aoi`.

@bgirardot this doesn't completely match the url from TM2, but mostly follows naming scheme of other API endpoints. Please comment on yes/no for that.

Default behavior. Accessing the url as listed above will prompt download for a file, or if accessed via API GUI, a link to download the file will show up. File downloaded is named `project.json`:
When download as a file is not set or set to yes

By specifying ?as_file=false, the user can load the data in browser:
When download as a file is set to no


hunt3ri:
Nice work @ethan-nelson looks good 👍

Issue #781, opened 2017-09-02
Translation probably needs updated sources


michael63-osm:
Looking at the German interface I see lots of untranslated strings. But https://www.transifex.com/hotosm/tasking-manager-3/ tells me that all strings have been translated.

bgirardot:
Yes, this correct, the initial application was translated, but since then there have been a number of small adjustments to text based on feedback. We will need to update the sources for transifex and push them up to get the new strings translated.

arun-aji-balaji:
There is no resource for Tamil strings

Andygol:
It would be great when transifex source was updated for the further translation

ethan-nelson:
@arun-aji-balaji I have added Tamil as a language on transifex and is now ready for translating https://www.transifex.com/hotosm/tasking-manager-3/language/ta/.

For the others, many languages were updated two months ago (https://github.com/hotosm/tasking-manager/tree/develop/client/locale). I will look into the ones that weren't.

ethan-nelson:
I'm closing this for now as we have a few tickets open regarding translation and the general idea is it's always in progress. The process of updating strings is now a bit more integrated into our deployment workflow, so be sure to check out Transifex for translating and reviewing if you see most strings out of date.

PR #794, opened 2017-09-19
Update de.json

michael63-osm:
The previous German localization contained severe errors, probably due to lacking context for the translators. I streamlined everything on Transifex, this new file has been downloaded from there.


michael63-osm:
@bgirardot  These are the fixes which we discussed recently. Any chance of this getting merged in the near future?

bgirardot:
Hi Michael, I have to look into this more, I am not super good at transifex and the whole process.

michael63-osm:
Hi Blake, there is not much process know-how involved here. I changed the strings on Transifex, pulled them from there, they are contained in the `de.json` file which is the contents of this pull request. It should replace the existing `de.json` file. Looking at the "files changed" tab reveals that I just changed words in many strings and made sure that the formatting of the file is preserved. So this should not break anything.

Still to keep in mind: my issue #781 that `en.json` needs to be updated and put on Transifex for the modified user interface is still valid.

bgirardot:
@michael63-osm We worked on the translations over the weekend and now I understand the process better and where we keep the language files.

I pulled all the translations from transifex and ethan merged them in so they should include your updated DE translation.

You can review them on https://tasks-stage.hotosm.org/

Please let me know if you see any issues on the staging site.

Thank you very much for updating the DE translation, I see what you mean about how much needed improvements.

bgirardot:
@michael63-osm Hi Michael, as I mentioned above, all of your changes should be merged in and now moved from stage to the main Tasking Manager now. If this is not the case please let me know and we can re-open this. Thank you very much for fixing the German translation.

PR #797, opened 2017-09-21
docker-compose.yml for starting quickly

mojodna:
To get a running version of the API, including PostgreSQL and populated data:

```bash
docker-compose up
```


dkocich:
@hunt3ri I tried this branch locally. It managed to install (step 10 was run twice) and serve the app on port 5000 but I was unable to connect to it for some reason.

mojodna:
@hunt3ri this is fixed. `runserver` wasn't binding to an appropriate interface within the container and port 5000 is hard-coded in various places, so it makes sense to keep that.

Note: this will only work where Docker binds to ports on `localhost`, i.e. no `docker-machine` due to hard-coded `127.0.0.1` hostname values and `5000` port values.

ethan-nelson:
I don't deal with Docker, but if we can get @david-hotosm confirmation that the `Dockerfile` or other changes won't mess with the deployment cycle then I think this should be merged!

answerquest:
hello, any progress on this front?

dakotabenjamin:
This works great and is a much needed change to improve usability for local testing. @mojodna how would you suggest we change the README accordingly?

dakotabenjamin:
My other question is how secret environment variables are being passed into the containers- it seems the standard `export` command isn't working for me. 
1. set placeholder values and request them be changed in the readme
2. since they aren't necessary for local testing, simply noting the warnings in the documentation will suffice

Other things to note: if you have postgresql running locally on port 5432, you will have to stop it

ethan-nelson:
I started a wiki at https://github.com/hotosm/tasking-manager/wiki/Docker-Integration to help with documentation. Because this is only focused on the API/backend right now, it looks like the database environmental variable is sufficient (though we can throw the others from the Ubuntu tutorial on the wiki if we ever add the frontend). 

@dkocich regarding the error, I played around with this for quite some time spinning up new EC2 instances and trying the process over and over again. The way I got it working was to stop the compose (Ctrl+C) after I got the warning and start it again. For some reason, the psql daemon isn't running when migrations tries to connect the first time or something.

Anyone have any thoughts on that? Otherwise, I don't mind just merging as is for now and we can improve as more are exposed to it.

dakotabenjamin:
If it's a matter of waiting for the postgresql server to turn on we can add a `sleep n` command before the migration runs.

ethan-nelson:
Went ahead and finally merged this to get it in (and available for improvement or expansion by others). I know we are wanting to consolidate some of these scripts in `devops`, but wasn't sure where to put it yet.

Issue #793, opened 2017-09-19
Add Bing imagery layer to TM3 map views


bgirardot:
Hi Linda,

We have the key from microsoft,

> AioPAglzP9Qw32KN17dOkKYRdSlzj7W5kIQY6zct_UCmGc0WRxAh-QeiRBpRUgrv

This is the page about how to do it in OL3:

https://openlayersbook.github.io/ch04-interacting-with-raster-data-source/example-04.html

And this is the place in the code I think it needs to go, I am not 100% sure, but I think:
https://github.com/hotosm/tasking-manager/blob/develop/client/taskingmanager.config.json

Could you add that in at your earliest convenience please?
 cheer
blake

dcfrankland:
Hi Blake - just to inform that Linda is away training this week and committed to various other projects next week. If someone else is able to take this on then I would suggest doing so. 

Thanks,
Dave

PR #802, opened 2017-10-16
Add bing imagery to layer switcher

harry-wood:
Add bing imagery as a configureable layer option, enabled by default to appear
in the layer switcher

fixes: https://github.com/hotosm/tasking-manager/issues/793

Looks like this:
screen shot 2017-10-17 at 01 00 49


ethan-nelson:
Tested this and it works great on my end! I'm not too versed in OpenLayers, though, so not sure if there's anyone else who wishes to review?

PR #826, opened 2017-10-19
Update _contribute.scss

alekno:
This fixes some Firefox rendering issues. Solved for me text alignment and box overflow bugs.

### DIFF
--- a/client/assets/styles/sass/_contribute.scss
+++ b/client/assets/styles/sass/_contribute.scss
@@ -106,15 +106,15 @@
   }
 
   &__content {
-    padding: 1.3rem 1.8rem 1.75rem;
+    padding: 1.3rem 1.8rem;
     overflow: hidden;
 
     @include media(medium-up) {
-      padding: 2.5rem 2.5rem 3rem;
+      padding: 2.5rem 2.5rem;
     }
 
     dt {
-      padding-top: .2em;
+      /*padding-top: .2em; */
     }
   }
 


pgiraud:
Can you please provide a before/after screenshot?

alekno:
bild eingefugt am 2017_09_19 11_58 am
bild eingefugt am 2017_09_18 08_04 am

pgiraud:
What do you think about #759 and #760 ?
Also see https://github.com/hotosm/tasking-manager/issues/758

alekno:
Yes they are about this render issues. For me it seems to be fixed in firefox with the pull.

bgirardot:
@alekno Ok, I just tested this PR manually against develop branch as it is right now, which has some modification to the card to move the horizontal rule down. That is done and merged, but the BEGINNER and other status are still mis aligned with the label.

But this fix, this pull request makes things worse in my quick chrome and firefox testing, see below:

2017-10-21_22h25_54
2017-10-21_22h28_42

So this should stay closed.

If you can checkout develop branch again and fix the mapping level vertical alignment issue that would a PR I would merge right away :)

Sorry for the confusion, this is all based on me thinking this is one of the focuses of your efforts and concern.

cheers
blake

bgirardot:
Oh, and if anyone can figure out how to just tighten up all the line spacing in the cards that would be pretty sweet too. especially that red text, seems way too spacy.

Issue #808, opened 2017-10-18
Make pagination links buttons


smit1678:
Small usability issue. Helpful to have a pointer hand when mousing over the pagination links. 

![tm-mouseover](https://user-images.githubusercontent.com/796838/31701222-af37dea4-b3c6-11e7-8518-9dda02d0ba0d.gif)

pgiraud:
I missed that one when I worked on this PR https://github.com/hotosm/tasking-manager/pull/752

verdy-p:
I vote for it !

verdy-p:
And may be the list could be just refilling automatically when we reach the bottom, or could be refilled at bottom by clicking on a "Load more" button.

As each project is a now in a tile with constant width and height, these tiles should flow as inline boxes automatically and fill up the available width on screen (not necessarily two columns, may only one on smartphones in portrait mode, may be more on large screens...)
But it was designed here only as a table with two columns (not fun for mobile phone users).

xamanu:
Done for TM4.

PR #841, opened 2017-10-20
Add pointer for pagination

ethan-nelson:
Still not the best styling, but this will at least make clicking a bit more intuitive for now.


ethan-nelson:
Also pinging #808

PR #846, opened 2017-10-21
Include other layers to release

ethan-nelson:
The other map layers, namely population density and Bing, are currently only configured to run on local instances. This is adding those to the "release" environment, which will make them appear on the tasks-stage. and tasks. sites.

cc: @bgirardot this is why https://github.com/hotosm/tasking-manager/commit/4a0a7a2d4dd7f92b63434305dfbaab6ffa1bd04f#commitcomment-25107692 is happening. To confirm, we want both of these layers on the live instance?


ethan-nelson:
Duplicate of #853

PR #851, opened 2017-10-21
Update base translation strings

ethan-nelson:
I've generated a new listing of strings for translation. You can view the pull request diff to see changes. Alternatively view files with [new strings](https://gist.github.com/ethan-nelson/fee8e1d912888c80b5a1ad4f5d6d026c#file-strings_new-json) and [deprecated strings](https://gist.github.com/ethan-nelson/fee8e1d912888c80b5a1ad4f5d6d026c#file-strings_removed-json). I checked about 15 of the deprecated strings throughout the list and did not find instances where a string was marked as removed that is still in the codebase as a translatable string.

[This is the script I used to generate the files](https://gist.github.com/ethan-nelson/046af2e346e590b4fd7f3997801809db), which when cleaned up a bit we could throw in the codebase until an AngularJS solution is implemented.

This file is currently sorted by string beginning letter. We can leave it ordered in context of the files if you prefer. Note because of the sorting order, strings that start with lower case are at the bottom of the file currently.


bgirardot:
@ethan-nelson thank you very much for this. should we get rid of that blank string at the top?

ethan-nelson:
Sounds good. I believe that was in the code base somewhere but I will remove it.

ethan-nelson:
Should we, for now, leave old strings in the file? Or are you okay with removing them as I have?

bgirardot:
@ethan-nelson @pgiraud Please see #794 It is an update of the German translation, based on the original strings in the original en.json.

@michael63-osm Are these already updated in Transifex?

michael63-osm:
Of course I revised the translation on Transifex. The file in #794 is the one I pulled from Transifex after fixing the errors there.

bgirardot:
@michael63-osm then I still am not sure the PR for de.json is right. I swear all of that gets deploy/added or left out of the code repository totally when we deploy, we do tx -pull and it gets all of those at deployment time.

Again, I am not sure, I do not have transifex all figured out. I did a few months ago, but have forgotten exactly how it works.

ethan-nelson:
Closing this as we will be updating translations through Transifex first.

PR #853, opened 2017-10-22
Adds the confing change needed to deploy bing and pop density to 'release'

bgirardot:
Ethan, this makes it so the newly added layers show up in our testing environment, which I assume is very close to 'release' we just don't use any tools to manage.

See if this works for you and if so please merge it.

You should see:

2017-10-21_21h51_04


bgirardot:
@ethan-nelson If you know how to just merge the feature branch into hotosm/tasking-manager/, we can deploy that branch to a production environment very easily. I think I know how.

ethan-nelson:
@bgirardot my understanding from Slack earlier in the week is commits deploy to tasks-stage.hotosm.org from the develop branch automatically (with some delay due to the whole deployment process).

ethan-nelson:
You may have to Ctrl+Shift+R to do a hard refresh, but this has indeed auto-deployed to http://tasks-stage.hotosm.org/contribute?difficulty=ALL from develop. Awesome job on automation @david-hotosm

PR #858, opened 2017-10-22
Let the pm's see the draft/archived projects

### LINKED ISSUE #814: TM3 bug: impossible to see all projects (drafted, or archived are now invisible) in list of projects
We can no longer see the list of all projects (including drafts and archived)
We only see those that are published.

This was in TM2, making the admisnitration of projects now difficult if we can't find them easily: a draft is created, but will be published later, but how to see it? A project completes, but we want to edit some properties (e.g. post a link to download data), or just this archival is temporary to suspend a project that will continue later (e.g. after making instruction changes, or changing and testing a better imagery source).
The project can then be forgotten and will be difficult to locate.

Note how some users have already had difficulties to create projects that they did not fill completely (multiple attempts were they left the draft half-completed, sometimes completely empty, or duplicated by error): we need a way to maintain the list of active projects or modify some of them to put a lionk going to another project or explaning that it is currently suspended.

The status of projects is then likely to be changed at any time by project admins but we need to see the whole set of projects when we are logged in as project admins.

So please restore that functionality !

pgiraud:
I tried to quickly reproduce what we had in the version 2.
The idea is to let the project managers to be able to see the projects that are not in "published" status on the contribute page. Other users will continue to see the "published" only projects.

127 0 0 1-5000-contribute-difficulty all

fixes #814


bgirardot:
This sounds perfect. I will test. thank you thank you @pgiraud

alekno:
I am not sure this is the optimal solution to the draft problem. In TM2 I always struggled to get the view how an usual mapper is presented the hompage for navigation. The visible drafts in "urgent" colluded with the overview.
In TM3 we have the nice dashboard. Please exhaust this new nice concept. The dashboard just needs an "additional" table view for *global* projects. This global should be searchable and categorizable/organisaton. Maybe this view state should be saved.
So in my daily routine I would just have a browser link to the RESTful address, have a quick view on progresses, do changes, edits, coms and leave it. Work done!

verdy-p:
This is the same bug as the one I reported previously in #814.

pgiraud:
> This is the same bug as the one I reported previously in #814.

@verdy-p you probably missed that but this is not a bug report. This is a pull request instead (with patch inside).  ;-)

bgirardot:
@pgiraud Yes, but if when committing, you say "fixes #814 " it will auto close the issue.

I think you taught me that :)

We need to do that so we auto generate change logs or the commits comments should be auto generated change log.

bgirardot:
I can not get this to work.

I merge in pierre's changes, but it still does not show me the archived and draft projects.

And, does this PR add projects to the database?

I use this command after I make every change: 
cd client/ && npm install && gulp build && cd .. && venv/bin/python manage.py runserver -d

Is there something else I need to do? I have been shift-reloading.

I actually think I saw this work once, but then a shift reload and it disappeared again.

But I saw projects that were not in my test set, I only have 3, when it was done, I saw about 6

bgirardot:
I got it working, seems like I had to shift reload a few times or I did it too many times, but it works.

But it seems to show them to everyone and should only show to Project Managers and Admins?

pgiraud:
You're right sorry. The caching system is a bit agressive, and once the request to the service has been done once, then all users get the same result. Will see if this can be changed. I'm afraid it cannot.

pgiraud:
I just changed the way the role is taken into account. This doesn't seem to be affected by the cache anymore.
Please have a try.

hunt3ri:
PM's can see their draft and archived projects on their dashboard - which I'd hope is a much better way of managing them??

image

verdy-p:
only the creator can see these projects, you block the working teams and create a bottleneck on a single person, that still cannot see also the other projects of their team.

Such thing should be open to any TM project admin, to see not only their own projects they created themselves, but also projects from others (we already have filters if needed).
So we would see again the lists we could have in TM2.

hunt3ri:
I'd still suggest that extending the dashboard would be the better approach.  Perhaps adding a switch to see all draft and archived projects??

bgirardot:
@hunt3ri I totally agree, the Project Manager's dashboard would be great to extend that would help address this.

But this fixes it now, pretty simply and more in line with what folks are used to from TM2 so I think this first and then we can look at how the dashboard could be used as well, as we need the dashboard for campaign tags.

alekno:
Can this reverted easily? When we have a dashboard solution ( as @hunt3ri mentioned), I would like to have it rethought after a conceptional overhaul of the management UI. Just because I like separation of presentation and management as it originally has been presented to me.

bgirardot:
@alekno I do not see why this would need to be reverted to also add the dashboard functionality, including some management functionality on the dashboard.

What am I missing? Why would this need to be reverted if we added the dashboard for campaigns/organizations ?

alekno:
Just remember the last 3 desaster campaigns. In TM2 I had to navigate thru all the "urgent" drafts before I could figure out what is presented to the ordinary mapper.

bgirardot:
I see. Ok, I am sure we can accommodate that somehow, a checkbox to remove draft and archived from the list for PM's would do that.

But this is not something I want to wait on, the PR in question here fixes perhaps the most urgent issue the tm3 has.

bgirardot:
@pgiraud @ethan-nelson @smit1678 Can we get this merged into develop branch as soon as possible please?

ethan-nelson:
+1 on the check box for including or excluding in search in the future

pgiraud:
Given the tests from several persons, I think it's ready to be merged. Merging.

verdy-p:
Still I don't know if we'll be able to see all projects independantly of who created them: can different project admins working in cooperation for the same projects see all the projects they created (like in TM2), or only those each one created himself (on his own dashboard).

We still need a global dashboard for all project admins/creators at least, or for everyone to see like it was in TM2: so yes the global "contribute" page should offer a way to see all projects (including drafts and archived, even if this is not the default view). For now there's none, and the only way to see these drafts/archives is to attempt to view a specific project number.

And it's still impossible for anyone (including the project creator) in TM3 to do any premapping in drafts, or post-mapping after the archiving if it's not in "published" state (this was not the case in TM2). And this is also a severe regression of functionality.

verdy-p:
Note: I suppose this is fixed ONLY in "http://tasks-stage.hotosm.org/contribute?difficulty=ALL" but still not on the normal instance. Your comment "fixed" should also say that it is still not really deployed.

verdy-p:
Note also that the "tasks-stage" instance allows mapping in "draft" project, but still does not allow splitting any tile as long it is not "published".

bgirardot:
@verdy-p 

1. Yes, you can see draft and archived status for any project, regardless of creator.
2. Yes, a dashboard solution is also still needed.
3. Yes, no one can map/validate projects that are in draft/archived status, I created a new issue for that #892 
4. Please understand that as developers, we have a good grasp on how branches are used here in GitHub so we know what 'fixed' means.

Things in a "feature branch" (any branch not 'develop' or 'master' is a 'feature branch' is still in local testing only on individual developer's machines.

As soon as something gets merged into the 'develop' branch, we consider it production ready, but can be further tested publically on tasks-stage. 'develop' branch is auto deployed to tasks-stage

Once 'develop' branch is tested, we merge it into 'master' and then it is auto deployed to production, tasks.hotosm.org

5. I could not map things in draft on tasks-stage, that is why i created a new issue for it.

verdy-p:
But your #892 is a clear duplicate of my first bug #866.

PR #863, opened 2017-10-23
contribute text alignment fix

alekno:


pgiraud:
@alekno You added some modifications to line-height in your last commit. I don't think it has something to do with text alignment.
If possible it would be nice to fix one issue at a time. It helps keeping track of the modifications.
Is it possible for you to rework on (rebase) your commits so that this pull request is only addressing the text aligment issue? And address the line-spacing issue in an other PR?

PR #874, opened 2017-10-25
Proposed update to text and links

smit1678:
This PR proposes updates to text and links across the site to clarify connection and linking to OpenStreetMap. Key updates are on landing page, footer, Learn page, and About page. 

To do: 
- [x] review text
- [x] review and add translations where necessary

### DIFF
--- a/client/app/about/about.html
+++ b/client/app/about/about.html
@@ -3,13 +3,16 @@
         <header class="section__header">
             <div class="inner">
                 <h1 class="section__title">
-                    {{ 'About the OSM Tasking Manager' | translate }}
+                    {{ 'About the Tasking Manager' | translate }}
                 </h1>
                 <p>
-                    {{ "The OSM Tasking Manager is a mapping tool designed and built for the Humanitarian OpenStreetMap Team's collaborative mapping process. The purpose of the tool is to divide up a mapping project into smaller tasks that can be completed rapidly with many people working on the same overall area. It shows which areas need to be mapped and which areas need the mapping validated." | translate }}
+                    {{ "The Tasking Manager is a mapping tool designed and built for the Humanitarian OpenStreetMap Team's collaborative mapping process in OpenStreetMap. The purpose of the tool is to divide up a mapping project into smaller tasks that can be completed rapidly with many people working on the same overall area. It shows which areas need to be mapped and which areas need the mapping validated." | translate }}
                 </p>
                 <p>
                     {{ 'This approach allows the distribution of tasks to many individual mappers in the context of emergency or other humanitarian mapping scenario. It also allows monitoring of the overall project progress and helps improve the consistency of the mapping (e.g., elements to cover, specific tags to use, etc.).' | translate }}
+                </p>
+                <p>
+                    All of this work is done through <a href="https://www.openstreetmap.org/" target="_top">OpenStreetMap</a>. OpenStreetMap is the community-driven free and editable map of the world, supported by the not-for-profit <a href="http://wiki.osmfoundation.org/wiki/Main_Page" target="_top">OpenStreetMap Foundation</a>. Read more on the <a href="https://wiki.openstreetmap.org/wiki/Main_Page">OSM Wiki</a> or join the discussion with your <a href="http://wiki.openstreetmap.org/wiki/Mailing_lists#Country-Specific_Lists">local OSM community</a>.
                 </p>
             </div>
         </header>
@@ -29,7 +32,7 @@
                 <div class="inner">
                     <h2>{{ 'Sponsorship and Funding' | translate }}</h2>
                     <p>
-                        {{ 'The OSM Tasking Manager was designed and built for the' | translate }}
+                        {{ 'The Tasking Manager was designed and built for the' | translate }}
                         <a href="https://hotosm.org" target="_blank" rel="noopener">Humanitarian OpenStreetMap Team.</a>
                     </p>
                     <p>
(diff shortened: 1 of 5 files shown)


pedrito1414:
Hi Nate, it looks good.

Two things... I think the main headline is weak. I would go for something like "Maps for people that need them". The current headline is a bit generic and not very motivating.

On this one:
image

I would leave out "digitising imagery, ..." etc and just go for, "...and would like to help, visit our Learn page...." The language is technical and bit confusing for a newbie...

Last thing... On here:
image
I think you need an 's' on scenrios...

Hope that helps!

smit1678:
> same remark as before: an URL embedded in HTML must be encoded so that ampersands (&) will be represented as named or numeric character entities (&) otherwise it does not pass the strct validation and browserrs start using the "trick" parsing mode, with known defects (basically emulating the bugs of old versions of Internet Explorer). this is still required for HTML5 conformance

@verdy-p Tested this out and this seems to break the Contribute page. When encoding the `&`, Chrome then automatically encodes the `=`. The application then can't read the parameters and the the Contribute page breaks.

verdy-p:
@smith you are ***wrong*** here !
Chrome does NOT encode the "=" in query strings (this is not needed for HTML, it is part of the encoded query string in URL but "=" is safe in HTML and does not need further encoding).
ONLY the "&" must be HTML-encoded.
Do not confuse HTML-encoding (needed for embedded an correctly encoded URL in an HTML document) with the URL encoding.
Once again understand how this is layered:
* first: the encoding of specific parts in URLs (names and values in query strings)
** this creates a valid query string where the separate "&" and "=" are left unencoded as they are NOT part of these parts; the parameter names themselves and their values however are %XX-encoded from their UTF-8; this is the so called "URL-encoding" intially intended for www-forms encoding, and that can be used in URLs for HTTP "GET" method (in query strings) or in body contents (outside the URL for HTTP "PUT" method)
** the domain names are encoded differently using Punycode;
** the "/"-separated paths are encoded as specific in the HTTP protocol.
** the fragment identifier is left as plain UTF-8 at this stage (it is not part of the HTTP protocol and will only be used by renderers of the document, in its own internal UTF-8 encoding).
* second: assemble the URL by concatenatinging the URI scheme prefix ("http:" or "https:"), the path (starting by "/"), the query string (starting by "?") and the document fragment identifier (an anchor name or id in HTML. The URL you get can be pasted in the address bar of a browser.
* third: embed the URL in the HTML document: this the only place where you need to HTML-encode further some characters reserved by the HTML syntax: "&" changed to "&amp;", "<" changed to "&lt;" and ">" changed to "&gt;" : this is **really required** for conformance with HTML (not at all for conformance with URL, it is compeltely independant of it!); in practice this affects only the "&" present in query strings to separate the parameters in a query string, but it may happen as well in paths, or in fragment identifiers if an anchor is named with one of these non-recommanded characters).

smit1678:
@verdy-p I tested this locally and it didn't work. Can you open up a separate pull request against this branch and recommend the changes you're talking about? Please test locally on your end to make sure that it works.

verdy-p:
@smith: reread correctly what I wrote: only the "&" must be changed to "amp;" in URLs embedded in HTML, and this will REALLY work (for now the W3C HTML validator complains about this bug because not all web browsers will accept these non-conforming HTML code which brings to unparsable HTML issues and then to disabled/non-working links, or broken rendering of the whole page as the browser will attempt to use a non-HTML5 legacy parser with quirks enabled, and various CSS quirks enabled to mimic what was in IE6 or other old Netscape browsers!)

ethan-nelson:
Translations can come separately so I'd say merge it when ready to go. We're working on exporting all the new strings in a separate PR.

alekno:
@verdy-p You are right in the HTML context. But here we are in the JS (more specific AngularJS) context. You can see the links are in the brackets` {{ "..."}}`. This means it will be replaced with html after the JS engine translates the code.

verdy-p:
Even if the JSAngular will replace the brackets with the provided URL with its template engine tool, it does not mean that in that context you need to "HTML-ize" it in the HTML value that the string will contain (using Javascript regexps). If this is substituted blindly, you'll still get an incorrect HTML value within the JS string.
So this reference to Javascript string is out of topic: Javascript strings do not know that this is HTML, Javascript strings are just (UTF-16 encoded) strings which are not limited to HTML.
Here you are adding two other layers on top of HTML: the templating engine for how it processes string values (independantly of HTML they may contain or not, and the fact that in fact an URL alone is still not valid HTML, but remains embedded in the "<a href=*></a>" element as an attribute value which requires its own escaping also for quotation marks), and the Javascript code syntax.
For now the final HTML page is still not valid (look at W3C HTML validator results): this is fully independant of the fact you use Javascript or some tempalte en gine to generate the HTML code, or that your javascript code may also set attriobute values using DOM properties without ever needing to do explicitly any HTML-encoding. or other escaping, the HTML validation rules are independant of the way pages are generated in the underlying invisible code.

alekno:
See https://stackoverflow.com/questions/14512583/how-to-generate-url-encoded-anchor-links-with-angularjs
We need to have to use the native encodeURIComponent in javascript.

smit1678:
@verdy-p @alekno Now that we've merged this, can you guys start a separate Issue to track this conversation? 

@verdy-p Can you create a branch in your fork and propose a pull request for others to review with your recommendations? If you're not comfortable with writing some code, say so in new issue and then we can find someone to take up writing any code.

alekno:
Why did you merge? We were just on track for a solution.

smit1678:
@alekno This PR was primarily for text updates. Let's create a separate PR for any URL encoding changes.

alekno:
@smit1678
 client/app/home/home.html line 34 is about our discussion. It needs an angularjs filter.

Issue #627, opened 2017-06-27
Project lists do not show number of people actively mapping


bgirardot:
The TM2 showed how many people were currently mapping or validating a project (how many people had a task square locked)

russdeffner:
For project managers and mapathon organizers, this is valuable information. i.e. if I have a team employed, I'd like to be able to quickly see that the number of mappers is relatively equal to my team, i.e. they are all working the project assigned.  For mapathons it's also nice to show/tell 'hey, we have X number of mappers working on this right now' versus totals at the end.

hunt3ri:
@pgiraud how did you calculate the counts last time??  I'm thinking the only practical way to do it is to track usage in the DB - an in-memory approach isn't really practical with the amounts of projects and need to decrement counter if user hasn't been seen for a while?

pgiraud:
Actually, as far as I remember, the count of active users for a project was just the number of currently locked tasks. With the auto-unlock mechanism, inactive users eventually disapear as soon as tasks are unlocked.

That may not be easy to do in TM3 for several reasons including the project list request caching. Also the new data model (with only one table for both locks and states) may make things more difficult.

How difficult do you think it is to know how much tasks are locked?

hunt3ri:
@pgiraud that's a genius solution 💯 It should be pretty easy to get the counts using that apporach.  The tasks table has a task_status column.  So 1 = LOCKED_FOR_MAPPING and 3 = LOCKED_FOR_VALIDATION

So the following SQL would get the project level activty

select count(1) from tasks where task_status in (1, 3) and project_id = n

and if we wanted app wide project activity on the homepage a little cached API that dropped the project_id eg

select count(1) from tasks where task_status in (1, 3)

Could see if I could pull something together over an evening

hunt3ri:
@pgiraud @bgirardot can you remember if you could see the number of active mappers on the contribute page, or did you have to open the project to see them, or were they visible in both places?

bgirardot:
Both places. Thank you for thinking about this item iain

hunt3ri:
PR submitted, let me know what you think :)

ethan-nelson:
My understanding is this was solved with #879. Please reopen if that is not the case.

PR #879, opened 2017-10-25
Displays active mappers on contribute and mapping page

hunt3ri:
PR to show active mappers on contribute and project page to address #627  Couple of things to note:

* Counts use locked tasks as a proxy for active users (thanks @pgiraud for the awesome idea)
* Counts are cached for 30 seconds to reduce DB load, could reduce time a bit if needed
* @bgirardot you'll want to review the text I've used on the HTML and make sure it's translated as needed
* I'm not really a designer so someone might want to look at how counts are displayed, but works for now :)

Give me a shout if there's any questions

Thanks
Iain

### DIFF
--- a/client/app/contribute/contribute.html
+++ b/client/app/contribute/contribute.html
@@ -164,8 +164,7 @@
                                                     <!--<uib-progressbar
                                                             value="project.percentValidated"></uib-progressbar>-->
 
-                                                        <dt class='num--large'>{{ project.activeMappers }}</dt>
-                                                        <dd>{{ 'Active Mappers' | translate }}</dd>
+                                                        <dd ng-show="project.activeMappers > 0" title="{{ 'Active Mappers' | translate }}"><span class="fa fa-user"></span> {{ project.activeMappers }}</dd>
                                                     </dl>
                                                </section>
                                             </a>
--- a/client/app/contribute/contribute.html
+++ b/client/app/contribute/contribute.html
@@ -163,6 +163,10 @@
                                                         <dd>{{ 'Validated' | translate }}</dd>
                                                     <!--<uib-progressbar
                                                             value="project.percentValidated"></uib-progressbar>-->
+
+                                                        <dt class='num--large'>{{ project.activeMappers }}</dt>
+                                                        <dd>{{ 'Active Mappers' | translate }}</dd>
+                                                  </dl>
                                                </section>
                                             </a>
                                         </div>
--- a/client/app/contribute/contribute.html
+++ b/client/app/contribute/contribute.html
@@ -166,7 +166,7 @@
 
                                                         <dt class='num--large'>{{ project.activeMappers }}</dt>
                                                         <dd>{{ 'Active Mappers' | translate }}</dd>
-                                                  </dl>
+                                                    </dl>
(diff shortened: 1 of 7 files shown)


client/app/contribute/contribute.html
```diff
@@ -135,7 +135,7 @@ <h6 class="withDescription">Search</h6>
                                             <a href="/project/{{ project.projectId }}" class="project-result__card"
                                                  ng-repeat="project in contributeCtrl.results track by $index">
                                                 <section class="panel project-result__content">
-                                                    <span>{{ project.priority }}</span>
+                                                    <span>{{ project.priority }} </span>
```
pgiraud: Unrequired change.

server/models/postgis/project.py
```diff
@@ -308,6 +311,20 @@ def get_aoi_geometry_as_geojson(self):
...
 
+    @staticmethod
+    @cached(active_mappers_cache)
+    def get_active_mappers(project_id) -> int:
+        """ Get count of Locked tasks as a proxy for users who are currently active on the project """
+
+        active_mappers_query = f'''SELECT COUNT(1) 
+                                    FROM tasks 
+                                   WHERE task_status IN (1, 3)
+                                     AND project_id = {project_id}'''
```
pgiraud: Could this be turned into a more SQLAlchemy-like syntax?

pgiraud: Please check https://github.com/hotosm/tasking-manager/pull/882

pgiraud:
Maybe we could merge this one first and then address the "validator's count" later in an other PR. In my opinion it's already ready to be merged.

pgiraud:
I usually prefer when conflicts are resolved by doing an interactive. I can do it easily if you prefer @hunt3ri. Otherwise I can live with it. Please tell me.

pgiraud:
What about showing the "xx active mappers" only when there is at least one?

Here's how it looks with this PR.

capture d ecran 2017-10-29 a 12 06 02

In TM2, it was represented by an icon and only displayed when there were active users.
Here's how it was looking.

capture d ecran 2017-10-29 a 11 59 45

Any opinion?

hunt3ri:
@pgiraud using an icon is a great idea, saves worrying about translation too - happy for you to implement.

re hiding if 0 - Personally I prefer consistency in a UI, so I would show the 0 if there are no active mappers, rather than hide the icon.  But if you think hiding is better I'm cool with that.

pgiraud:
I understand your preference for consistency. However, I think that hiding the icon gives an additional information. You can immediately figure out which project has active users.
A similar UI can be seen in the github issues. There you can easily figure out which issue has comments:
https://github.com/hotosm/tasking-manager/issues

I'll work on this as soon as possible. Provided that we already have an icon library used in TM3. Which doesn't seem to be the case.

hunt3ri:
The client is using font-awesome so hopefully a suitable icon in there

pgiraud:
Updated. New rendering looks like this:
capture d ecran 2017-10-29 a 16 06 25

I kept the text as the value for the `title` attribute in case people want to know what the icon means.

hunt3ri:
Looks good 👍

PR #890, opened 2017-10-27
Count only tasks locked for mapping

pgiraud:
It makes more sense to exclude the tasks locked for validation. Even more when we take into account that a validator can lock several tasks at a time.

### DIFF
--- a/server/models/postgis/project.py
+++ b/server/models/postgis/project.py
@@ -317,8 +317,7 @@ class Project(db.Model):
         """ Get count of Locked tasks as a proxy for users who are currently active on the project """
 
         return Task.query \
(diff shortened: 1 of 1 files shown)


alekno:
So only mappers are counted for activity? Why not filter "unique count" of validators plus active mappers?

hunt3ri:
Yes totally agree that excluding tasks looked for validation makes sense.  Saw @russdeffner comment on slack about locking multiple tasks for validation would create huge mapper counts 😄

hunt3ri:
@alekno re unique count of validators - good idea.  We store the user user_id of the user who has the task locked, so if we did a GROUP BY on user_id we'd be able to count validators too.  I'll try and find some time over the weekend to take a look, if @pgiraud doesn't beat me to it 😉

Issue #875, opened 2017-10-25
Who locked a task?


pgiraud:
If you select a locked task on the map, it looks like it's impossible to know who's working on it currently.
The history only shows actions when they're finished.

In TM2, this information was at the top of the "contribute tab" by giving information about the task current status (locked or not) and the user name.
And it was also available in the history for the task.

![screenshot from 2017-10-25 09-29-13](https://user-images.githubusercontent.com/319774/31985966-097c9300-b967-11e7-80a0-8931482342cc.png)

pgiraud:
See https://github.com/hotosm/tasking-manager/issues/588 as well.

pgiraud:
My mistake, the information is already available on the `advanced` block. I still think that the current locked status should be available in the history block.

bgirardot:
@pgiraud It is also available on the "Activity and Stats" page.

It should be on the Task like your example, easy to find, otherwise you have no idea who is working on a task.

If we could get it to the main Task page that would be great.

olafveerman:
We ran into the same issue today during a validation effort. It would be great if the current mapper would be shown more clearly.

bgirardot:
pierre fixed this a while back, thank you @pgiraud

PR #897, opened 2017-10-29
Show the current lock state in history

### LINKED ISSUE #588: Locking Task for mapping does now show in history
When a task is locked for mapping an entry should be put in the task history.

2017-06-20_23h12_51

pgiraud:
Fixes #588 and #875

### DIFF
--- a/client/app/project/project.html
+++ b/client/app/project/project.html
@@ -934,8 +934,8 @@
                                         </div>
                                         <div class="metadata">
                                             <span am-time-ago="item.actionDate | amUtc | amLocal"></span>
-                                            <span ng-show="item.action === 'LOCKED_FOR_MAPPING'"> for {{ item.actionText |  amDurationFormat : 'minute'}} </span>
-                                            <span ng-show="item.action === 'LOCKED_FOR_VALIDATION'"> for {{ item.actionText |  amDurationFormat : 'minute'}} </span>
+                                            <span ng-show="item.action === 'LOCKED_FOR_MAPPING' && item.actionText"> for {{ item.actionText |  amDurationFormat : 'minute'}} </span>
+                                            <span ng-show="item.action === 'LOCKED_FOR_VALIDATION' && item.actionText"> for {{ item.actionText |  amDurationFormat : 'minute'}} </span>
                                         </div>
                                     </div>
                                 </div>
--- a/client/app/services/message.service.js
+++ b/client/app/services/message.service.js
@@ -209,7 +209,7 @@
         function formatUserNamesToLink(text){
             var regex = /@\[([^\]]+)\]/gi;
             // Find usernames with a regular expression. They all start with '[@' and end with ']'
-            var usernames = text.match(regex);
+            var usernames = text && text.match(regex);
             if (usernames) {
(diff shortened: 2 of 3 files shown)


hunt3ri:
I'm not entirely convinced by these changes, they are hard to parse mentally and non obvious  Could we do something within in the loop similar to  (pseudocode)

if action == LOCKED_FOR_VALIDATION && ! action_text 
    'Currently locked by user xx'

would be A LOT easier to understand.  Happy to discuss???

pgiraud:
I agree that it's not simple to read. However, I can't find any good solution to make it clearer.

The problem in my opinion is that `actionText` itself is not really meaningful. It can be *a comment*, *a state* or *a duration*.
I still need to hide the "fox xxxx sec/min" statement so I still need to check if there's an `actionText`. It's still required in my changes. I can't change it to something else IMO.

At first, I was happy to change to "Currently locked by..." but then I thought that the "xxx minutes ago" was also a relevant information. And I think I prefer to read "Locked by xxxx / xxx minutes ago".
The "for xxx minutes" would be there to differenciate locked/free state. This may not be distinctive enough though.
screenshot from 2017-11-07 11-53-47

verdy-p:
Ideally also, thre's no need to keep a "locked" state in the history if it has been unlocked:
- if it has been unlocked automatically without being freed explicitly by the user you should change the status to "locked by xxx for 2 hours then unlocked automatically"
- if the user unlocks it without clicking "done" or "validate" or"split", remove the lock from the history, it is back to normal without any action (we cannot track what has been done externally in OSM, or keep it only if it's more than 10 minutes)
- if the user clicks "done" or "validate", change to "locked for xxx minutes and mapped/validated" (keep the number of minutes only if it's more than 10 minutes).
- if the user clicks "split", change to "locked for xxx minutes and mapped/validated" (keep the number of minutes only if it's more than 10 minutes)

Many users are just using "Start mapping" to only look at the map in the editor (this is only a rapid check, not possible only within the slippy map of TM) but will not do anything (either because there's really nothing to add/remove/modify or because the user prefered not changing anything) and then unlock the task immediately to return to the previous state (todo, invalidated, bad imagery, done, validated). These temporary locks without any action are pollutting the history I think.

May be the TM could check the user's OSM edit history if it has some recent changeset (in the last 2 hours only) whose bounding box covers a part of the task, and see if we need to really keep the lock history as a real edit session, instead of just estimating it.

More generally, only the last lock is useful if it is still active (and this won't last for more than 2 hours and is already displayed in the current state).

bgirardot:
My testing on this shows that it puts an entry on the task that says it is checked out, but that if you let the checkout timeout after two hours, then we are back to nothing showing up in the history.

Starting and stopping mapping manually works fine, we get an entry in the task history.

So I would say if it could work with the timeout scenario everything would be perfect.

bgirardot:
Hi @hunt3ri I would like to merge this in. Do you have any further thoughts on it?

hunt3ri:
I don't want to hold it up, if you're keen to merge.  Go for it.

bgirardot:
@ethan-nelson Thank you very much for testing this also. And thank you @pgiraud !

PR #911, opened 2017-11-03
Add link to task gpx when mapping or validating

### LINKED ISSUE #616: Contribute - Map/Validate - Provide link to download gpx file for task

ethan-nelson:
Fixes #616. I have no experience with this front end framework, so this may not be the best way?

edit

validate

### DIFF
--- a/client/app/project/project.html
+++ b/client/app/project/project.html
@@ -408,6 +408,9 @@
                                                 </span>
                                             </div>
                                         </div>
+                                        <div>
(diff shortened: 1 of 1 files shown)


pgiraud:
Please be careful with words order when translated.
'Download blah blah' may turn to 'blah download blah' in other languages.

I know it's a bit tricky but translations should take care of this. So instead of splitting the sentence into 2 parts, it should be only one.
I don't want to block this PR however since this problem probably happens elsewhere already.

ethan-nelson:
Yes definitely understood on that--I couldn't find an easy way to do a translation with html in the string. I will merge for now since this is an issue consistently coming up on the mapper-support and do some follow-up assessing how widespread the split strings are in the front end.

Issue #829, opened 2017-10-19
Completion stats appear to be borked


Boggedy:
Prior to cutover this task was at 80% vaild and has been worked on continually since.

http://tasks.hotosm.org/project/3643/dashboard

Since cutover to TM3 the validation counter appears frozen @ 77%

Since cutover to TM3  some tasks have been invalidated from time to time and the task completion metric should therefore vary from time to time between 99% and 100% depending. It is nailed on 100% despite some tasks being invalidated right now

verdy-p:
I see only 100%/99%, so it's not frozen.
You've probably got a proxy caching issue on your side (possibly still using the former server address even if the DNS was changed so the domain name go to the new server instance after the version switch).

ethan-nelson:
This should be fixed now based on #913 and #1123. Please reopen if you continue to see issues.

ethan-nelson:
Nevermind, I see validation is stuck at 99%. This could be a corner case of a bad imagery task that was validated, then invalidated. It will require some more DB digging.

xamanu:
Related to #1460

xamanu:
Closed in favour of the follow up issue #1527

PR #913, opened 2017-11-03
Round project percent done/validated down

### LINKED ISSUE #868: Overachieving projects
We seem to have at least one project trying to out-do the others: 
overachiever

101% validated?

### LINKED ISSUE #909: Completion Stats should not round up
This is partially related to #829 

I almost archived a project because the stats showed 100% mapped and 100% validated; however, luckily I looked closer and saw there were several tasks still needing either mapping and/or validating. It appears the stats round up and that is causing the confusion.  It would be much better to truncate the stats percentage (i.e. show lower than actual completion) so mappers don't 'overlook'/look away from a project that is not complete because the stats are rounded up to 100%. It is already difficult to get mappers to take the last tasks without the tricky stats display.

ethan-nelson:
Rather than using the round function, int will floor the percent mapped or validated down to the nearest whole number--only 100% will happen when 100% are done. 99.9% will go to 99. 

Fixes #909, #829. Possibly the solution to #868 as well.

### DIFF
--- a/server/models/postgis/project.py
+++ b/server/models/postgis/project.py
@@ -297,8 +297,8 @@ class Project(db.Model):
         centroid_geojson = db.session.scalar(self.centroid.ST_AsGeoJSON())
         summary.aoi_centroid = geojson.loads(centroid_geojson)
 
-        summary.percent_mapped = round((self.tasks_mapped / (self.total_tasks - self.tasks_bad_imagery)) * 100, 0)
-        summary.percent_validated = round(((self.tasks_validated + self.tasks_bad_imagery) / self.total_tasks) * 100, 0)
+        summary.percent_mapped = int((self.tasks_mapped / (self.total_tasks - self.tasks_bad_imagery)) * 100)
+        summary.percent_validated = int(((self.tasks_validated + self.tasks_bad_imagery) / self.total_tasks) * 100)
 
         project_info = ProjectInfo.get_dto_for_locale(self.id, preferred_locale, self.default_locale)
         summary.name = project_info.name


pgiraud:
Approved. I wouldn't close #868 though.

verdy-p:
But I think there's still an issue now with "self.tasks_bad_imagery": if it is not zero, it can cause "mapping percentages" to be > 100%, or fully completed projects to have "validation percentage" < 100%, and sometimes it will cause a division by zero exception for the "mapped percentage".
I don't understand the logic where this "self.tasks_bad_imagery" is
* substracted from the total used as divisor for the first "mapping percentage", but
* added to the total used as divisor for the second "validation percentage".

For me "bad imagery" does not mean it is "done" or "validated", it is still blocking the validation, but may still be considered "done". So
* it should NOT be substrated from the total used for the divisor of the first percentage, but added to the numerator (this will avoid the division by zero)
* it should NOT be added to the total used for the numerator of the second percentage (it results sometimes in validation >100%), even if this means we cannot reach 100%: bad imagery may often be replaced by another one (for example without clouds), and validation may be also done by local knowledge or using on ground imagery instead of aerial, or GPX tracks, or waiting for newer imagery (in HOT, this may be done with drones in emergency situations or by special request to satellite imagery providers to open some areas).

```
 -  summary.percent_mapped = int((self.tasks_mapped / (self.total_tasks - self.tasks_bad_imagery)) * 100)
 -  summary.percent_validated = int(((self.tasks_validated + self.tasks_bad_imagery) / self.total_tasks) * 100)
 +  summary.percent_mapped = int((self.tasks_mapped + self.tasks_bad_imagery) / self.total_tasks * 100)
 +  summary.percent_validated = int(self.tasks_validated / self.total_tasks * 100)
```

PR #915, opened 2017-11-03
Include pm status in search cache

ethan-nelson:
Currently, the hash for the project search that allows it to be cached does not include the recently introduced `is_project_manager` variable. This means some searches return draft and archived projects even for non-project managers.

The following PR should alleviate that by caching the project manager status in that hash.

### DIFF
--- a/server/models/dtos/project_dto.py
+++ b/server/models/dtos/project_dto.py
@@ -111,7 +111,7 @@ class ProjectSearchDTO(Model):
                 hashable_mapping_types = hashable_mapping_types + mapping_type
 
         return hash((self.preferred_locale, self.mapper_level, hashable_mapping_types, self.organisation_tag,
-                     self.campaign_tag, self.page, self.text_search))
+                     self.campaign_tag, self.page, self.text_search, self.is_project_manager))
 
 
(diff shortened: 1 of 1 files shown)


ethan-nelson:
Not sure the best way to test this on a local install--does the cache work for non-AWS deployments?

ethan-nelson:
This _can_ be tested locally and solves the following problem:

User 1, PM, logs in and searches with criteria A. That is cached now without information of PM status.
User 2, non-PM, logs in and searches also with criteria A. If many searches haven't happened since, the cache from User 1 is returned--meaning the non-PM can see the archives and drafts in the results.

Another problem now arises:

It seems `tm.authenticated_user_id` doesn't refresh if a non-logged in user is accessing resources. Other places this value is called usually are in methods with `login_required`, which means anonymous users aren't handled in those.

bgirardot:
We know this still leaves one small part of the issue remaining, but I think it works well enough to merge.

The only case it does not solve is for non-logged in users visiting the 'contribute' page, but it seems unlikely to be an issue in actual usage, however, I will create a new issue for that remaining case.

Issue #866, opened 2017-10-23
Allow PMs to map/validate/split tasks in "draft" or "archived" projects


verdy-p:
When creating a project in TM3, the "draft" or "archived" state forbids any editing or splitting of unmapped tiles, or changing a mapped tile as "done" or invalidating it or marking it as "validated", EVEN for project admins or for the project creator, this is a regression compared to TM2.

verdy-p:
adjusted to match the same tags as in #892 (an equivalent later report)

bgirardot:
@verdy-p Please do not put "TM2 Regression" in the title of issues. There is a label for that now so it is redundant and makes the title of the issue hard to read.

verdy-p:
it was there since the beginning, before you asked mer to add labels; at that time there was still no "regression" label that I created later.

ethan-nelson:
For reference, this can be fixed with checking whether a user is a PM or admin near the beginning in https://github.com/hotosm/tasking-manager/blob/develop/server/services/project_service.py#L87-L116 and https://github.com/hotosm/tasking-manager/blob/develop/server/services/project_service.py#L132-L151

bgirardot:
This should be all taken care of with the merges listed above

PR #920, opened 2017-11-04
Disallow validation on projects in draft or archive status for mappers

### LINKED ISSUE #919: Any user can validate on non-published projects
If a regular user visits a project that is archived, for example, they can validate/invalidate tiles.

Needs a check here: https://github.com/hotosm/tasking-manager/blob/develop/server/services/project_service.py#L136 like there is on https://github.com/hotosm/tasking-manager/blob/develop/server/services/project_service.py#L92-L93

ethan-nelson:
Originally, anyone was able to validate or invalidate tasks on a project whether it was in draft, archive, or published mode. This restricts validation in archived or draft statuses to only project managers.

Fixes #919. Part of #866 is also resolved with this (only the validating).

### DIFF
--- a/server/models/postgis/statuses.py
+++ b/server/models/postgis/statuses.py
@@ -57,6 +57,7 @@ class ValidatingNotAllowed(Enum):
     USER_NOT_VALIDATOR = 100
     USER_NOT_ACCEPTED_LICENSE = 101
     USER_NOT_ON_ALLOWED_LIST = 102
+    PROJECT_NOT_PUBLISHED = 103
 
 
 class UserRole(Enum):
--- a/server/services/project_service.py
+++ b/server/services/project_service.py
@@ -134,6 +134,9 @@ class ProjectService:
         """ Check if the user is allowed to validate on the project in scope """
         project = ProjectService.get_project_by_id(project_id)
 
+        if ProjectStatus(project.status) != ProjectStatus.PUBLISHED:
+            return False, MappingNotAllowed.PROJECT_NOT_PUBLISHED
+
         if project.enforce_validator_role and not UserService.is_user_validator(user_id):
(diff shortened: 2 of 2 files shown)


server/models/postgis/statuses.py
```diff
@@ -57,6 +57,7 @@ class ValidatingNotAllowed(Enum):
     USER_NOT_VALIDATOR = 100
     USER_NOT_ACCEPTED_LICENSE = 101
     USER_NOT_ON_ALLOWED_LIST = 102
+    PROJECT_NOT_PUBLISHED = 103
```
ethan-nelson: These are defined in the postgis directory...does that mean we will need to add a database upgrade?

pgiraud: I'd be interested to know as well. I thought it was an ORM only thing but it looks like postgreSQL uses theses value internally. Most probably to add a constraint on the possible values.
What I don't get though is that we never use this class for any table in the model. Looks like there's no need for an Enum here.
Oh wait! This is not an `SQLAlchemy Enum`. It's from the python core library.
No database migration required, IMHO.

ethan-nelson:
The most recent update ensures that project managers can still validate on projects that are not published (i.e. archived or draft).

russdeffner:
@ethan-nelson why just validating? I like to test my projects in draft (i.e. map) and often  we do splitting, etc. - could see pm wanting to mark the bad imagery themselves as well before publishing. Is there another issue for 'all functions for pm's in draft/archive'?

ethan-nelson:
Hey @russdeffner, thanks for the feedback. Yes that functionality for mapping is definitely on the list to do and is possible following similar to what I did here--I may do it here in a few minutes. I just didn't want to combine the validation and mapping additions in the same PR. Hope you have a good holiday!

PR #922, opened 2017-11-05
Add authentication on search

ethan-nelson:
The end goal of this PR and prior (#858) is to allow project managers to see archived and draft projects in searching. We _don't_ want regular users to see draft and archived projects. This could normally be achieved by just adding `@tm.pm_only(False)` and `@token_auth.login_required` wrappers around the call, except we **also** want anonymous users to search for projects.

Unfortunately, the current state does not support that. The `tm.authenticated_user_id` variable is not set on a search call because

1) no authorization header is supplied and
2) there is no authentication token wrapper to define `tm.authenticated_user_id` based on the current user.

So we want a function that is (pseudo) `@token_auth.refresh_credentials` to solve these two items.

(1) is solved by adding the `HTTP_AUTHORIZATION` headers to the search call in the frontend and providing those as an input to the API call (L204-L209 of `server/api/project_apis.py`). This is in line with the other API endpoints that include authorization and is fairly straightforward.

(2) is a bit more complicated. To my knowledge from rummaging through the code and documentation for the flask_httpauth library, there isn't a function that will take the authorization headers as an input, parse the token from that, and update `tm.authenticated_user_id` to be either a) `None` if anonymous or b) the correct user id if logged in.

So, I've taken code from `auth_token.login_required` (https://github.com/miguelgrinberg/Flask-HTTPAuth/blob/master/flask_httpauth.py#L65-L71) that extracts the token from the authorization headers and implemented that here: `request.environ.get('HTTP_AUTHORIZATION').split(None, 1)[1]`

Then I call a pre-existing function that checks whether a token is valid and, if so, sets the `tm.authenticated_user_id` to be that of the user (https://github.com/hotosm/tasking-manager/blob/develop/server/services/users/authentication_service.py#L40): `verify_token(...)`.

After this, we know that `tm.authenticated_user_id` is associated with the user that is logged in. Or if they're not logged in, `verify_token` will fail due to an empty token and we will continue on in the function.

Please let me know if I'm interpreting any of these things the wrong way, or if there _is_ a function with this ability in the code.

### DIFF
--- a/server/api/project_apis.py
+++ b/server/api/project_apis.py
@@ -253,6 +253,8 @@ class ProjectSearchAPI(Resource):
             search_dto.campaign_tag = request.args.get('campaignTag')
             search_dto.page = int(request.args.get('page')) if request.args.get('page') else 1
             search_dto.text_search = request.args.get('textSearch')
+
+            # See https://github.com/hotosm/tasking-manager/pull/922 for more info
             try:
                 verify_token(request.environ.get('HTTP_AUTHORIZATION').split(None, 1)[1])
                 if UserService.is_user_a_project_manager(tm.authenticated_user_id):
--- a/server/api/project_apis.py
+++ b/server/api/project_apis.py
@@ -7,7 +7,7 @@ from server.models.dtos.project_dto import ProjectSearchDTO, ProjectSearchBBoxDT
 from server.services.project_search_service import ProjectSearchService, ProjectSearchServiceError, BBoxTooBigError
 from server.services.project_service import ProjectService, ProjectServiceError, NotFound
 from server.services.users.user_service import UserService
-from server.services.users.authentication_service import token_auth, tm
+from server.services.users.authentication_service import token_auth, tm, verify_token
 
 
 class ProjectAPI(Resource):
@@ -201,6 +201,11 @@ class ProjectSearchAPI(Resource):
             - application/json
         parameters:
             - in: header
+              name: Authorization
+              description: Base64 encoded session token
+              type: string
+              default: Token sessionTokenHere==
+            - in: header
               name: Accept-Language
               description: Language user is requesting
               type: string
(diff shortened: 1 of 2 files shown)


bgirardot:
This is also currently deployed https://tasks-demo.hotosm.org/ for testing.

My testing shows it works as intended.

bgirardot:
@hunt3ri Iain, could you please look at this and offer your opinion?

server/api/project_apis.py
```diff
@@ -248,9 +253,12 @@ def get(self):
             search_dto.campaign_tag = request.args.get('campaignTag')
             search_dto.page = int(request.args.get('page')) if request.args.get('page') else 1
             search_dto.text_search = request.args.get('textSearch')
-            if tm.authenticated_user_id and \
-                    UserService.is_user_a_project_manager(tm.authenticated_user_id):
-                search_dto.is_project_manager = True
+            try:
```
hunt3ri: Seems sensible enough to me.  I would add a comment here to explain what's happening as the search functionality is getting v. complex so might help new comers understand the intent.

ethan-nelson:
Also note this approach is similar to https://github.com/hotosm/tasking-manager/blob/56d1d13056bbea342132cf6e78b12f01c73d5dd3/server/api/mapping_apis.py#L61-L63. 

At some point I can work on making this an actual wrapper so that we can use the same code for anywhere authentication is optional.

ethan-nelson:
Tests will be forthcoming this evening (US time)!

ethan-nelson:
To transfer info from slack to here, the tests currently available are focused more on the functions as opposed to the actual API calls (which in this case is where the changes were). Unless/until we decide we want to start building out tests for the API, I think this won't really be testable.

I do want to reiterate the gist of these changes is already employed [elsewhere in the code](https://github.com/hotosm/tasking-manager/blob/56d1d13056bbea342132cf6e78b12f01c73d5dd3/server/api/mapping_apis.py#L61-L630) and there seems to be agreement from others. Thus I think this is eligible to merge to develop.

bgirardot:
@ethan-nelson How hard would it be to add a checkbox(es) to filter out or filter in drafts and archived projects? I am just afraid that this always displaying draft and archived projects will result in a lot of complaints that now it is too hard to find things because so much space is taken up by the draft/archived ones. 

I would guess it is basically an additional filter field much like the ones that exist now along the left hand side.

ethan-nelson:
@bgirardot let me take a look this evening or tomorrow. I will do it in a separate PR because I don't want to convolute the problem this one addresses for future references, but merging that in tandem with this PR would be just fine.

alekno:
This PR and former #922 as a quick workarrounds for the managing issue is in my opinion distraction for a real management solution. This fix is still TM2 thinking (regression) of doing things.
I am still for empowering the project dashboard for management and searching as a solution that fits into the  new TM3 concept.

ethan-nelson:
@alekno if possible, please throw that into an issue so it doesn't get lost.

I don't lead activations so I unfortunately don't know much what you mean about doing this in a project dashboard. Do you mean a list of all projects you are listed as a PM or admin on as opposed to the contribute search page?

PR #934, opened 2017-11-08
Update task GPX/XML file APIS to allow retrieval of all tasks

ethan-nelson:
The API endpoint that allows task downloads in GPX or OSM XML files currently requires task numbers to be specified. Some are interested in download all of the tasks in this form, so this PR allows for that.

Also these two endpoints did not have the best error handling in the event of a project or task specified not existing, so I've added a few checks for that.

For some reason I am unable to get tests running locally anymore, but I will try to add a test if I can get that fixed soon.

### DIFF
--- a/server/api/mapping_apis.py
+++ b/server/api/mapping_apis.py
@@ -346,8 +346,7 @@ class TasksAsGPX(Resource):
             - in: query
               name: tasks
               type: string
-              required: true
-              description: List of tasks required
+              description: List of tasks; leave blank for all
               default: 1,2
             - in: query
               name: as_file
@@ -368,8 +367,6 @@ class TasksAsGPX(Resource):
             current_app.logger.debug('GPX Called')
             tasks = request.args.get('tasks')
             as_file = strtobool(request.args.get('as_file')) if request.args.get('as_file') else False
-            if tasks is None:
-                return {"Error": 'No tasks supplied in querystring'}, 400
 
             xml = MappingService.generate_gpx(project_id, tasks)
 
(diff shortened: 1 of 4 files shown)


bgirardot:
Can you help me with the URL to test this?

bgirardot:
@ethan-nelson @smit1678 whoops, found an issue in this PR. tasks-as-osm-xml does not have an "as_file" option.

should not stop merging this now, but we will need another PR to add that I assume.

bgirardot:
@ethan-nelson @smit1678 I have a PR to address the lack of file download on the osm xml so i think this is taken care of. will make the PR shortly. (I just copy/pasted from ethans work and changed the file extension)

bgirardot:
@ethan-nelson Whoops, I did find an error. The new functionality works great, leave the tasks value blank to get all the tasks.

but if you try and use it as it worked before by putting in a task number, it errors.

2017-12-14_16h25_34

ethan-nelson:
@bgirardot Very strange, let me investigate this evening--I recall it used to work.

bgirardot:
Ya, honestly, I do not remember testing the individual downloads, I was only eager to see the whole project downloads working, which they do.

I do see the individual task downloads work on the current production site.

And I had the same error exactly the same way if I was using the gpx or the osm xml api calls, 1 task or 2 tasks specified.

bgirardot:
@ethan-nelson I see you specifically changed:

-        if not tasks or tasks.count() == 0:
+       if not tasks or len(tasks) == 0:

in server/services/mapping_service.py at L170 and L201

And the error is basically saying there is no len() for the object type tasks is (query.filter.*.all())

does going back to count() just fix this? I am assuming you changed to the len() for some reason

bgirardot:
eh, I fixed it in my usual hack'ey way. One sort of test works for what the list of tasks code returns and the other test works for what the all tasks code returns.

I just moved the offending if-then up to be part of the if-then-else block that determines which function to call depending on if it is getting individual tasks or all tasks and then used which ever test worked for the return value:

2017-12-15_01h06_47

All my testing both with found/not found projects and found/not found task id numbers, says this works ok.

If we can live with this, lets go with it.

Issue #592, opened 2017-06-21
enable export grid option for all, and add user id info on it


ViolaineDo:
It would be so helpful to :
- Enable any user to download the grid of the project, which is currently only possible by the project  managers.
- add the user id that completed the task, or validated it on the exported grid, so that it would give more flexibility on the validation process...

bgirardot:
@ethan-nelson @pgiraud This is the issue violane has as urgent and I agree, it is urgent.

verdy-p:
I agree it is missing, but you can still get the data using your browser:
- Go to the project page where the map is displayed, open the browser's developer panel, go to its "network" section.
- Refresh the page, and look at the list of "XHR": you'll see one request whose result in the content body is the JSON file used to render all tiles on the map. You can copy-paste this JSON to a text editor and save.
- You don't need to be a project manager to get this data (and it works also on TM2, or other TM instances where you're not a project admin). Basically this XHR is part of the documented TM server API.

pgiraud:
I may be wrong but the the request @verdy-p talks about has a simple URL which should look like:
http://tasks.hotosm.org/api/v1/project/xxx

hunt3ri:
@pgiraud is correct.  We built the restful API to make it easy for developers to extract TM info without having to scrape the page.  It's documented here

http://tasks.hotosm.org/api-docs

ViolaineCartONG:
Hi @hunt3ri Actually, I couldnt find the tool in the API (I'm not a developper). I tried _GET /api/v1/project/{project_id}/tasks-as-osm-xml_ but the issue is it looked like I needed to specify specific tasks I want to extract. I want all so I tried with "*", it returned "500" error . Nevertheless  with the adress @pgiraud suggested, I could extract the info only related to the grid. It's far from straight forward for our retired volunteers ^^ Is there a thing that could make it easier, a thing that I missed? Thanks

verdy-p:
For project 3648 (just an example) the following address is correct:

http://tasks.hotosm.org/api/v1/project/3648

It is effectively this URL which is used by the rendered map in the project page to retrieve the GeoJSON.

The suggestion by @pgiraud is equivalent to the one I made by looking for XHR requests in the network panel of the developers tools built in modern browsers, the address is exactly the same.

How can it be simpler?

bgirardot:
@verdy-p It could be simpler if when you clicked on the url it went to a save file dialog box, with a good default file name, instead of displaying a mess of xml.

[...]

PR #939, opened 2017-11-08
Add api for geojson of all task features

ethan-nelson:
The API endpoint `/api/v1/{project_id}` currently returns a geojson stream that contains both project metadata and tasks with geometry and metadata. While #938 allows for one to download that information in a json format, it is still not a valid geojson file that can be opened in JOSM.

This PR provides that functionality. The hope is this endpoint can be used to reproduce the `Export` button as seen in TM2 for PMs.

### DIFF
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -90,7 +90,7 @@ def init_flask_restful_routes(app):
     from server.api.health_check_api import HealthCheckAPI
     from server.api.license_apis import LicenseAPI, LicenceListAPI
     from server.api.mapping_apis import MappingTaskAPI, LockTaskForMappingAPI, UnlockTaskForMappingAPI, StopMappingAPI,\
-        TasksAsGPX, TasksAsOSM, UndoMappingAPI
+        TasksAsJson, TasksAsGPX, TasksAsOSM, UndoMappingAPI
     from server.api.messaging.message_apis import ProjectsMessageAll, HasNewMessages, GetAllMessages, MessagesAPI,\
         ResendEmailValidationAPI
     from server.api.messaging.project_chat_apis import ProjectChatAPI
@@ -137,6 +137,7 @@ def init_flask_restful_routes(app):
     api.add_resource(HasUserTaskOnProjectDetails,   '/api/v1/project/<int:project_id>/has-user-locked-tasks/details')
     api.add_resource(MappedTasksByUser,             '/api/v1/project/<int:project_id>/mapped-tasks-by-user')
     api.add_resource(ProjectSummaryAPI,             '/api/v1/project/<int:project_id>/summary')
+    api.add_resource(TasksAsJson,                   '/api/v1/project/<int:project_id>/tasks')
     api.add_resource(TasksAsGPX,                    '/api/v1/project/<int:project_id>/tasks_as_gpx')
     api.add_resource(TasksAsOSM,                    '/api/v1/project/<int:project_id>/tasks-as-osm-xml')
     api.add_resource(LockTaskForMappingAPI,         '/api/v1/project/<int:project_id>/task/<int:task_id>/lock-for-mapping')
--- a/server/api/mapping_apis.py
+++ b/server/api/mapping_apis.py
@@ -7,6 +7,7 @@ from schematics.exceptions import DataError
 
 from server.models.dtos.mapping_dto import MappedTaskDTO, LockTaskDTO, StopMappingTaskDTO
 from server.services.mapping_service import MappingService, MappingServiceError, NotFound, UserLicenseError
+from server.services.project_service import ProjectService, ProjectServiceError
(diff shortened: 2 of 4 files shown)


ethan-nelson:
This may fix #592

verdy-p:
So basically you want to have a link with the list of tasks. Yes this was a valid GeoJSON, but highly fragmented by splitting all tiles with some properties for each task that are not useful at all (id, x, y, zoom) and unpredictable (this will change each time a task is split and it is unnecessarily complex).

It would be better to have a an API which returns only the "AOI" (already returned also as a valid GeoJON in one property of the global JSON), and a variant that contains the "preprocessed AOI" (with its clipping, or its added buffer and extension to squares). These two should still contain most other properties of the project (name, description), and should be returned as a single "Feature", not a Features Collection which is overly complex and requires complex editing in JOSM or other GeoJSON editors).

Note that the two existing GeoJSON currently returned in the global JOSN of the project is not necessarily fully conformant: a strict GeoJSON should be made so that "outer" rings are ordered anticlockwise, and "inner" rings are ordered clockwise. I know that OSM has no such requirement now (except for lands limited by coastlines), but this correction is done by all OSM renderers (in osm2gis tools) and as well should be done early on AOI geometries submitted to the Tasking Manager before any further processing by clipping or by extension with buffers and squares.

verdy-p:
Also could the API endpoint `/api/v1/{project_id}` return an additional `"type":"TaskingManagerProject"` property in its root object instead of an untyped JSON ?

May be it could return also some properties of the API as
  `"API":{ "url":".../api/", "version":"1", "lang":"en" }`
or info for debugging requests as
  `"debug":{ "status":"OK", ... }`
for all documented APIs, that will be useful when they fail (the "status" should be displayable, it is different from the HTTP response status line and intended to be read by humans, may be eventually localized preferably in the language specified by the client but effectively in the language indicated by "lang":"en" in the "API" object)

This would make it coherent with GeoJSON and other common formats whose ALL objects (features, collections, geometries) have a "type":* property.

ethan-nelson:
Hi @verdy-p, the AOI is already available from an endpoint I added recently. I'm of the opinion to not try and add simplifications or things related to that when returning the shape because I believe that some project managers like to take the AOI from a project and reuse it for another project (though, granted, we now have the clone project feature). So I like to leave it in as original of a data format as possible. Having a _second_ endpoint with a simplified geometry, or perhaps adding a flag to the AOI endpoint with simplify=True could be a good idea. Could you open a ticket with that idea so we can keep track?

This specific feature--namely a geojson of all the tasks with information about each task--was requested by a couple of people, so that's why I'm adding it. It may not be useful to all, but if it's useful to some then I think it's worth it. I'm in favor of having more endpoints in the API than less as everyone's needs are slightly different and, with more available information people can try to build more things.

Regarding the right hand rule on the shapes, that is all determined by the geojson library in python and thus something I don't want to touch with a 50-meter pole. If there are features or flags native to the library we can switch to ensure conformity, I am all ears.

And regarding adding a type for the solo project json, if you don't mind could you open an issue with that and the other information so I don't forget. It would be an easy fix, but I don't want to make it at the present time until the other PR gets merged to avoid resolving git conflicts.

Thanks for all of the feedback!

bgirardot:
@ethan-nelson Clone project clones everything BUT the AoI so there might actually be more need for it now :)

verdy-p:
I make a difference between the AOI and the effectively area covered by the task manager and that are completed and validated: it is larger because of the optional "buffer" which may be added around it before splitting, and because when we select suare tiling, this buffer is extended again so that we don't find tiny residues: the initial tiles are all the same tiles. If we choose manual splitting at creation fo the project, instead of square tiling, the tiles may have random forms. In all cases the tiles will contionue to be split along the life of the project when anyone splits tasks.
The exact geometry of subdivided tiles has no interest for other projects than TM itself for interacting with tasks to do. They are still globally part of the initial area that covers at least (but not only) the initial AOI.
The AOI may be very complex in terms of gemetry, but the geometry of tasks is also overcomplified and not relevant, all I'd like to have is a single geometry object that contain the whole union of all task areas, but not without the additional splitting points. When projects are splitted with square tiles, the result is a very simple polygon (sometimes multipolygon when there are several disconnected areas) whose geometry uses only basic horizontal or vertical lines. This is not true if we use the "clip" option where this coverage polygon will be exactly the same as the initial AOI.
In TM2, the "Export" link provided the set of tiles (including the optional buffers) splitted, but it was still usable and is still what is now in the global project JSON (in the "tasks" property), and here also we have no details about the project (project number, name, instructions...). This has not changed.
So I wonder which area you want to export. §I suppose that people would want the initial AOI but the larger area effectively covered by all tiles in the TM project (which also fully includes the optional buffer around the AOI)  would be useful to get accurate view of what is really mapped in a project (and will also be mapped again when we clone a project)....

Issue #673, opened 2017-08-15
mobile phone warning should have a close option


Nick-Tallguy:
I often review tasks on my way to a mapathon I will be teaching at - using my phone on the train.

The warning to mobile phone users is good but there should be an option to hide it, perhaps on a per user basis.

Typed on my phone on a train!

bgirardot:
@LindaAlblas @felskia Can we add this easily?

It being based on screen size also causes an issue. I could not demo using an overhead projector because it forced my screen size to something that caused this to display.

russdeffner:
I would say this is a priority, if say I need to publish a disaster project on the road even if it's tough like it was on mobile with TM2, let me struggle my way but I don't see any way to get that done on a phone now; at least on iOS I couldn't figure out how to log in or anything.

verdy-p:
See also bug #820.

harry-wood:
I started looking into this but actually [this line](https://github.com/hotosm/tasking-manager/blob/develop/client/index.html#L118) is supposed to be showing an "X" button, and the functionality for dismissing the banner is all in place already. I think a style bug is meaning that it's not showing up.

pgiraud:
@harry-wood it looks like the CSS rule `hot-ds-icon-xmark` is never defined.
And unfortunately this kind of commit messages doesn't really help https://github.com/hotosm/tasking-manager/pull/413

pgiraud:
It looks like the different css rules prefixed with `hot-ds-` (actually used in several places in the code) aren't declared anymore. They were previously retrieved from https://github.com/hotosm/hot-design-system but this is not the case anymore.
https://github.com/hotosm/tasking-manager/commit/164f58cbaf9d59f3c3e8c6c57fcb26e5791a6925#diff-05d76e6130fce41e9f34b6a1aa08bd50L18

@felskia Should we replace the hot-ds-xxx classes with something else?

ethan-nelson:
For reference, here are the remaining hot-ds classes in case someone is interested in working on those: 

```
server/web/static/dist/app/components/account-nav/account-nav.html:    <i class="hot-ds-icon-sm-xmark pull-right pointer" ng-click="accountNavCtrl.hideNotification()"></i>
server/web/static/dist/app/project/project.html:                                           class="hot-ds-icon-circle-question pointer">
server/web/static/dist/app/project/project.html:                                           class="hot-ds-icon-circle-question pointer">
[...]

PR #946, opened 2017-11-12
Fix mobile banner closer

### LINKED ISSUE #820: TM3: Minimum page width (1033 pixels) is largely excessive and prevents too many users using TM
The minimum display width to use TM3 is excessive: 1033 pixels. At 1032 pixels we get an invitation to use Mapswipe instead.

Once again this is a issue of bad layout design. There's ample space to use TM3 on a desktop even when we want the browser only on half the width of the display, beside another window. It's even impossible to use it with some desktop if we have the developer console opened in the browser. The whole flowing of the layout is to rethink to be more versatile (think about flexible CSS and HTML layout designs).

I do not see really why TM3 should not work on a smartphone (with a flexible layout) even in portrait mode (about 640px wide should be enough: use tabs or similar to select what blocks to show seletively and be careful about encoding order of display blocks, use sliding menus if needed).

Even the tiles map should be renderable in that restricted width (provided you don't use the whole height and allow carefule sizing of titles, and don't waste excessive margins, and keep space at bottom for textual details, or a separate selectable tab (for the possibly long history and so on).

Avoid scrolling as much as possible, except for arbitrarily long lists such as the history or comments).

This was not so much a problem with TM2. So I see this as a regression bug.

### LINKED ISSUE #933: Enable a way to dismiss the MapSwipe banner
There doesn't seem to be a way to get rid of the MapSwipe banner when viewing the TM on mobile devices or at a smaller screen resolution (e.g. a resized window on a laptop). We had this issue crop up at the mapathon tonight where a low-res projector effectively made the TM think it's open on a mobile device and the MapSwipe banner was obscuring the actual task window.

ethan-nelson:
This PR replaces a deprecated css style to close the mobile banner with a style currently present in the code base.

Fixes #933 
Fixes #820 

modal

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -115,7 +115,7 @@
 <!-- mobile message -->
 <div class="mob-disclaimer" ng-show="!hideMobileMessage">
     <p>
-        <i class="hot-ds-icon-xmark pull-right pointer" ng-click="hideMobileMessage = true"></i>
+        <i class="modal__button-dismiss pull-right pointer" ng-click="hideMobileMessage = true"></i>
         {{ 'In order to make full use of the many features in the Tasking Manager, please use a desktop or laptop computer rather than a mobile device. If you would like to contribute on a mobile device, please use MapSwipe instead.' | translate }}
     </p>
     <a href="https://mapswipe.org" class="button button--achromic">{{ 'Go to MapSwipe' | translate }}</a>


ethan-nelson:
The TM on a small width sans banner:
postclosure

harry-wood:
Guess this also fixes https://github.com/hotosm/tasking-manager/issues/673

PR #957, opened 2017-11-15
Merge a few front end changes to master

ethan-nelson:
A few changes to sync between develop and master:

* Prevent failure in project creation file upload simply because the file extension is not all lowercase; instead compare it in a case-agnostic way.
* Adjust styling on the Mapswipe mobile banner to display the already-implemented close option.
* Update some translations from the codebase state at launch time (translations for changes since will come later).


ethan-nelson:
@david-hotosm can you confirm that all we need to do to make changes go live is merge from develop to master? That there's nothing else special required?

cc: @bgirardot

PR #960, opened 2017-11-16
Added new language Malgasy

pgiraud:


ethan-nelson:
I get a 404 error in console when switching to this language since there is no translation file. Not sure if that's a big deal, but perhaps we can touch an empty file for now? I'm not sure what the turnaround will be on translated strings.

pgiraud:
Empty language file added.

PR #979, opened 2017-11-30
Fix notification close

ethan-nelson:
The notification window previously could not be closed because it was referencing a deprecated style icon (similar to the MapSwipe banner). This fixes that:

mod

### DIFF
--- a/client/app/components/account-nav/account-nav.html
+++ b/client/app/components/account-nav/account-nav.html
@@ -59,6 +59,6 @@
     </div>
 </a>
 <div ng-show="accountNavCtrl.userMessages.newMessages && accountNavCtrl.newMessageNotification" class="unread-messages-popover">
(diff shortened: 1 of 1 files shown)


ethan-nelson:
If someone has an idea for moving the `x` elsewhere, I'm all ears. This was as kludgy of a fix as it gets.

bgirardot:
@ethan-nelson I agree on the placement of the X. That is why I have not approved this yet.

But, I think we should not work on this one anymore, someone is going to make a pr to fix this sometime soon and they have the improved placement of the X so let us wait for their pr and I am going to close this one in the interests of reducing the list of outstanding PRs since I think we are not going to merge it. 

If we do not get a PR to fix it in a slighlty improved way, we can reopen this one.
Thank you ethan! hope this is ok with you.

PR #982, opened 2017-12-06
Provide access limitation capability

ethan-nelson:
In some cases, it may be necessary to limit the ability for a user to interact with the tasking manager. This adds that functionality.


ethan-nelson:
Currently deployed to http://tasks-demo.hotosm.org

bgirardot:
@ethan-nelson Thank you very much for knocking this out, very rarely needed, but very important when it is needed.

Issue #641, opened 2017-06-28
Using the TM for Community Building


rachblevine:
I wanted to suggest 2 things:
1) We stop archiving our tasks.  We've accomplished so much, and hiding that doesn't help.  This is the best way for local groups or even random volunteers, to be aware of the work we've done in a community, so why not make it easier for them?
2) Have a logged in- people counter on the main TM page.  Knowing how many others are helping out at any random time can only help boost our sense on community!
Thanks!

hunt3ri:
Hi @rachblevine this came up on the slack channel yesterday.  I'm proposing the following solution, let me know what you think

![image](https://user-images.githubusercontent.com/1523510/32885551-b9f510e0-cab5-11e7-9461-aaaed0e0d7a3.png)

The text is just suggested by me, and can be easily altered.  cc @russdeffner @ethan-nelson @pgiraud  @dalekunce

smit1678:
I like the idea. I know your mockup isn't the final solution, but in terms of design and placement of the data, I might suggest a fully separate section in order to help keep the text hierarchy and flow. Perhaps something along the lines of [OAM](https://openaerialmap.org/), but styled and a better layout to fit the numbers we want to show for the TM: 

![openaerialmap 2017-11-16 10-09-34](https://user-images.githubusercontent.com/796838/32885772-681765f6-cab6-11e7-87e0-e31fb5551f4d.png)

hunt3ri:
Thanks @smit1678 I hear you - although you're pushing my v. limited design skills to the limit 😉 Consistency is good, and your design will be easier to implement on page.  Second draft (colours/fonts to be tweaked by an actual designer)

![image](https://user-images.githubusercontent.com/1523510/32887638-7f19f880-cabc-11e7-8c79-15aef6783401.png)

rachblevine:
I love it! I think having it down at the bottom there on the homepage will be a wonderful welcome for our new mappers.  Thanks so much for following up on this.

Rachel Levine
GIS Operations and Training Coordinator
International Services | American Red Cross
202-536-7535 | skype: rachblevine

hunt3ri:
Hi @felskia 

I've got a rough and ready version of the homepage stats working on the branch called **homepage-stats**.  It would be awesome if you could find some time to style it, as per discussion above.

[...]

PR #993, opened 2017-12-13
Homepage stats, community engagement.

hunt3ri:
This PR shows stats on the homepage as per the discussion on #641 .  

Thanks @LindaAlblas for helping with styling and @rachblevine  for the initial idea and @smit1678 for look and feel idea

Points to note, stats are being cached for 30 seconds so counts will not change until 30 seconds complete.  Query is slightly intensive so will cause a slower page load on request that regenerates counts, but I think it's worth it.

Any questions give me a shout.


client/app/home/home.controller.js
```diff
@@ -0,0 +1,39 @@
...
+    function homeController(statsService) {
+        var vm = this;
+        vm.mappersOnline = 0;
+        vm.tasksMapped = 0;
+        vm.totalMappers = 0;
+
+        activate();
+
+        function activate() {
+            console.log('iain');
```
ethan-nelson: Easter egg!

client/taskingmanager.config.json
```diff
@@ -1,7 +1,7 @@
 {
   "development": {
     "EnvironmentConfig":{
-      "tmAPI": "http://tasking-manager-staging.eu-west-1.elasticbeanstalk.com/api/v1",
+      "tmAPI": "http://localhost:5000/api/v1",
```
ethan-nelson: I think you may have left this committed while testing locally. (I have to edit these as well at times, but I think if we change the hard coded urls it may wreak havoc on the deployment process.)

ethan-nelson:
Awesome work! This looks slick and functions as expected. I left a few comments that are just minor housekeeping.

A question I do have is whether there is an easy-to-implement way to hide the counts until the stats are loaded (e.g. could they be blank strings that are then populated with the stats result?). This is perhaps more noticeable on a slow connection, or if you are running your database on an old 5k rpm hard drive, but the site shows `0 total mappers` while the page has initially loaded and the queries are running--and it can be upwards of a few seconds at times. This is perhaps unwarranted but I fear the hypothetical that a potential mapper or funder may think 'oh no they have no mappers' and close the homepage before diving further.

client/app/home/home.html
```diff
@@ -8,6 +8,32 @@ <h1 class="section__title">
...
 </header>
+<section class="section__body--secondary--dark">
+    <div class="homepage-stats-container row--centered">
+        <div class="inner">
+            <ul class="homepage-stats__cards">
+                <li id="homepage-stats-1" class="homepage-stats__card">
+                    <div class="stats-card__amount">
+                        <strong>{{ homeCtrl.mappersOnline | number }}</strong>
+                    </div>
+                    <div class="stats-card__title">Mappers Online</div>
```
ethan-nelson: Could we make this and the other two card titles translatable?

client/taskingmanager.config.json
hunt3ri: This is the test config, to our staging environment at thinkwhere which I will shortly be pulling down, so suggest we point this to your demo environment.

This won't impact deployment, as deployment uses the prod config

hunt3ri:
Thanks for the feedback @ethan-nelson.  I hear you about all counters at 0 before they update.  I think I could call the stat method at bootup time to ensure they are properly initialised.  Will try to update soon.  Have a good Christmas

client/taskingmanager.config.json
ethan-nelson: Good to know, thanks!

hunt3ri: As mentioned previously, our staging environment has been removed, so I think localhost is most sensible default for now.

client/app/home/home.html
hunt3ri: @ethan-nelson I think this work would be better done as a seperate PR.

I did take took at how I might generate the transifex keys but there's no good docs on the process.  @LindaAlblas @bgirardot can you describe the process for generating the transifex keys?

client/app/home/home.controller.js
hunt3ri: Fixed

hunt3ri:
@ethan-nelson ready to review again.  Fixed all issues you raised with exception of translation which I think would make sense to do as a seperate PR.  Let me know what you think?

ethan-nelson:
I will take a look this evening! Cheers!

client/app/home/home.html
ethan-nelson: To make it translatable, the text would be `{{ 'Mappers Online' | translate }}` instead of `Mappers Online`. But yes we can definitely do that in another PR.

hunt3ri:
Thanks @ethan-nelson 👍

PR #999, opened 2017-12-19
Adds file download option for tasks-as-osm-xml api

bgirardot:


bgirardot:
@ethan-nelson This should be a pretty quick review for you. I just copy pasted and changed the file name extension from the tasks-as-gpx api you fixed up.

PR #1005, opened 2017-12-25
Allow project managers to map archived projects

ethan-nelson:
This provides project managers the ability to mark tiles as mapped or unmapped for projects that are either draft or archived. 

Fixes part of #866. This is a sibling to the #920 PR.

### DIFF
--- a/server/services/project_service.py
+++ b/server/services/project_service.py
@@ -97,7 +97,7 @@ class ProjectService:
 
         project = ProjectService.get_project_by_id(project_id)
 
-        if ProjectStatus(project.status) != ProjectStatus.PUBLISHED:
+        if ProjectStatus(project.status) != ProjectStatus.PUBLISHED and not UserService.is_user_a_project_manager(user_id):
             return False, MappingNotAllowed.PROJECT_NOT_PUBLISHED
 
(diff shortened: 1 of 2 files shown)


bgirardot:
@ethan-nelson I will give it a test now. I see there are a few tests that failed in shippable.

One looks expected and needs refinement I guess, there is a test to make sure a non published project can not be mapped. I guess that needs refinement to make sure it can not be mapped by mappers, but can be mapped by project managers.

thank you very much ethan!!

bgirardot:
@ethan-nelson It seems to allow me as a non PM to validate an archived project? Does that seem possible? It did not let me map, correctly gave me 'PROJECT_NOT_PUBLISHED' error message, but seemed to allow validation.

The same for draft status, can not map, but can validate as a regular user. 

I think the correct behavior should be only PMs can map or validate, draft or archived projects @russdeffner does that sound right to you too?

ethan-nelson:
Hi Blake, thanks for the review. I will take a look at the failing tests and refine as necessary. The validation by non-PMs as you mention should be fixed through #920.

bgirardot:
@ethan-nelson I see this passes tests now. I will test locally again shortly.

PR #1010, opened 2017-12-27
Adds direct dl links for aoi and tasks api end points

bgirardot:


bgirardot:
Not super pretty, but functional for project managers. Hopefully ali can help make it look a little better in the new year.

2017-12-28_01h03_20

bgirardot:
@ethan-nelson Thank you very much, I will try that, the target blank addition.

PR #1013, opened 2018-01-05
Remove hard-coded swagger url

ethan-nelson:
Currently, the url that directs users to the Swagger API from `/api-docs` is hard-coded in the `server/config.py` file. This can present a problem if the app url is changed to another domain as users may be directed to the wrong Swagger API if the second url is not changed as well.

To mitigate this, the PR removes the hard-coded host and uses the `APP_BASE_URL` variable that is already used to define the first part of the url.

### DIFF
--- a/server/config.py
+++ b/server/config.py
@@ -42,7 +42,7 @@ class EnvironmentConfig:
 class ProdConfig(EnvironmentConfig):
     APP_BASE_URL = 'https://tasks.hotosm.org'
     API_DOCS_URL = f'{APP_BASE_URL}/api-docs/swagger-ui/index.html?' + \
-                   'url=https://tasks.hotosm.org/api/docs'
+                   f'url={APP_BASE_URL}/api/docs'
     LOG_DIR = '/var/log/tasking-manager-logs'
     LOG_LEVEL = logging.DEBUG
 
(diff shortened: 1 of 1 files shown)


bgirardot:
@ethan-nelson It looks like it needs the "f" at the start of the continued line.

> API_DOCS_URL = f'{APP_BASE_URL}/api-docs/swagger-ui/index.html?' + \
>                    f'url={APP_BASE_URL}/api/docs'

ethan-nelson:
Ah yes sorry. That's what I get for fixing flake8's too long of lines and not testing it afterwards!

Fix coming soon.

ethan-nelson:
Hm so it looks like we enabled CircleCI on forked PRs (:+1:) but some config is a bit messed up such that the database can't be accessed for tests. Looking through the config file...

smit1678:
@ethan-nelson yes, flipped the switch yesterday after talking to @bgirardot. I didn't see any env variables within CircleCI related to the database so I suspected that this wouldn't be an issue. Can look through as well to see how we can pass the right variables.

ethan-nelson:
I think this may be the problem: we have it set to not pass the secret CircleCI environmental variables to forked branch builds for security reasons and that's all good. But one of those variables is `SHIPPABLE=TRUE` and the code is looking for the `SHIPPABLE` environmental variable to determine whether to skip database tests or not (e.g. [[1]](https://github.com/hotosm/tasking-manager/blob/develop/tests/server/integration/services/test_mapping_service.py#L19) [[2]](https://github.com/hotosm/tasking-manager/blob/develop/tests/server/integration/services/test_validation_service.py#L16)).

CircleCI _does_ set the `CI` variable to true when building (see https://circleci.com/gh/hotosm/tasking-manager/745 and click on `Spin up Environment`) and that seems like a variable that could be used across multiple CI platforms (for example, if someone is using Travis). So, I'd propose we switch from using the `SHIPPABLE` environmental variable to `CI` instead in those test checks. I can get that assembled in a PR soon since we won't be able to merge without passing CI.

PR #1015, opened 2018-01-07
Prevent file caching

ethan-nelson:
Flask's [default](http://flask.pocoo.org/docs/0.12/api/#flask.Flask.send_file_max_age_default) is to set a 12 hour file cache header. This PR sets it to 0 so that users can immediately access new information from API file downloads.

### DIFF
--- a/server/config.py
+++ b/server/config.py
@@ -19,6 +19,7 @@ class EnvironmentConfig:
         'access_token_url': 'https://www.openstreetmap.org/oauth/access_token',
         'authorize_url': 'https://www.openstreetmap.org/oauth/authorize'
     }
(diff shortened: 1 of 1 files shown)


ethan-nelson:
e.g.  
```
{
  "access-control-allow-origin": "*",
  "cache-control": "public, max-age=0",
  "connection": "close",
  "content-disposition": "attachment; filename=3457.geoJSON",
  "content-type": "application/json",
  "date": "Sun, 07 Jan 2018 18:19:35 GMT",
  "expires": "Sun, 07 Jan 2018 18:19:35 GMT",
  "server": "Werkzeug/0.12.2 Python/3.6.1"
}
```

PR #1020, opened 2018-01-12
Update env var CI flag

ethan-nelson:
See https://github.com/hotosm/tasking-manager/pull/1013#issuecomment-356638724 for background.

Since we are relying only on CircleCI now, we need to rely on another environmental variable to flag whether it's a CI build or not. (As an aside, developers should be running tests locally as well to ensure tests that require database access pass.)

The behavior that this fixes only surfaces on PRs from branches of forks because the CircleCI secure variables (including `SHIPPABLE=true` that we manually set) are not passed into those for security reasons. But, `CI` and `CIRCLECI` are set to `true` in the builds so we can use that instead. And we settled on `CI` to generalize it in case of future CI platform changes or for users using other platforms.

### DIFF
--- a/tests/server/integration/models/postgis/test_project.py
+++ b/tests/server/integration/models/postgis/test_project.py
@@ -17,7 +17,7 @@ class TestProject(unittest.TestCase):
     def setUpClass(cls):
         env = os.getenv('CI', 'false')
 
-        # Firewall rules mean we can't hit Postgres from Shippable so we have to skip them in the CI build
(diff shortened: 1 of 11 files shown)


ethan-nelson:
Another aside: CircleCI does have database support if we want to remove this restriction in the future (https://circleci.com/docs/2.0/postgres-config/).

ethan-nelson:
@bgirardot if you don't mind approving this to get it merged in so I can rerun CI on some of the other PRs.

PR #1033, opened 2018-01-25
Revert init of homepage stats in manage.py

### LINKED ISSUE #1032: Unable to init database
When running `manage.py db upgrade` on a new empty database I get the error `relation "tasks" does not exist`

Looks like it was introduced here https://github.com/hotosm/tasking-manager/commit/0ed20f22e4e36fe995cc825f55384017972dc13f#diff-632b651e5579af0e928ed63482e9bc77

After commenting out `init_counters()` I was able to init the db
https://github.com/hotosm/tasking-manager/blob/develop/manage.py#L20

ethan-nelson:
When starting with a fresh database that you plan to upgrade with alembic, the initialization of the homepage stats (eliminated in this PR) will cause everything to fail. For now, we can remove the init and rethink how we want to do that. This won't affect the stats displaying--it will just not run the query on boot to have the stats already cached.

Fixes #1032

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -3,7 +3,7 @@ import base64
 from flask_migrate import MigrateCommand
 from flask_script import Manager
 
-from server import create_app, init_counters
+from server import create_app
 from server.services.users.authentication_service import AuthenticationService
 from server.services.users.user_service import UserService
 
@@ -17,7 +17,6 @@ for key in ['TM_DB','TM_SECRET','TM_CONSUMER_KEY','TM_CONSUMER_SECRET','TM_ENV']
 
 # Initialise the flask app object
 application = create_app()
-init_counters(application)
 manager = Manager(application)
 
 


bgirardot:
Can we not just wrap that line in a test for the command line args?

pseudo code:

If command line arguments do not = 'db upgrade'
do init stats

ethan-nelson:
That could be another way to do it for sure. I would just expand from my original text, though, that this throws a bunch of error text regardless of what you are running on manage.py _unless_ the database has been initialized (e.g. `runserver`). I don't know if that would be more graceful of an error than when someone tries to access the db through the app and it fails then.

hunt3ri:
:( Can we not just catch an exception here and swallow it.  Error on creation of a completely a new database is the definition of an edge case @ethan-nelson

ethan-nelson:
Alright, I'll go ahead and close this one

hunt3ri:
Thanks @ethan-nelson although you're putting the pressure on me now to actually do the PR 😉

ethan-nelson:
No sweat--I can fix this one this evening otherwise.

PR #1036, opened 2018-01-27
Add fix for AWS rate limiting when sending messages to all

### LINKED ISSUE #961: Message all contributors not working
I have not been able to message all contributors for any of the Iran Earthquake 2017 projects.

hunt3ri:
This hopefully fixes #961 - my suspicion is we're hitting AWS Rate Limits which are set at 14 emails per second.  So a a large project can easily have 50+ contributors - so when we can the AWS API to send all 50, Amazon will time us out.

This PR add a 0.5 seconds sleep after every 5 emails.  (Reduced number in case other emails being sent at same time)

### DIFF
--- a/server/services/messaging/message_service.py
+++ b/server/services/messaging/message_service.py
@@ -1,4 +1,5 @@
 import re
+import time
 
 from cachetools import TTLCache, cached
 from typing import List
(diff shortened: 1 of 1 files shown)


bgirardot:
@hunt3ri Thank you very much @hunt3ri ! (hard one to test, I think we will just have to merge it in and give it a test on production, no where else do we have dozens of email addresses to test with)

hunt3ri:
@bgirardot re testing - yes I think you can only find out in production.  It's a low risk change (given send all isn't working anyway 😉 ).  However, if the problem remains we can revert it later.  Fingers crossed

Issue #1031, opened 2018-01-24
Stuck in login loop


ethan-nelson:
We've received a few reports about this--and I just got stuck in it a bit ago--where you click on a link to the TM and are asked to log in, you log in, and then are redirected back to the login page.

I think this occurs if your session auth token has expired and you click on a link to a specific project or task. By default, the JS will redirect you to the page asking you to login because of the expired auth. But it seems like the "redirect back to" page is taken to be the login page as opposed to that specific project or task, so once you login you are redirected back to that.

This will require some more digging to pinpoint the problem exactly.

ethan-nelson:
After encountering this a few times, I think I have figured it out. Scenario:

1) Someone has previously logged in but their token is expired
2) They are presented with the log in or return screen (https://tasks.hotosm.org/login)
3) Though there is a login button in the body of the page, they click login at the top right
4) The login process is started and they log in as normal. However, the login button on the top right of the screen set the return url once logged in to https://tasks.hotosm.org/login since that is where they originated.

PR #1037, opened 2018-01-29
Use flask global object to store auth details for lifetime of request.

hunt3ri:
Hi @ethan-nelson on a seperate project I realised my assumptions about the scope of variables during the lifetime of a request were incorrect, meaning variables were persisting across multiple requests.

This PR moves the storage of authentication details into the flask.g global object that ensures they have a lifetime of only one request.  This is how flask recommends you store these sort of temp request specific data.  More info is here http://flask.pocoo.org/docs/0.12/api/#application-globals

This change may improve the login loop issue in #1031 , although I think that may be more of a JS problem.

I also extended the time of checking for new messages to 30 seconds, as that request is generating a lot of noise in the console.

### DIFF
--- a/client/app/components/account-nav/account-nav.directive.js
+++ b/client/app/components/account-nav/account-nav.directive.js
@@ -60,7 +60,7 @@
                 if (vm.account.username) {
                     checkIfUserHasMessages();
                 }
-            }, 10000);
+            }, 30000);
 
             // Catch clicks and check if it was outside of the menu element. If so, close the dropdown menu.
             $document.bind('click', function(event){
--- a/server/api/grid/split_task_apis.py
+++ b/server/api/grid/split_task_apis.py
@@ -4,7 +4,7 @@ from schematics.exceptions import DataError
 from server.models.dtos.grid_dto import SplitTaskDTO
 from server.models.postgis.utils import NotFound
 from server.services.grid.split_service import SplitService, SplitServiceError
-from server.services.users.authentication_service import token_auth, tm
+from server.services.users.authentication_service import token_auth, tm, who_made_request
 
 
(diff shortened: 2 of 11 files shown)


ethan-nelson:
Hey Iain, thanks for the PR! It's great you were able to find this (I haven't dug into Flask auth much) and is a neat addition. (Ditto on thoughts that JS is responsible for the loop.)

Just FYI I am in the middle of moving, so I won't be able to review this or the other PR until later in the week. Apologies for that delay--though others are welcome to review if available!

hunt3ri:
Hey @ethan-nelson no problem, I think this will definitely improve reliability of login.  Happy for anyone else to review, but I'm sure it can wait a few days.  Good luck with your move!!

ethan-nelson:
Hey Iain,

I had some time to test this, but ran into a problem:

```
Task GET API - unhandled error: '_AppCtxGlobals' object has no attribute 'is_pm_only_resource'
```

I think it may be stemming from the util.py changes, but I'm still trying to digest that file so I can't say for sure :).

hunt3ri:
Thanks @ethan-nelson, it definitely will be.  I thought I'd caught all issues, however, clearly not.  Can you provide the API you were using and the stack trace to pinpoint the bug

ethan-nelson:
I can get that to you tonight. It occurred when trying to select a task on a project that was already loaded, so the GET /api/v1/project/{project_id}/task/{task_id} endpoint.

hunt3ri:
Cool thanks, my suspicion is that variable isn't initialised on unauthenticated routes, I think the original version defaulted the var to False.  Just need to add a bit of handling for that scenario.

willemarcel:
It was already implemented with other solution.

Issue #1018, opened 2018-01-12
my task "box" isn't a box


govvin:
The tasks we generate aren't box-shaped, so the current warning seen on iD isn't exactly suitable. It's not a major issue, just a niggling description.

Perhaps change the text to something like: `Do not edit outside this colored area` ?

![image](https://user-images.githubusercontent.com/5821688/34855420-9813bd7c-f77a-11e7-88b9-c65bc35eee56.png)

bgirardot:
This was fixed by the merge listed above. Thank you for bug report!

PR #1048, opened 2018-02-17
change the task gpx boundary label

fiennyangeln:
#1018

### DIFF
--- a/server/services/mapping_service.py
+++ b/server/services/mapping_service.py
@@ -180,7 +180,7 @@ class MappingService:
 
                     # Append wpt elements to end of doc
                     wpt = ET.Element('wpt', attrib=dict(lon=str(point[0]), lat=str(point[1])))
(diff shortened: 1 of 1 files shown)


ethan-nelson:
Thank you for your contribution @fiennyangeln !

bgirardot:
@sunidhiraheja I apologize, I have to suggest reverting this merge into develop because I think this is incomplete.

The text string you changed is for the text on the nodes, but there is also the same text that appears in the middle of a task square somehow that still says "box" please see below:

2018-03-14_22h36_06

If you could find where that additional mention of "box" is and change it, then everything will be good with this pr.

fiennyangeln:
@bgirardot I think it is originally that way on my PR. It is becauseon the PR i made based on the issue, I just change the label for boundary, not for those standing in the middle. Do you want me to change the title for the middle as well?

Issue #952, opened 2017-11-14
Improve validation email subject line to indicate project


bgirardot:
![2017-11-14_04h36_29](https://user-images.githubusercontent.com/265831/32761763-721e03d8-c8f5-11e7-8819-99e43c890d72.png)

mnoblecourt:
Hi,
Could we had an option "not receive messages when your task is validated" (unless there's a comment, like TMv2) to avoid advanced mappers being spammed? For some of our volunteers, they receive so many of these automatic messages (which need to be deleted manually!) that they don't even check their inbox anymore, and miss important communications.
Thanks!
Martin

fiennyangeln:
@bgirardot Hi I wanna work on this issue. Can I add [ProjectID] in the beginning of the subject?

bgirardot:
How about:

"Your mapping in Project nnnn, on Task nnn has just been validated"

@fiennyangeln Thank you for working on this!

ethan-nelson:
I believe the original issue (as well as the secondary regarding opt out) has been fixed now in #1051. Please reopen if that is not the case--thanks!

Issue #553, opened 2017-06-12
Messages only mention task number


bgirardot:
Messages when mentioned on a task only mention the task number, but not the project title and/or number.

Would be better if it said you were mentioned on Project number 23, Marawi Mapping, task 41

fiennyangeln:
Hi I want to take this issue.

ethan-nelson:
This should be fixed with #1051. Please reopen if that is not the case.

PR #1049, opened 2018-02-18
add message task info

fiennyangeln:
#553

PR #1051, opened 2018-02-19
Improve Task comment and validation message subject lines

fiennyangeln:
fixes issue #553 and fixes issue #952

### DIFF
--- a/server/models/postgis/project.py
+++ b/server/models/postgis/project.py
@@ -306,6 +306,10 @@ class Project(db.Model):
 
         return summary
 
+    def get_project_title(self, preferred_locale):
+        project_info = ProjectInfo.get_dto_for_locale(self.id, preferred_locale, self.default_locale)
+        return project_info.name
+
     def get_aoi_geometry_as_geojson(self):
         """ Helper which returns the AOI geometry as a geojson object """
         aoi_geojson = db.engine.execute(self.geometry.ST_AsGeoJSON()).scalar()
--- a/server/services/messaging/message_service.py
+++ b/server/services/messaging/message_service.py
@@ -104,7 +104,7 @@ class MessageService:
             message = Message()
             message.from_user_id = comment_from
             message.to_user_id = user.id
-            message.subject = f'You were mentioned in a comment on {link}, {project_title}, Task {task_id}'
+            message.subject = f'You were mentioned in a comment in Project {project_id}, on Task {task_id}'
             message.message = comment
             message.add_message()
             SMTPService.send_email_alert(user.email_address, user.username)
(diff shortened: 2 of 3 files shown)


bgirardot:
@fiennyangeln Hi, thank you very much for the pull request.

A couple of things:

1. I might not have been clear enough, but when I wrote "Project nnnn" I meant the word "Project" and then the project number, like 4326. e.g., "You mapping in Project 4326, on Task 251 has been validated"

I am worried about long project titles in message subject lines, but I will do some testing to see how they look, but I think we might want to stick with project number instead of title.

2. This pull request adds the project title to the subject line of a message left on a comment on a task, which might be needed, I will look at that as well.

But the issue you linked to and are trying to fix is the subject of a message that gets sent after a task is validated.

Basically any message should subject should have the project number or project title in the subject line, but in this case it was about the messages that get sent when a task is marked as validated.

I will keep testing this PR as it might be fine as is, but it does not address issue #952

bgirardot:
@fiennyangeln I should mention, I am not much of a programmer so I can do testing and such, but unfortunately, I can't offer advice on even where to look to fix this error. I am hoping you know more about it than I do :)

bgirardot:
Oh, I see, this is supposed to fix issue #553 I am going to update the pr title then as it is not the pr for the validation subject, but it is essentially the same issue as the validation message issue, just this message is generated for any comment directed to a user in a task comment box, and the other issue is for messages that are auto generated when a task is marked "validated".

bgirardot:
@fiennyangeln This appears to just be a duplicate of #1049

fiennyangeln:
Hi! I also see that error but I didnt change anything inside the Mapping, so I initially think that the error was there from the very start. But sure I will try look at it :)

fiennyangeln:
So i decided to open pull request and do the testing via the circleci instead

fiennyangeln:
Hi I fixed this one as well. Pls check :)

bgirardot:
Hi, that fixed the hard error.

But it seems that in the original code the {link} call had the "Task nnnn" text built into it so now your PR shows the task number twice:

2018-02-20_06h33_17

I also think that we do not want the full project title in the subject, it just gets too long that way. Maybe it could in the body of the message? 

I would keep what you added to get the project title I think, that might be useful somewhere else.

But for the subject line of the messages, please have it in the format I indicated on the issue related to validation email:

"You were mentioned in a comment in Project 4323, on Task 43"

cheers
blake

fiennyangeln:
Ok I have removed the project title from the subject.

bgirardot:
@ethan-nelson This is another one I think is ready to merge, it works based on my testing, just could use a quick look over the code itself.

There is a function added that ended up not actually being used, but I suggested leaving it in so we could make use of it in the future, for example if we decide to change the text of the email to include the project title.

bgirardot:
@fiennyangeln I am embarrassed to say I found an problem with the PR even after I merged it into the develop branch.

When some one sends a message to a user from a comment on a Task, the is no link to the task in the message anywhere. There is in the current code base.

2018-03-14_23h20_14

2018-03-14_23h14_26

2018-03-14_23h17_47

fiennyangeln:
@bgirardot sorry for my carelessness. About the message is there anything need to be added? I will do another PR for this

PR #1052, opened 2018-02-20
add userId

### LINKED ISSUE #975: Add OSM `uid` to API replies
When requesting information about a user (for example in [/users/search-all](https://tasks.hotosm.org/api/v1/user/search-all)), it would be useful to have their OSM id alongside their username. This would help join API replies to other OSM information.

fiennyangeln:
fixes #975

### DIFF
--- a/server/models/dtos/user_dto.py
+++ b/server/models/dtos/user_dto.py
@@ -1,6 +1,6 @@
 from schematics import Model
 from schematics.exceptions import ValidationError
-from schematics.types import StringType, IntType, EmailType
+from schematics.types import StringType, IntType, EmailType, LongType
 from schematics.types.compound import ListType, ModelType, BaseType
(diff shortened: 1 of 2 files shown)


bgirardot:
@kamicut Hi Marc, is there any chance you can test this locally or at least review the code to see if it meets your needs? 

It is not a use case I understand so I can not tell.

@ethan-nelson Your feedback as well, does this fulfill the guidance you put on the original issue?

bgirardot:
@ethan-nelson thank you for reviewing, thank you @fiennyangeln for making the pr and @kamicut you can test this on https://tasks-stage.hotosm.org/ in a few minutes (hopefully :)

bgirardot:
@kamicut Oppss, might not actually add the uid to the api call, i see ethan says lays the groundwork for

Issue #977, opened 2017-11-29
Project Creation Organization Tag Autocomplete does not work


Nieldejong:
Me and my colleagues have published multiple projects in the name of the 'Netherlands Red Cross', but the tag has never been saved as a default. @russdeffner mentioned that the organization should be saved in the TM3 after the first entry, but that does not happen at the moment. Therefore, it prob is a bug.

If this is not a bug, could the Netherlands Red Cross organization be added manually? 

When searching through projects by organization, you can see that multiple organizations have encountered this problem. For example, both the Clinton Health Access Initiative and the Clinton Access Health Initiative have been entered and are searchable:

![image](https://user-images.githubusercontent.com/31616896/33383251-1bf4a4dc-d523-11e7-8732-933c3eea0c43.png)

verdy-p:
To create an organization tag in the project, so that you can select it later, you need to type its name, but DON'T press TAB or click outside, when the combobox opens its popup, press SHIFT+TAB which will keep you in the Combox editable field, but will place your cursor at start: the string will be transformed into a tag "button" and then you can save the edit, which will create the tag. Later, you'll be able to select this tag in any other project.
But I agree that we should have a better way to create an organization, not just as a simple tag, but as a real object which could have its own properties and profile, and its own list of sponsored TM projects

Nieldejong:
Hi @verdy-p, thank you for your prompt response. 

Until now we have tagged our projects by writing the full name and then clicking outside of the box. It is not the exact same way as you described, but we have managed to attach the right tag to our project. The problem was that afterward, the tag was not automatically saved, so it never came up in the drop-down menu for a next projects. Not a huge problem, but it does make the process prone to, for example, spelling mistakes. 
I just tried it your way, which was more convenient. But it did not solve the problem yet.

To show you an example, we have a couple of projects with the right tags (as in the print screen below), but the name 'Netherlands Red Cross' still does not come up automatically (also in the print screen). 

![image](https://user-images.githubusercontent.com/31616896/33437706-0e8fc1d4-d5e9-11e7-9070-565946d1f47f.png)

ethan-nelson:
[...]

PR #1054, opened 2018-02-21
fix autocomplete for organisation and campaign in editprojectadmin

fiennyangeln:
#977 
@bgirardot hi! please review the changes :) thx!

### DIFF
--- a/client/app/admin/edit-project/edit-project.controller.js
+++ b/client/app/admin/edit-project/edit-project.controller.js
@@ -430,8 +430,10 @@
          * Get organisation tags
          * @returns {Array|*}
          */
-        vm.getOrganisationTags = function(){
-            return vm.organisationTags;
+        vm.getOrganisationTags = function(query){
+            return vm.organisationTags.filter(function (item) {
+                return (item && item.indexOf(query) > -1);
+            });
         };
 
         /**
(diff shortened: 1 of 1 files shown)


bgirardot:
@fiennyangeln This looks like it works great!

If it is easy only, if not I will merge it in as is because it is basically exactly right as spec'ed out now, but if it is easy could it be case insensitive?

Right now to match Red Cross or British Red Cross, I have to type "Red" with the capital R, if I just type red, it does not match anything.

If that is not a quick easy fix, no problem adn I will merge this in, thank you very very much!

fiennyangeln:
I am currently trying to look at other solution to enable it. But it might make it slower. 

But I will first try to figure out why my environment suddenly not working today.

fiennyangeln:
hi @bgirardot please check:)

fiennyangeln:
@bgirardot Hi! I change this one to enable case insensitive. Please review :)

bgirardot:
@fiennyangeln Works great! Thank you very much.

@ethan-nelson Might get a chance to review the code, I am less worried about performance on this because it is not on the main Contribute page, this gets used a lot less often.

but this works much better now, pretty important fix, I would like to get it merged in asap.

PR #1055, opened 2018-02-22
Fix #986

### LINKED ISSUE #986: Project Admin - Validate all tasks messes up "Contributions" on Activity and Stats page
If you use the Validate all tasks button in the project admin, it seems to make the system think the person who clicked "Validate all tasks" is the person who mapped and validated every task as shown in the Activity and Stats page.

It makes sense that 'validate all tasks' should have some affect on the contributions field on activity and stats, but it needs to work differently.

2017-12-09_14h01_58

sunidhiraheja:
According to the reported issue, when a project admin 'A' clicked "Validate all tasks", the application marked tasks as having been mapped by 'A' regardless of whether or not he/she actually mapped them.

The flow of this operation is as follows:

- The frontend sends a `POST` request to `/admin/project/<int:project_id>/validate-all`
- The backend calls `ValidatorService.validate_all_tasks` passing the project ID and the ID of the authenticated user (in this case the project admin, NOT the mapper). This ID is necessary to record the user performing the current action (validation).
- However, `validate_all_tasks` also sets the `mapped_by` value of each task to this user ID.

Removing the line that does so fixes this issue, because we need not record/update the mapper ID here. It is already taken care of in `task.unlock_task:314`

### DIFF
--- a/server/services/validator_service.py
+++ b/server/services/validator_service.py
@@ -217,7 +217,7 @@ class ValidatorService:
                                               Task.task_status != TaskStatus.BADIMAGERY.value).all()
 
         for task in tasks_to_validate:
-            task.mapped_by = user_id  # Ensure we set mapped by value
+            task.mapped_by = task.mapped_by or user_id  # Ensure we set mapped by value
             if TaskStatus(task.task_status) not in [TaskStatus.LOCKED_FOR_MAPPING, TaskStatus.LOCKED_FOR_VALIDATION]:
                 # Only lock tasks that are not already locked to avoid double lock issue
                 task.lock_task_for_validating(user_id)
--- a/tests/server/integration/services/test_validation_service.py
+++ b/tests/server/integration/services/test_validation_service.py
@@ -77,5 +77,5 @@ class TestValidationService(unittest.TestCase):
         ValidatorService.validate_all_tasks(self.test_project.id, self.test_user.id)
 
         for task in self.test_project.tasks:
-            self.assertEqual(task.mapped_by, self.test_user.id)
+            self.assertIsNotNone(task.mapped_by)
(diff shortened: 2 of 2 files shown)


sunidhiraheja:
In addition, as pointed out by @pgadige, `ValidatorService.validate_all_tasks` was mapping all task types except those with status `BAD_IMAGERY`.
I think that a task should only be marked validated once it has been mapped, so that leaves us with the following task types: MAPPED, LOCKED_FOR_VALIDATION and INVALIDATED (not sure about this one, do we want a task mapped again after it has been invalidated?).

Open to suggestions, I think this needs some discussion.

sunidhiraheja:
For tasks that do not already have a `mapped_by` ID set, we set it to the project manager's ID.

sunidhiraheja:
Update: For tasks that do not already have a `mapped_by` ID set, we set it to the project manager's (the user clicking the button) ID.

bgirardot:
@sunidhiraheja Thank you very much for this, I can't test it however because it is not from a unique branch in your fork of this repo.

I am not sure what to reccommend to you, but if you could create a unique branch in your fork with these changes and make a new pull request that is the best way to handle this.

None of us here are git/github experts except pierre ;) and so always using a unique branch for a fix/feature is needed.

We call those 'feature' branches and they are what all code contributions should be on, a feature branch you create on your fork. Basically like all your other pull requests have been, those are exactly right. And if I knew how to fix this or work with it being on the develop branch I would, but I do not so if you could just close this one and make a new pull request from a feature branch that would really make this go faster :)

sorry for the bother.

sunidhiraheja:
Thanks for pointing that out, I should have used a feature branch.
It's up on https://github.com/hotosm/tasking-manager/pull/1059 now! :)

PR #1056, opened 2018-02-22
Fix #1041: Add mark all tasks as mapped button to project edit - actions

### LINKED ISSUE #1041: Mark all tasks as mapped/completely mapped
There's an option to validate/invalidate all tasks of a project, but not to mark all as done. There may be few times when this is needed to be used but I'd find it useful as a project manager.

screen shot 2018-02-01 at 9 01 30 am

sunidhiraheja:


bgirardot:
@sunidhiraheja I am testing this now.

bgirardot:
@sunidhiraheja Please see my comment on the original issue.

My initial testing shows this basically works, but we have to figure out the best way to handle the already 'validated' task squares.

My thinking is that if a task square is already Validated, it should stay that way, this functionality is more about marking non mapped tasks (ready or invalidated) as mapped and not about resetting validated tasks to a mapped state.

As it is now it resets everything to 'mapped' except bad imagery, which seems right, but then when looking at the stats on the project (Activity and Stats button on map) it still shows as 6% validated.

2018-02-23_15h43_12

bgirardot:
@sunidhiraheja Ok, the creator of the original issue agrees, if a task is already 'validated' it should stay that way and we are only concerned with unmapped tasks to mapped.

cheers, blake

sunidhiraheja:
Okay, so after my last commit, this is the current status of all TaskStatus types:

READY -> MAPPED
LOCKED_FOR_MAPPING -> MAPPED (now by person clicking the button)
MAPPED -> MAPPED (now by person clicking the button)
LOCKED_FOR_VALIDATION -> MAPPED (person validating loses the validation lock)
VALIDATED: unchanged
INVALIDATED -> MAPPED
BADIMAGERY: unchanged

Should there be any changes to this, particularly the MAPPED -> MAPPED case?

Thanks for pointing out the discrepancy in the stats bar, I think this is an existing problem which I could also reproduce with the Validate All button. I'll get to this in a couple of hours.

bgirardot:
@sunidhiraheja I think that is all right, except.....

> MAPPED -> MAPPED (now by person clicking the button)

That should not change who marked it MAPPED, it should stay with the original person who marked it mapped, not the person clicking the button.

Thank you again for working on this, it seems really close :)

sunidhiraheja:
Okay, mapped tasks are now not mapped again.

As for the stats bar, I found that the percentage values are cached for 600 seconds on the backend (in `server.services.project_service:12`).
This behavior is consistent across the validate_all, invalidate_all and the newly added map_all buttons.

sunidhiraheja:
Hi, any updates on this?

ethan-nelson:
Hi @sunidhiraheja, I made one minor change to fix the issue of already mapped tasks not being marked mapped again. If you have some time, would you mind merging the latest develop into this branch? We had some deployment issues with the Docker file that have been fixed (but cause the tests to fail if not merged). Otherwise I can try doing that to the branch. Thanks for your help--this works great and is a nice addition!

PR #1057, opened 2018-02-22
Fix #995: Add countdown timer to checked out task 

### LINKED ISSUE #995: Add timer to indicate time left on a checked out task
The two hour time out seems to work, but there is no UI to indicate the count down as in TM2.

sunidhiraheja:
image

This timer picks up the lock action time from the task history and updates itself using the angular moment directive `am-time-ago`.

fixes issue #995

### DIFF
--- a/client/app/project/project.controller.js
+++ b/client/app/project/project.controller.js
@@ -8,7 +8,7 @@
      */
     angular
         .module('taskingManager')
-        .controller('projectController', ['$timeout', '$interval', '$scope', '$location', '$routeParams', '$window', 'moment', 'configService', 'mapService', 'projectService', 'styleService', 'taskService', 'geospatialService', 'editorService', 'authService', 'accountService', 'userService', 'licenseService', 'messageService', 'drawService', 'languageService', 'userPreferencesService', 'moment', projectController]);
+        .controller('projectController', ['$timeout', '$interval', '$scope', '$location', '$routeParams', '$window', 'moment', 'configService', 'mapService', 'projectService', 'styleService', 'taskService', 'geospatialService', 'editorService', 'authService', 'accountService', 'userService', 'licenseService', 'messageService', 'drawService', 'languageService', 'userPreferencesService', projectController]);
 
     function projectController($timeout, $interval, $scope, $location, $routeParams, $window, moment, configService, mapService, projectService, styleService, taskService, geospatialService, editorService, authService, accountService, userService, licenseService, messageService, drawService, languageService, userPreferencesService) {
         var vm = this;
--- a/client/app/project/project.controller.js
+++ b/client/app/project/project.controller.js
@@ -8,10 +8,9 @@
      */
     angular
         .module('taskingManager')
-        .controller('projectController', ['$timeout', '$interval', '$scope', '$location', '$routeParams', '$window', 'configService', 'mapService', 'projectService', 'styleService', 'taskService', 'geospatialService', 'editorService', 'authService', 'accountService', 'userService', 'licenseService', 'messageService', 'drawService', 'languageService', 'userPreferencesService', projectController]);
+        .controller('projectController', ['$timeout', '$interval', '$scope', '$location', '$routeParams', '$window', 'moment', 'configService', 'mapService', 'projectService', 'styleService', 'taskService', 'geospatialService', 'editorService', 'authService', 'accountService', 'userService', 'licenseService', 'messageService', 'drawService', 'languageService', 'userPreferencesService', 'moment', projectController]);
 
-    function projectController($timeout, $interval, $scope, $location, $routeParams, $window, configService, mapService, projectService, styleService, taskService, geospatialService, editorService, authService, accountService, userService, licenseService, messageService, drawService, languageService, userPreferencesService) {
-
+    function projectController($timeout, $interval, $scope, $location, $routeParams, $window, moment, configService, mapService, projectService, styleService, taskService, geospatialService, editorService, authService, accountService, userService, licenseService, messageService, drawService, languageService, userPreferencesService) {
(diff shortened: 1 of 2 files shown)


bgirardot:
@sunidhiraheja I will download and test, thank you very much!

bgirardot:
When I try it, I get the following error (FF and Chrome on Ubuntu 16.04):

"Timer
If you do not complete or release this task in time, it will be automatically unlocked Invalid date"

It would also be better if it was just the time remaining up next to the Mapping header and then the text was a tool tip.

but lets get it working first where it is and we can worry about where to locate it on page after that.

Do I need to maybe install something to get this to work?

I did not see any obvious errors in the console related to this.

2018-02-23_15h16_18

sunidhiraheja:
Hi, Thanks for the review.
I got it work now, I had not taken into account the fact that multiple tasks can be locked at once (in case of validation).
Can you please verify? I'll reposition if so.

bgirardot:
I am still not having a lot of luck with this :(

When i check out for mapping I see this:

2018-02-25_06h37_07

It never changes and if I refresh the page I see this:

2018-02-25_07h12_43

If I checkout for validation, I see the same "invalid date" message.

The date, time, TZ on both my virtual machine and my host machine is set correctly.

Sorry I do not have better news.

sunidhiraheja:
I'm really sorry for that, I've tried something to fix it.

bgirardot:
hi @sunidhiraheja Your latest fix does fix the "2 hours ago" issue and says something like in 2 hours.

but it does not seem to count down at all, it just kept saying in 2 hours.

and when I refreshed the page, it just displayed 'invalid date' again.

sunidhiraheja:
The invalid date issue should be fixed now.
There were three end points where the lock times should be updated - on mapping locks, on validation locks and on page reloads (when no locks actually happen but the lock time records set while locking are lost)

The last one is what I was missing.

sunidhiraheja:
The updating of the countdown timer relies on how moment.js (the library used for similar purposes in the project) interprets a timestamp. 
The same mechanism is used on the project dashboard to show how long it has been since a project was created and last updated.

If we want a more verbose countdown timer, I'll try something else (probably setInterval based updates, which is what was used in tasking manager 2)

bgirardot:
@sunidhiraheja Testing this now. Yes, I think for sure we need a little more specific countdown like we had in tasking manager 2, see below:

2018-03-01_18h15_39

 That number basically counted down each minute.

I am about to create another issue that is related to all of this and revealed itself clearly in my testing. It might be an easy fix for you since you have such a good understanding of this at the moment.

sunidhiraheja:
@bgirardot It looks like this in the latest commit:

image

Can you please confirm if it works for you?

client/app/project/project.html
```diff
@@ -285,7 +285,11 @@ <h3>{{ 'Choose a task' | translate }}</h3>
                             <!-- task viewing and mapping UI's -->
                             <div ng-show="projectCtrl.mappingStep === 'viewing' || projectCtrl.mappingStep === 'locked'">
                                 <!-- task viewing UI -->
-                                <h3>{{ 'Mapping' | translate }}</h3>
+                                <h3 style="display: inline-block;">{{ 'Mapping' | translate }}</h3>
+                                <div ng-show="projectCtrl.mappingStep === 'locked'" style="display: inline-block; float: right;">
+                                    <i class="fa fa-clock-o"></i>
+                                    <span>{{ projectCtrl.getLockTime() }} minutes left</span>
```
ethan-nelson: Could you make `minutes left` a translatable string? (Same with L654 below)

client/app/project/project.controller.js
```diff
@@ -8,10 +8,9 @@
...
         .module('taskingManager')
-        .controller('projectController', ['$timeout', '$interval', '$scope', '$location', '$routeParams', '$window', 'configService', 'mapService', 'projectService', 'styleService', 'taskService', 'geospatialService', 'editorService', 'authService', 'accountService', 'userService', 'licenseService', 'messageService', 'drawService', 'languageService', 'userPreferencesService', projectController]);
-
-    function projectController($timeout, $interval, $scope, $location, $routeParams, $window, configService, mapService, projectService, styleService, taskService, geospatialService, editorService, authService, accountService, userService, licenseService, messageService, drawService, languageService, userPreferencesService) {
+        .controller('projectController', ['$timeout', '$interval', '$scope', '$location', '$routeParams', '$window', 'moment', 'configService', 'mapService', 'projectService', 'styleService', 'taskService', 'geospatialService', 'editorService', 'authService', 'accountService', 'userService', 'licenseService', 'messageService', 'drawService', 'languageService', 'userPreferencesService', 'moment', projectController]);
```
ethan-nelson: It looks like there is an extra `moment` added near the end of this definition.

ethan-nelson:
@sunidhiraheja this is awesome! I've checked out a few tasks and am letting them time out to double check nothing errors out. If you can fix the two minor comments I made, then this should be ready to go. Thanks!

ethan-nelson:
Sorry Sunidhi, one more thing. After the two hour limit, the timer starts to go negative since tasks don't automatically refresh. Would it be possible to have an if statement in the frontend so that if the number of minutes is less than 0, "automatically unlocked" is displayed in that timer area? I can work on adding that if you don't think you will have time to work on it soon.

sunidhiraheja:
@ethan-nelson Thanks for reviewing! I've pushed a new commit with those changes.
I had also noticed the counter going in reverse, but left it there because of the other pending PR. Sorry I forgot to mention it.

ethan-nelson:
@sunidhiraheja no problem--I should be more aware of the PRs already filed. Thank you for your help and changes!!

Issue #943, opened 2017-11-10
Duplicated organisation and meaningless campaign names in list


Jorieke:
For example 
Médecins Sans Frontières and MSF 
Clinton Health Access Initiative and Clinton Access Health Initiative

verdy-p:
See also bug #509

ethan-nelson:
This would be a good project for someone interested in diving into all of the components that make up the server backend. To implement a removal mechanism for tags, it will entail:

* Adding a method in the Postgres model for the tag table that will remove a given organization (https://github.com/hotosm/tasking-manager/blob/develop/server/models/postgis/tags.py)
* Adding a service function that 1) ensures no projects currently are assigned to a given organization then 2) calls that method to remove the organization (https://github.com/hotosm/tasking-manager/blob/develop/server/services/tags_service.py)
* Adding an API endpoint to call that service with a given tag (https://github.com/hotosm/tasking-manager/blob/master/server/api/tags_apis.py)

Also note that organizations and campaigns are part of the same table and have essentially duplicate calls, so once you have done this for an organization you can copy+paste the code again and swap out organization for campaign.

hunt3ri:
Hi folks, if there's community interest, I'd be happy to do a Google Hangout / screen share on this

Live coding, what could possibly go wrong 😉 

cc @ethan-nelson @pgiraud @smit1678 @bgirardot

verdy-p:
I think that the organization name and its tag will likely be often different (e.g. no whitespace in hashtags, or possibly abbreviations, but whitespace in real unabbreviated organization name, shown on its profile page). Tags tend to be shorter as they are easier to type (note how we can select user names or organization names page names in social networks: we also have a disambiguating description which is useful to display along with the tag matched, there's a prefered tag along other possible aliases all disambiguated by the added description shown in the list of search results, and a prefered name)

verdy-p:
Also, like users, organizations may have a registered icon in their profile, which can be shown when they talk by themselves or in lists or organization. These small icons would be smaller derivations from their logo (which is displayed more completely in their own profile page).

fiennyangeln:
hi i want to take this. is it still available?

bgirardot:
@fiennyangeln You are already working on this :)

This is directly related to your work on the campaign tags only showing tags that have project associated with them.

I can review this more, but the key is, if a campaign or organization has no projects (draft, archived or published), it should just disappear from the lists

bgirardot:
[...]

PR #1062, opened 2018-02-26
add api delete organisation tag

fiennyangeln:
#943 [WIP] I added api to delete organisation tag using post request since organisation tag can be long. If this works, I will try to implement similar things for campaign tag. Pls review :) @ethan-nelson @bgirardot

### DIFF
--- a/server/api/tags_apis.py
+++ b/server/api/tags_apis.py
@@ -1,5 +1,7 @@
-from flask_restful import Resource, current_app
-from server.services.tags_service import TagsService
+from flask_restful import Resource, current_app, request
+from server.services.tags_service import TagsService, TagsServiceError
+from server.models.postgis.utils import NotFound
+from server.services.users.authentication_service import token_auth
 
 
 class OrganisationTagsAPI(Resource):
@@ -23,6 +25,58 @@ class OrganisationTagsAPI(Resource):
             return tags.to_primitive(), 200
         except Exception as e:
             error_msg = f'User GET - unhandled error: {str(e)}'
+            current_app.logger.critical(error_msg)
+            return {"error": error_msg}, 500
+
+    @token_auth.login_required
+    def post(self):
+        """
+        Delete the specified mapping license
+        ---
+        tags:
+            - tags
+        produces:
+            - application/json
+        parameters:
+            - in: header
+              name: Authorization
+              description: Base64 encoded session token
+              required: true
+              type: string
+              default: Token sessionTokenHere==
+            - in: body
+              name: body
+              required: true
+              description: JSON object for deleting organisation tag
+              schema:
+                  properties:
+                      organisation_tag:
+                          type: string
+                          default: organisation
+        responses:
+            200:
+                description: Tag Deleted
+            401:
+                description: Unauthorized - Invalid credentials
+            403:
+                description: Organisation tag belongs to some project
+            404:
+                description: Organisation Tag Not Found
+            500:
+                description: Internal Server Error
+        """
+        try:
+            organisation_tag = request.get_json().get('organisation_tag')
+            print(organisation_tag)
+            TagsService.delete_organisation_tags(organisation_tag)
+            return {"Success": "Organisation Tag deleted"}, 200
+        except TagsServiceError as e:
+            error_msg = f'{str(e)}'
+            return {"Error": error_msg }, 403
+        except NotFound:
+            return {"Error": "Organisation Tag Not Found"}, 404
+        except Exception as e:
+            error_msg = f'Organisation Tag DELETE - unhandled error: {str(e)}'
             current_app.logger.critical(error_msg)
             return {"error": error_msg}, 500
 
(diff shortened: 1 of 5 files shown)


server/api/tags_apis.py
```diff
@@ -26,6 +28,58 @@ def get(self):
             current_app.logger.critical(error_msg)
             return {"error": error_msg}, 500
 
+    @token_auth.login_required
```
hunt3ri: You should add the @tm.pm_only() decorator here as well.  This decorator means you must be logged in and have the Project Manager role to use the API.  I'd suggest only Project Managers can delete tags.  Without the decorator any logged in user could delete a tag.

server/api/tags_apis.py
```diff
@@ -26,6 +28,58 @@ def get(self):
...
+    def post(self):
```
hunt3ri: You should use the delete verb rather than post here.  You could also pass the tag through as a param, which will save you fishing it out of the request, eg

`def delete(self, org_tag: str)`

server/api/tags_apis.py
```diff
@@ -26,6 +28,58 @@ def get(self):
...
+                description: Unauthorized - Invalid credentials
+            403:
+                description: Organisation tag belongs to some project
+            404:
+                description: Organisation Tag Not Found
+            500:
+                description: Internal Server Error
+        """
+        try:
+            organisation_tag = request.get_json().get('organisation_tag')
```
hunt3ri: Pass as param, see comment above.

server/models/postgis/project.py
```diff
@@ -383,6 +384,23 @@ def all_tasks_as_geojson(self):
 
         return project_tasks
 
+    @staticmethod
+    def get_all_organisations_tag(preferred_locale='en'):
+        query = db.session.query(Project.id,
```
hunt3ri: Add docstring here to describe purpose of method

server/services/tags_service.py
```diff
@@ -12,3 +20,20 @@ def get_all_organisation_tags():
     def get_all_campaign_tags():
         """ Get all org tags"""
         return Tags.get_all_campaigns()
+
+    @staticmethod
+    def delete_organisation_tags(organisation_tag):
+        """ Delete organisation tag"""
+        organisation_tags = Project.get_all_organisations_tag().tags
```
hunt3ri: If the purpose of the method is to delete duplicate tags it might be better to pass the tag ID rather than the name, but this is probably fine??

server/services/tags_service.py
```diff
@@ -12,3 +20,20 @@ def get_all_organisation_tags():
     def get_all_campaign_tags():
         """ Get all org tags"""
         return Tags.get_all_campaigns()
+
+    @staticmethod
+    def delete_organisation_tags(organisation_tag):
```
hunt3ri: Add type hint for all parameters

server/models/postgis/tags.py
```diff
@@ -53,3 +53,12 @@ def get_all_campaigns():
...
         return dto
+
+    def delete(self):
+        """ Delete the user in scope from DB """
+        db.session.delete(self)
+        db.session.commit()
+
+    @staticmethod
+    def get_by_organisation_tag(organisation_tag):
+        return Tags.query.filter_by(organisations = organisation_tag).one_or_none()
```
hunt3ri: If we have true duplicates and two tags match a string, I think this will error saying two found when one expected.  Possibly worth writing a test for this.

server/api/tags_apis.py
fiennyangeln: ok will add this.

server/models/postgis/tags.py
fiennyangeln: yes i think of this too. but when i see the method used to insert organisation tag, i see that it takes care of the duplicate (it wont add the organisation if it already exist). do you have any idea to further prevent it?

server/api/tags_apis.py
fiennyangeln: i initially tried using delete. but then i think if the organisation tag is quite long, it might cause trouble. however If what we pass is an ID then should not be a problem. but from the problem description, it seems like the thing to be passed is the organisation tag. what do you think?

server/models/postgis/tags.py
hunt3ri: @fiennyangeln in that case I'm happy to use the string

server/api/tags_apis.py
hunt3ri: @fiennyangeln I saw your comment re the fact that insert code handles identical dupelicates, in which case I'm happy that we pass the actual string we want to delete here.

hunt3ri:
@fiennyangeln one additional point, you need to register the new API endpoint with Flask-Restful.  Take a look at https://github.com/hotosm/tasking-manager/blob/develop/server/__init__.py to see how it's done in the init_flask_restful_routes function.  Give me a shout if you get stuck 👍

fiennyangeln:
@hunt3ri  ok will try adding it.

fiennyangeln:
Hi @hunt3ri I modified it. Should I do the same for campaign tag?

hunt3ri:
Hi @fiennyangeln sorry for delayed reply.  Yes I think the same delete API for campaigns will be useful.

xamanu:
This looks like to become obsolete with #1531

Issue #1002, opened 2017-12-20
Checkout-Checkin not shown in task history if task's 2 hour time limit reached


bgirardot:
If I checkout a task for mapping and then let it expire after 2 hours there are no entries in the task history. There should be an entry for the checkout and an entry for checking it back in, or at least one entry that says 'checked out for 2 hours' kind of message.

PR #1064, opened 2018-03-03
Add AUTO_UNLOCKED_FOR_* action in TaskHistory

sunidhiraheja:
Fixes https://github.com/hotosm/tasking-manager/issues/1002

For testing, I set the auto unlock time to 10 seconds. Here are the results:

image
image

### DIFF
--- a/client/app/project/project-dashboard.html
+++ b/client/app/project/project-dashboard.html
@@ -29,7 +29,7 @@
                                                 href="/user/{{ item.actionBy }}">{{ item.actionBy }}</a></span>
                                         <span ng-show="item.action === 'LOCKED_FOR_VALIDATION'">{{ 'Task locked for validation by' | translate }} <a
                                                 href="/user/{{ item.actionBy }}">{{ item.actionBy }}</a></span>
-                                        <span ng-show="item.action === 'UNLOCKED'">{{ 'Task checked out for 2 hours by' | translate }} <a
+                                        <span ng-show="item.action === 'AUTO_UNLOCKED'">{{ 'Task automatically unlocked by' | translate }} <a
                                                 href="/user/{{ item.actionBy }}">{{ item.actionBy }}</a></span>
                                         <span ng-show="item.action === 'STATE_CHANGE' && item.actionText === 'BADIMAGERY'">{{ 'Task marked as bad imagery by' | translate }} <a
                                                 href="/user/{{ item.actionBy }}">{{ item.actionBy }}</a></span>
--- a/client/app/project/project-dashboard.html
+++ b/client/app/project/project-dashboard.html
@@ -29,6 +29,8 @@
                                                 href="/user/{{ item.actionBy }}">{{ item.actionBy }}</a></span>
                                         <span ng-show="item.action === 'LOCKED_FOR_VALIDATION'">{{ 'Task locked for validation by' | translate }} <a
                                                 href="/user/{{ item.actionBy }}">{{ item.actionBy }}</a></span>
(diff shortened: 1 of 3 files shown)


bgirardot:
@sunidhiraheja Testing this now. Where did you set the 10 second time out?

sunidhiraheja:
@bgirardot I changed the { 'LOCKED', 'UNLOCKED' } action history to only include 'UNLOCKED' now, for when a task is automatically unlocked after timeout.
The LOCKED action is deleted, like before.

This is because maintaining the LOCKED action in the task history introduces a lot of complications elsewhere (I noticed this while working on https://github.com/hotosm/tasking-manager/issues/1065)

The new task history looks like this:

screenshot from 2018-03-06 04-40-39

Does this work?

I _can_ make it work if we decide to keep the LOCKED action in the task history, just that it'll take longer since a lot of the flow is written assuming that there's no LOCKED actions in a task's history if it is free for locking at any given point.

sunidhiraheja:
This is blocking https://github.com/hotosm/tasking-manager/issues/1065

bgirardot:
@sunidhiraheja Thank you very much for this, my apologies I have not been able to test it yet. But I will do that right now!

My initial reactiion to your comments above is to keep things as much like they work now as possible and not do any thing that requires great changes to a lot of code. 

If you say this is the best way for now, I trust you on that and I think we should take this (after I test it) 

I also see you are working on #1065, great, perfect, I will follow up there as someone else wanted to look at it and I did not see you were working on it.
 
Testing now......

thank you agian!

bgirardot:
@sunidhiraheja Hi. first off thank you again for your contributions here, very much appreciated. I do have some requests on this though please:

1. any "unlock" or "UNLOCK" text that you added as part of this PR could you please change it to "auto_unlock" or "AUTO_UNLOCK" to reduce confusion please. unlocking happens manually and automatically, using 'unlock' here makes things less clear. this is all related to auto unlocking.

2. is there a reason we can not use AUTO_UNLOCKED_FOR_MAPPING and AUTO_UNLOCKED_FOR_VALIDATION and then put in the actual time duration into the action_text like is done for mapping and validating now?

3. #2 above would also allow the messages displayed in the UI to not be generic and hard coded.

( I am somewhat concerned that the time to auto unlock tasks could change in the future, if not actually be a setting and I do not want to hard code the 2 hour assumption anywhere we can avoid it. )

bgirardot:
@sunidhiraheja I would even be fine if we just ended up with LOCKED_FOR_MAPPING and LOCKED_FOR_VALIDATION and then the 2 hour or whatever the actual time duration was.

But there is no reason for that total time to not be part of the tasks history

I like the idea of AUTO_UNLOCKED because then that would allow any analysis of the time spent on the project to selectively remove those if they choose to, but not required.

sunidhiraheja:
Hi, thanks a lot for the review!

1. Unlocked is now referred to as Automatically Unlocked in this use case.
2. Used AUTO_UNLOCKED_FOR_MAPPING and AUTO_UNLOCKED_FOR_VALIDATION, and action text is now set to the lock duration, but the lock duration is set manually (programmatically, just like how tasks are chosen to be automatically unlocked). This is so because regardless of the fact that we have the LOCK time, the time at which the auto unlock is executed is not exactly the allowed lock duration, but almost always a little bit more than that, to maintain a fair balance between performance and utility.
3. This is taken care of with the changed actions.

ethan-nelson:
@sunidhiraheja  if you can try merging a recent develop branch into this at some point I can test it and get it merged soon. Normally we wouldn't have to, but again because of those deployment changes in the files, the older PRs prior need that merge. Thanks!

Issue #1019, opened 2018-01-12
dense waypoint nodes of gpx task boundary, unreadable text.


govvin:
This is somehow related to https://github.com/hotosm/tasking-manager/issues/1018.

Every waypoint node of the gpx task boundary has a name, containing the warning message.  If the nodes are too dense, it displays like this, which can be confusing for users:

![image](https://user-images.githubusercontent.com/5821688/34855955-8e93b736-f77d-11e7-9f13-b8738dfbd6bb.png)

I'm not sure if i's it possible to render the warning along the line?

ethan-nelson:
Also see #636

fiennyangeln:
Hi i want to take this issue. Should I limit the number of warning message displayed per point / should i remove them alltogether and change it into one warning message?

ethan-nelson:
@fiennyangeln we still want the text there. It is just a problem when the object has very fine detail that text on every single point causes this.

Perhaps a resolution for now could be if the number of points in the GPX is greater than some number `N`, we only add the label on approximately every `N / 6` points. I would say N > 20 could be a good start.

bgirardot:
I have seen really bad ones, even 1 every six nodes would be a problem.

Is there anyway to check of the node is at a 90 degree point?

Otherwise, i think we might consider just removing them totally if there are more than N nodes and leaving the one set of text in the center that says do not edit outside of this area.

ethan-nelson:
A node situated somewhere in the center with text for shapes that have more than a few points is a great idea! (and probably much easier)

fiennyangeln:
@ethan-nelson @bgirardot i have submitted a PR ( for N>20 Make label on approximately N/6 points). Please review it :D What do you mean by checking 90 degree point? Is it between 3 adjacent point?

bgirardot:
@fiennyangeln I am going to skip testing this because it is not the solution to this problem. There are tasks that end up with hundreds of points so even N/6 is way way too many. I apologize, but as mentioned above, even one every 6 nodes is just too many.

What I mean about "90 degree nodes" is that if we only put the message on nodes that were at points that were at "right angles" or "90 degrees" that would limit the number of message. I do not know if you can test for that or not, please do not spend very much time on it unless it is very easy to test (see below illustration).

A much better solution is to just limit the number of messages. let us say 4 max.  So work out some way to evenly distribute the 4 messages, but no more than 4 + the one in the middle. 

Please see the illustration below:

[...]

PR #1068, opened 2018-03-11
make gpx boundary label less dense for more than 20 points.

fiennyangeln:
Bug #1019


bgirardot:
@ethan-nelson I think so. Unfortunately there are tasks with many vertices and just reducing them is not quite enough.

ethan-nelson:
Alright thanks. And thanks @fiennyangeln for all of your work on PRs.

PR #1070, opened 2018-03-14
Updates Learn page with TM video

bgirardot:


bgirardot:
@smit1678 I tested this locally, everything works, it is only html changes, can you please quickly review and merge?

PR #1075, opened 2018-03-17
Add API get user by id

ethan-nelson:
This adds the ability to retrieve a user by id number instead of username. This mostly leverages functions already in the codebase. Note the API URL endpoint required a different path (`/user-id/` instead of `/user/`) as usernames composed only of numbers would still be valid at the `/user/` path.


willemarcel:
@ethan-nelson sorry for the delay. PR approved

PR #1077, opened 2018-03-18
Update frontend on auto unlock

### LINKED ISSUE #1065: Auto unlocked task stays in locked UI state
When a task times out due to the 2 hour limit, it seems to get released just fine, but the UI shown to the user still looks like they have the task checked out.

It should instead:

1. put up a dialog that indicates this task expired and was auto unlocked after 2 hours
2. keep the user on that selected task in case they want to check it out again.

Below is are screen shots of what it looks like after a time out. You can see the map is updated to reflect they no longer have a lock on the selected task, but the UI has remained the same as if they do. The second screen shot shows that the system correctly identifies they can not mark a task as mapped or stop mapping or anything else because they do not have it locked.

2018-03-01_16h58_32

2018-03-01_16h58_56

sunidhiraheja:
Fixes #1065 

Here's a screenshot for an auto unlocked mapping task:
image

The behavior is exactly the same for validation tasks.

### DIFF
--- a/client/app/project/project.controller.js
+++ b/client/app/project/project.controller.js
@@ -562,7 +562,6 @@
                     vm.lockedByCurrentUserVectorLayer.getSource().clear();
                 }
                 vm.lockedTasksForCurrentUser = [];
-                refreshProject(vm.id);
                 if (vm.mappingStep === 'locked' || vm.validatingStep === 'locked') {
                     vm.mappingStep = 'viewing';
                     vm.validatingStep = 'viewing';
--- a/client/app/project/project.controller.js
+++ b/client/app/project/project.controller.js
@@ -38,6 +38,7 @@
         vm.taskUnLockErrorMessage = '';
         vm.taskSplitError = false;
         vm.taskSplitCode == null;
+        vm.wasAutoUnlocked = false;
 
         //authorization
         vm.isAuthorized = false;
@@ -190,6 +191,7 @@
             vm.taskSplitError = false;
             vm.taskSplitCode == null;
             vm.taskUndoError = false;
+            vm.wasAutoUnlocked = false;
         }
 
         /**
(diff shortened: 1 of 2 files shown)


ethan-nelson:
On initial testing, my tab became unstable after the first autounlock...looking at the network traffic on the browser, it may have been that the project was refreshing two times instead of one. This may, however, just be a limitation of the current project API. Or it could have been my local install. I'll have a bit more time tomorrow to dissect further. All the front end stuff looks great in any case!!

ethan-nelson:
Hi sunidhiraheja, thanks again for working on this! I dug into the issue a bit further. It looks like the refresh project function calls the add projects to task function that calls the update locked tasks for current user that checks for the autounlock that calls refresh project again, so it creates a replicating loop. I think it might be good to check for an auto unlock in the refresh project function instead since that is already done on the 10 second timing.

Additionally, I would set the auto unlock flag to false in the [onTaskSelection function](https://github.com/hotosm/tasking-manager/blob/2d4e5fc5178c45b79e2437ff12d3cf79cb249484/client/app/project/project.controller.js#L714). That way, when the user selects a new task (or reselects their auto-unlocked task) it should hopefully clear the message.

sunidhiraheja:
Hi @ethan-nelson, I had put the auto unlock check there because I detect an auto-unlock using the following logic:

If the user is on the task locked page, and the API says that there are no locked tasks at the moment, the current task must have been auto unlocked by the backend.

Should I just remove the refreshProject call from there? I'm not sure why I had put it at there at the time, since it's not necessary.

If not, should I send some response from the API in the refreshProject call to indicate an auto unlock?

About setting the auto unlock flag to false, I put it in the resetErrors function since it's an error, and it is called from within onTaskSelection in all if-else branches.

dakotabenjamin:
This works fine for me now, it's running on tasks-demo.hotosm.org. I'll approve and merge.

ethan-nelson:
Thanks @dakotabenjamin for merging and thank you @sunidhiraheja for this fix!! (Sorry I am on travel for a few days.) This PR coupled with the auto unlock countdown fixes a pretty major regression from TM2.

PR #1078, opened 2018-03-21
Add invalidated count to user DTO/API call

ethan-nelson:
Currently the `/api/user` endpoint returns a DTO about a user that includes the counts of tasks validated and tasks mapped, but not tasks invalidated. This information is already in the database and is desired by data consumers, so this PR adds the tasks invalidated count to the base user DTO and also when the DTO is populated.


ethan-nelson:
Easy review if you don't mind, @bgirardot

bgirardot:
@ethan-nelson Will review tonight, thank you for the guidance!

Issue #958, opened 2017-11-15
Add links on the User Profile to Pascal's tools like heatmap


bgirardot:

aaj013:
I'll try to add this

PR #1091, opened 2018-04-24
Add links on the User Profile to Pascal's tools like heatmap

aaj013:
Add links on the User Profile to Pascal's tools like heatmap #958


ethan-nelson:
Great, thanks for your help!!

Issue #902, opened 2017-10-30
Organisation and Campaign list - sorting


tastrax:
Given the lists in both these areas will continue to grow is it possible to have then sorted alphabetically to make it easier to find things.
![tm3-list-sort](https://user-images.githubusercontent.com/8775395/32154762-e925654a-bd86-11e7-8f25-bab2f0df97bf.png)
?

verdy-p:
See also bug #509

aaj013:
I'll try to do this @tastrax

aaj013:
Please check my PR @tastrax  @bgirardot

uwaiszaki:
@ramyaragupathy I think the issue is solved.

ethan-nelson:
Yes--indeed completed!

PR #1053, opened 2018-02-20
fix organisation tag to display only those with project list

fiennyangeln:
#881

PR #1092, opened 2018-04-24
Organisation and Campaign list - sorting

aaj013:
Organisation and Campaign list - sorting #902


ethan-nelson:
Hi @aaj013 thank you very much for your PR. It looks like this problem will be solved through the code in #1053 / #1148 that relates to overall tidying up the list of organizations and campaigns listed, so I am going to go ahead and close this. If you are interested in working on another bug, do ping me and I can steer you towards some other issues we are having. Thanks again for your interest!

Issue #1098, opened 2018-04-26
Translations needed for: Português, Lietuvos, Dansk, Tamil (தமிழ்), Sinhala  


aaj013:
Missing JSON files for following languages:

- Português
- Lietuvos
- Dansk
- Tamil (தமிழ்)
- Sinhala (සිංහල)
![missing languages](https://user-images.githubusercontent.com/5847582/39304935-0cfe17f4-4979-11e8-96c0-86687bd2a0ac.png)
@bgirardot

aaj013:
I've done translating and created the json files

xamanu:
The problem here are not the files, but having the TM translated into these languages by native speakers.

xamanu:
Related to #1678

xamanu:
This is not occurring with the current version. Considering this as fixed. Please reopen in case there is still a problem.

PR #1099, opened 2018-04-26
Missing JSON files for some language translations

aaj013:
Missing JSON files for some language translations #1098


ethan-nelson:
Hi @aaj013 thanks for your PR. Did you pull these down from Transifex manually? We usually do this semi-automated, but it's been a little while since we've updated. I want to make sure as otherwise these will get overwritten on the next pull down.

aaj013:
Hi @ethan-nelson, I just used English json file and translated the values it to required languages using Google Translate manually

ethan-nelson:
Hi again aaj013, thanks for your response (and for your other PR contributions as well). Unfortunately there is some hesitation that results from using a translator engine. Firstly, we try to seek out translators who speak the language so that the translation makes sense (computerized translations are better for some languages than others). Secondly, there is the matter of the Terms of Service with respect to the engine and taking results from that for the project.

If you speak any of these languages, I would strongly encourage you to check out the Transifex project at https://www.transifex.com/hotosm/tasking-manager-3/ and help translate in those.

PR #1105, opened 2018-05-11
Add New Relic agent

dakotabenjamin:
Enables New Relic monitoring agent in the app. 

@smit1678 I added you as a user, so you can see how it's working for tasks-demo.hotosm.org


smit1678:
@dakotabenjamin :+1: great, the monitoring looks like it will be helpful.

Issue #1086, opened 2018-04-19
when offering trim, warn that you can't split non square tasks


mikelmaron:
You can only split tasks that are square. We used the trim feature when setting up a project for an island, and only later found that only a few tasks are splittable. 

At the very least, offer a warning when creating tasks that the trim will result in X squares that are not splittable.

mikelmaron:
Or explain this on the the task page as well (ie don't hide the split button, but disable it with a hover)

PR #1106, opened 2018-05-13
Improve GPX box

ethan-nelson:
Based on a recent discussion, this PR removes the warnings from nodes on a task GPX. Now the warning is only shown in the middle of the task geometry.

This also fixes a lingering issue of the GPX warning saying `box` instead of `area`.


bgirardot:
@ethan-nelson Thank you very much Ethan, but I do not think this is an alternative to issue #1086

bgirardot:
@ethan-nelson I see, this is an alternative, building upon pull request #1068

PR #1107, opened 2018-05-13
Default projects to intermediate difficulty

### LINKED ISSUE #1081: Default project difficulty to "intermediate"
Project managers need to manually set projects good for beginners to beginner projects. We have a lot of projects that are not beginner projects marked as beginner.

ethan-nelson:
Fixes #1081

### DIFF
--- a/server/models/postgis/project.py
+++ b/server/models/postgis/project.py
@@ -43,7 +43,8 @@ class Project(db.Model):
     default_locale = db.Column(db.String(10),
                                default='en')  # The locale that is returned if requested locale not available
     author_id = db.Column(db.BigInteger, db.ForeignKey('users.id', name='fk_users'), nullable=False)
-    mapper_level = db.Column(db.Integer, default=1, nullable=False, index=True)  # Mapper level project is suitable for
+    mapper_level = db.Column(db.Integer, default=MappingLevel.INTERMEDIATE.value,
+                             nullable=False, index=True)  # Mapper level project is suitable for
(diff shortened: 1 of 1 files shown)


bgirardot:
@ethan-nelson Thank you very much for this. I am going to sit on this for a bit longer to make sure all the stake holders are aware of the change and @russdeffner and I can generate some guidance on what makes a "Beginner" project. Basically, we need to do this but we have to better support the PMs before we do it.

ethan-nelson:
@russdeffner do you have any insights into this, or @bgirardot ? Is this idea desired by the AWG?

russdeffner:
I think now that we have a fairly 'standard' on-boarding for new PMs it's less necessary, but still think it's a good idea - i.e. would rather see beginner projects 'mislabeled' as intermediate than an advanced projects mislabeled as beginner. Also, project difficulty should be more readily understood as it is part of the training.

ethan-nelson:
Hi Russ, thanks for the feedback. Thinking about other TM users, this could be a bit clumsy for someone setting up a fresh instance and wondering why people can't immediately edit a brand new project without much intervention. So I'm going to go ahead and table this (it's a trivial change if we later decide to go forward--though note that a database migration will also be required beyond this change).

ethan-nelson:
Ok chatting with Russ a bit more on Slack, the idea here is it would be okay to default the classification of projects to intermediate, but not _restrict_ the projects to intermediate contributors. I will add the database migration (migration being only setting the new default) and open a new PR.

Issue #635, opened 2017-06-27
Allow users to opt out of validation messages and emails


bgirardot:
Experienced users might wish to no longer get validation / invalidation emails.

Based on feedback I changed this to be opting out of the automated validation tasking manager messages totally as well as the email notices about those messages.

OttoROSM:
Hi Blake,
Came here to report this. At least do not send a standardized message to the e-mail address that there is a message in the tasking manager. Especially when the message in the Tasking Manager is a standard message where someone validated a tile.

Typically I would map multiple tiles after each other, so when someone else validates 10 in a row, I get 10 e-mails saying there is a message in the tasking manager for me.
Then I go to the tasking manager and need to open 10 times a similar message saying that my tile was validated and I need to keep up the good work. Then click delete for each message and confirm I want to delete each message and then go back to my e-mail inbox and delete 10 e-mails.

Rather I would receive the message directly in my inbox, or receive a summary report saying, 10 tiles were validated, or you have 10 messages in the tasking manager. 

Basically unless someone made a specific comment, I am fine with getting a simpel statistic saying 10/10 tiles you mapped were validated.

verdy-p:
10 mails in the TM manager itself that need to be clicked separately and deleted by another click, it's too much. And yes I'm not necessarily interested to have such inbox mails, or even notifications.
- there should be options to disable these notifications or mails in the TM messsaging box
- also an option to not forward these with more notifications or mails to our regular mailbox for such events (we need filters by notification type to enable/disable them at their respective source)
- also a way to select multiple messages in the TM box to archive/delete/mark them as read without having to click each one.

ethan-nelson:
For someone interested in working on this, here's a possible route that should cover most of the work:

1) In the users schema, add a database column for email notifications (https://github.com/hotosm/tasking-manager/blob/develop/server/models/postgis/user.py)
2) Add a corresponding field in the user data object for this column (https://github.com/hotosm/tasking-manager/blob/develop/server/models/dtos/user_dto.py)
3) Add a corresponding field in the user update api (https://github.com/hotosm/tasking-manager/blob/develop/server/api/users/user_apis.py#L52-L116)
4) Add a selectable option on the user profile to disable notifications (https://github.com/hotosm/tasking-manager/blob/develop/client/app/profile/profile.html)
[...]

PR #1060, opened 2018-02-25
Allow Opt out of all automated validation messages and emails

fiennyangeln:
#635

PR #1109, opened 2018-05-20
Stop sending messages when tasks are validated

JesseWeinstein:
Fixes https://github.com/hotosm/tasking-manager/issues/635 by simply removing the entire feature for sending messages when a task is validated.

### DIFF
--- a/server/services/messaging/message_service.py
+++ b/server/services/messaging/message_service.py
@@ -43,28 +43,6 @@ class MessageService:
         return welcome_message.id
 
     @staticmethod
-    def send_message_after_validation(validated_by: int, mapped_by: int, task_id: int, project_id: int):
-        """ Sends mapper a thank you, after their task has been marked as valid """
-        if validated_by == mapped_by:
-            return  # No need to send a thankyou to yourself
-
-        text_template = get_template('validation_message_en.txt')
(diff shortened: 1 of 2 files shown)


bgirardot:
Hi @JesseWeinstein Thank you very much for the pull request.

We will not be removing the feature, but we will add a way to opt out of validation emails. But it is important for newer mappers to get feedback that their mapping has been validated. 

We will add a message to the automated email so folks know they can opt-out.

Since we have another way to address this issue, i am going to reluctantly close your pull request, my apologies. 

If you have any other ideas for contributions, the best thing is to open or comment on an existing issue with your plan or ideas for the fix or feature. That way you can get feedback on the approach first.

JesseWeinstein:
That is completely fine -- I mostly wanted to make it just to move the discussion along. Is someone already working on the opt-out feature, or would it be useful for me to create a PR for that?

bgirardot:
I think there is a pull request for the opt out right now ready to merge.

JesseWeinstein:
I don't see any such PR in https://github.com/hotosm/tasking-manager/pulls -- do you remember what it's number might be?

dakotabenjamin:
If you're referring to #1060 I just merged it into `develop`, and opened a PR into `master`. Once that is merged it will function as intended on the site.

PR #1114, opened 2018-05-24
Match circle dependencies to Dockerfile

mojodna:
This wasn't breaking CircleCI yet (`python:3.6` is probably cached since it only changed recently), but it will soon.


dakotabenjamin:
I'm gonna hold off on merging this until later tonight so there arent any service interruptions while traffic is high.

PR #1116, opened 2018-05-31
EB db upgrade

dakotabenjamin:
This PR creates a script in EB that will run the python db upgrade after deployment. Please note that this happens every time an instance is deployed which is not ideal but for the most part will do nothing. 

@smit1678 @bgirardot @ethan-nelson this is a precursor to being able to re-merge the dev changes (new relic, etc) into `master`.


dakotabenjamin:
I forgot to mention I tested on the demo environment and it worked

smit1678:
@dakotabenjamin I updated the staging database to be up to date with Production as of today so we should be good to test.

PR #1119, opened 2018-06-05
fix missing maps badges link in User profile section

drklrd:
Current link to "Missing Maps Badges" takes to invalid URL. This commit will fix the link

### DIFF
--- a/client/app/profile/profile.html
+++ b/client/app/profile/profile.html
@@ -70,7 +70,7 @@
                                     </a>
                                    </p>
                                 <p>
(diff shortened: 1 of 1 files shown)


drklrd:
🙂

drklrd:
@smit1678 do this need any more changes from my side ? I see that the checks are failing. Do I need to do any other things in particular ?

ethan-nelson:
Hey @drklrd thanks for your help with this! The failing check is caused by an intermittent issue we're in the process of fixing related to the docker configuration. The fix for that should be deployed soon, after which we can merge your PR in!

drklrd:
@dakotabenjamin I have updated the branch and pushed the code again.

Issue #945, opened 2017-11-11
Completion Stats - Validation should not be higher than Mapped


russdeffner:
I noticed that project 3730 had a higher Validation percentage than Mapping percentage.  Looking at the project tasks, and later confirmed by @ethan-nelson that the tasks marked as "Bad Imagery" are taken into account in Validation percentage but not Mapping.

Here's what Ethan said on slack: 
it looks like mapped is fraction mapped over all tasks not including bad imagery, while validated is validated+bad imagery over all tasks (https://github.com/hotosm/tasking-manager/blob/develop/server/models/postgis/project.py#L300-L301). that should perhaps be refined a bit

johnwhelan:
The other comment is the only way to leave a comment at the moment is to open the tile for revalidation.  Does this get counted twice in the validation count?

John

bgirardot:
Another example of a project where the percent mapped is lower than the percent validated. Further, the percent mapped shows as 98%, but when looking at the project, the whole thing is 100% validated.

https://tasks.hotosm.org/project/3815

verdy-p:
I see 97% mapped 99% validated...

2017-12-06 18:29 GMT+01:00 Blake Girardot <notifications@github.com>:

bgirardot:
@verdy-p Thank you very much verdy, i had not looked at the stats page, i was going by the card on the contribute list. This should help us figure out what is going on.

I know someone mentioned on another item that how it was calculated was different in the two places, but I can't locate that now.

But from the look of the graphic, it looks like it should be 100/100. Seems like it is probably related to the bad imagery option now available.

![2017-12-06_20h11_45](https://user-images.githubusercontent.com/265831/33680422-de7aae86-dac1-11e7-86d9-153b0018167b.png)
![2017-12-06_20h11_25](https://user-images.githubusercontent.com/265831/33680430-e28bbac4-dac1-11e7-93ed-45cc2ee532a8.png)
![2017-12-06_20h10_47](https://user-images.githubusercontent.com/265831/33680442-e9b9429e-dac1-11e7-8609-d7b5efb2ca5b.png)

verdy-p:
Given the map is all green, there's no tile which is marked with "bad imagery" (it would be another color), but may be this occured in the past and a better imagery was used, the tile was invalidated and finally mapped and validated.
Is it possible that tiles that were present before a splitting are still counted and were not really marked as splitted and to ignore?

pgiraud:
What do you think about this one?
[...]

PR #1123, opened 2018-06-15
Include bad imagery in mapped, validation stat calculations

ethan-nelson:
Currently, bad imagery is excluded from mapped percentages but is added to validated percentages in the UI. This is opposite of what we want because bad imagery tasks _can_ be validated.

A task that is "ready to be mapped" can be marked mapped or bad imagery; these two choices are on an equivalent level in terms of task completion but are disjointed. (These two choices are not collectively exhaustive, though, as a state can be re-marked as ready to be mapped.) We therefore want the percent mapped to include both mapped and bad imagery counts.

Tasks either in a mapped or bad imagery state can be validated, after which they share the same status. Thus, we want the validated percentage to only count validated tasks over all tasks.

The statistics calculated and returned to the UI have been modified to match this. This fixes half of the problem mentioned in #945 (the other half related to invalidations that is more than just a display fix).

### DIFF
--- a/server/models/postgis/project.py
+++ b/server/models/postgis/project.py
@@ -297,8 +297,8 @@ class Project(db.Model):
         centroid_geojson = db.session.scalar(self.centroid.ST_AsGeoJSON())
         summary.aoi_centroid = geojson.loads(centroid_geojson)
 
-        summary.percent_mapped = int((self.tasks_mapped / (self.total_tasks - self.tasks_bad_imagery)) * 100)
-        summary.percent_validated = int(((self.tasks_validated + self.tasks_bad_imagery) / self.total_tasks) * 100)
+        summary.percent_mapped = int(((self.tasks_mapped + self.tasks_bad_imagery) / self.total_tasks) * 100)
+        summary.percent_validated = int((self.tasks_validated  / self.total_tasks) * 100)
 
         project_info = ProjectInfo.get_dto_for_locale(self.id, preferred_locale, self.default_locale)
         summary.name = project_info.name
--- a/server/services/project_search_service.py
+++ b/server/services/project_search_service.py
@@ -78,9 +78,9 @@ class ProjectSearchService:
             list_dto.organisation_tag = project.organisation_tag
             list_dto.campaign_tag = project.campaign_tag
             list_dto.percent_mapped = round(
-                (project.tasks_mapped / (project.total_tasks - project.tasks_bad_imagery)) * 100, 0)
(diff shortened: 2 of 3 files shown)


bgirardot:
@ethan-nelson Is there any chance this patch is not counting task squares as mapped when I mark them completely mapped? That seems to be the case in my testing, but the caching can get me sometimes. but at the moment it looks like if I mark as bad imagery, it correctly counts it as mapped, not as validated, but then if i checkout another task square and just mark it completely mapped, that does not seem to be reflected in the mapped % anymore.

bgirardot:
@ethan-nelson I am really close to approving this as is. I did a lot of mapping, validating, bad imagerying, invalidating, splitting, etc, etc, and it kept to exactly what I expected the percentages to be.

Really as far as I am concerned it could be merged in as is.

However, I did eventually end up with 100% mapped and 100% validated and everything looked good.... then I did "edit project" -> actions -> invalidate all.

It marked them all as red/invalidated, but my percentages did not go to 0%/0%. They went to 26% mapped and 0% validated and I have no idea why.

Up to you ethan, if you want to look and/or fix it so "invalidate all" button works as expected, great, if not, we should merge what you have here. by definition the invalidate all button is an edge case.

bgirardot:
@ethan-nelson Never mind, i can not easily reproduce the above issue using the validate/invalidate all buttons and various mapping, bad imagery, validate, invalidate, etc. It could very well be related to some other issue and nothing to do with this bad imagery change. I think this should be merged as is.

ethan-nelson:
Hey Blake, thanks for the comments and for your rigorous testing of these changes. The caching behavior can be modified here if you want to locally (or suggest a change to the repo as I don't know what a good time frame would be): https://github.com/hotosm/tasking-manager/blob/032b02da8bee70f47a0638b1add6c01fec540b6a/server/services/project_service.py#L12

Regarding the invalidate all behavior, I didn't look into that originally with this PR but I believe it stems from the problem here: https://github.com/hotosm/tasking-manager/blob/032b02da8bee70f47a0638b1add6c01fec540b6a/server/services/stats_service.py#L104-L106 where invalidation removes a mapped count for every task regardless of whether or not they were bad imagery or mapped (see this comment for more info: https://github.com/hotosm/tasking-manager/issues/945#issuecomment-397490146). I would consider that to be a separate issue from this PR because, though related, it isn't necessarily the result of UI but rather problems with the calculation in the database. I plan to dig into that problem as well but would like to keep it separate since it will require some recalculation of the database for a proper fix of prior tasks.

Issue #1127, opened 2018-06-26
Retrieve archived projects using non-authenticated API request


ramyaragupathy:
At present https://tasks.hotosm.org/api/v1/project/search endpoint outputs all `active` projects from Tasking Manager. When a session key is provided as a query parameter with this url, `archived` projects can also be retrieved. However, each session key seems to return the archived projects just once. Further API queries with the same session key returns only active projects. I could repeat this behaviour with my OSM user id (Project Manager access on TM) almost 3 times yesterday.

- Generate session key
- Use the session key with the test link in the [documentation](https://tasks.hotosm.org/api-docs/swagger-ui/index.html?url=https://tasks.hotosm.org/api/docs#!/search/get_api_v1_project_search). 
- Update the `projectStatuses` parameter to `ARCHIVED` and test the link

Would be great if the API could fetch all archived + active projects without the authentication key. This would be super useful to update stats in the new website, where the numbers are refreshed once every 6 hours. 

cc @smit1678 @sunidhiraheja

bgirardot:
This has come up before I think. I think it makes sense to allow access to archived project by non authenticated users. Even the front end UI should probably allow mappers to included archived projects in their searches.

ethan-nelson:
This has been added in #1130...will be pushed to master soon (TM).

ramyaragupathy:
@ethan-nelson @sunidhiraheja How long would this take to reflect in the API testing interface at swagger? I'm checking this https://tasks.hotosm.org/api-docs/swagger-ui/index.html?url=https://tasks.hotosm.org/api/docs#!/search/get_api_v1_project_search

bgirardot:
hi @ramyaragupathy Try:

https://tasks-stage.hotosm.org/api-docs/swagger-ui/index.html?url=https://tasks-stage.hotosm.org/api/docs#!/search/get_api_v1_project_search

that has the changes now. It will be next week sometime before it makes it to the production server.

ramyaragupathy:
Thanks @bgirardot 🙏 , I'm able to retrieve the archived projects through staging 🎉

PR #1130, opened 2018-07-03
Allow viewing archived projects by non authenticated users

sunidhiraheja:
The API now allows the behavior requested in https://github.com/hotosm/tasking-manager/issues/1127

While doing the frontend part, I noticed some problems that need to be addressed before I can push the change; discussing on Slack.

### DIFF
--- a/client/app/contribute/contribute.html
+++ b/client/app/contribute/contribute.html
@@ -49,20 +49,21 @@
                                   <form class="form">
                                       <fieldset class="form__fieldset">
                                           <div class="search-options-container">
-                                              <div class="form__group search-option" ng-show="accountNavCtrl.account.role === 'PROJECT_MANAGER' || accountNavCtrl.account.role === 'ADMIN'">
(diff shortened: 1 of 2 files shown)


bgirardot:
@sunidhiraheja Can you please add more information when you say "I noticed some problems..."

Could you please list the problems otherwise otherwise reviewers and fellow contributors have no idea what you might be referencing.

Are these problems with this PR or problems in another part of the software? etc...

sunidhiraheja:
@bgirardot Sure, sorry I missed that.
The problem I was facing was that some times, `ng-click` was firing before `ng-model` updated the corresponding value in the model, resulting in unexpected behavior.
So, I changed it to `ng-change`, which monitors the variable mapped using the corresponding `ng-model` directive and only then fires the search function.

sunidhiraheja:
Ready for review! :)

ethan-nelson:
Will take a look this evening--thanks!

PR #1133, opened 2018-07-06
Update OSM authentication for TeachOSM

dakotabenjamin:
Fix for #1132


geomantic:
@dakotabenjamin Just tested the TM and it looks like it's fixed now. Thanks for the prompt turnaround!

PR #1134, opened 2018-07-10
Adds text to emails and messages about opt out

bgirardot:


bgirardot:
@ethan-nelson When you get a few moments could you review this please. It is just changes to the text of the emails and messages that get sent to let folks know about the opt-out option now.

I have tested it locally with an smtp server specified and it all seems to work as expected. 

As I said, it was just simple text changes, not even translatable strings :(

smit1678:
I think this looks fine to me. 👍

bgirardot:
@smit1678 Thank you very much nate!

PR #1135, opened 2018-07-10
adds languages ar pl pt updates translations of the rest

bgirardot:


ethan-nelson:
Blake thank you for trudging through this to get the translations updated. I see pt and ar are all English strings, but I'm imagining that's because of a lack of translations. 

For the new languages, would you mind adding them to https://github.com/hotosm/tasking-manager/blob/4997d41394d32283c6cd30cb8bffc97cc6aefc2a/server/config.py#L34-L39 so that they show up?

bgirardot:
Hi @ethan-nelson I do not want to deflate the celebration, but this is just an update from the existing translations. Somehow that seemed like the right first step, get what has been translated merged in. Then merge in the new en.json you generated for us and get those strings translated.

But I wanted to get all the translator's work so far merged in.

Also, while it looks like I spelled Brazil wrong, this PR already includes config.py updates.

I assume the all english new languages are requests we got from the community to add a language. That happens pretty often, we get requests to add a language and then they never get translated, but let us see what happens, we usually wait a year or so before removing them.

Thanks ethan!
cheers
blake

ethan-nelson:
Woops, you're absolutely right. Thanks!

bgirardot:
I think I am ok:

" in Portuguese, the correct spelling of the country´s name is Brasil"

http://www.aboutbrasilia.com/facts/brazilia.php

bgirardot:
but i do see a mistake here, let me fix and push again. i think I have pt in there twice.

bgirardot:
I tested locally and it seems to still work fine.

PR #1141, opened 2018-07-18
Add time spent mapping on user profile

### LINKED ISSUE #1082: Display users time mapping on their profile page
It would be great if mappers had some sense of how much time total they have spent mapping. It would be even nicer to maybe let them put in a time period to see how much time they contributed for a set time period.

This comes up quite often with people who want to volunteer, but need to track their hours, there is no good way to do that, the tasking manager is probably the best place to try and keep track of that, especially since we have the data more or less.

sunidhiraheja:
Fixes #1082 
image

Ready for review

### DIFF
--- a/client/app/profile/profile.controller.js
+++ b/client/app/profile/profile.controller.js
@@ -15,6 +15,7 @@
         vm.username = '';
         vm.currentlyLoggedInUser = null;
         vm.userDetails = null;
+        vm.userStats = null;
         vm.osmUserDetails = null;
         vm.projects = [];
         vm.map = null;
@@ -52,6 +53,7 @@
             projectMapService.addPopupOverlay(hoverIdentify, clickIdentify);
             getUserProjects();
             getLevelSettings();
+            getUserStats();
         }
 
         /**
@@ -205,5 +207,17 @@
                 vm.mapperLevelAdvanced = data.mapperLevelAdvanced;
             });
         }
+
+        /**
+         * Get stats about the user
+         */
+        function getUserStats() {
+            var resultsPromise = userService.getUserStats(vm.username);
+            resultsPromise.then(function (data) {
+                // On success, set the detailed stats for this user
+                vm.userStats = data;
+            });
+        }
+
     }
 })();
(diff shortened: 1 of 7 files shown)


bgirardot:
Cool! Can't wait to test it.

ethan-nelson:
Hi @sunidhiraheja, thanks! I tried running this out of the box and ran into an error:

```
--------------------------------------------------------------------------------
CRITICAL in stats_api [/home/enelson/tasking-manager/server/api/stats_api.py:180]:
User GET - unhandled error: String does not contain a date.
--------------------------------------------------------------------------------
```

This may be because we only recently added the action text duration in #1064, so for older tasks there is nothing that can be calculated. Additionally, this error took about ~30 seconds of postgres processing to come up--at least on my local machine--so it required some computation. Considering those points, I'm wondering if we should reconsider the implementation of this. Users that have hundreds or thousands of task events (locks, unlocks, etc) will put a pretty heavy load on the database when the time spent is recalculated on every profile load.

What if instead we kept a running count of time that tasks are locked for mapping in the user column? Then in a given task event of unlocking/marked mapped/marked validated, the delta time would be added to the total time column at that stage. Finally, the stats endpoint would then query and display that total time column.

This would also require an initial calculation of all the check out times in some form (either a migration script or a function). It would need to be coordinated to minimize downtime or database load, but is not unfeasible I don't think.

What are your thoughts. Does this seem reasonable? @bgirardot or others, any thoughts?

ethan-nelson:
Two amendments to the above:

Regarding the error: diving a little further, it looks like mapping durations were not calculated during migration (https://github.com/hotosm/tasking-manager/blob/fe0b152dac8740c046ebc6b159718c2fcb3f00f2/devops/tm2-pg-migration/migrationscripts.sql#L181). What we could do for the time being is focus on time calculations only based on TM3 mapping (perhaps adding `TM3 only` or something on the display). So we could add a try or if clause to ignore cases where there's a none occurring or an error thrown in the loop that sums them. 

Regarding the timing: it appears the major blocker in timing was the user's project contribution listing and not this one (which really needs review in #1144), so I think the load caused by the stats will not be bad.

Bottom line: if you can make the changes adding `TM3 only` to the text and adding an if action_text is not None block to the time addition, this should be good to go!

ethan-nelson:
Thanks--I will take a look again very soon.

ethan-nelson:
Thank you again @sunidhiraheja for this feature addition!!

Issue #917, opened 2017-11-04
Different amounts of "Tasks Mapped" and "Tasks Validated" - Web User Profile vs. API User


pa5cal:
When I compare the amounts of "Tasks Mapped" and "Tasks Validated", I get different results:

HOT Tasking Manager User Profile - tshedy 
(http://tasks.hotosm.org/user/tshedy)
"1992 - Tasks Mapped" and "2351 - Tasks Validated"

HOT Tasking Manager API - user 
(http://tasks.hotosm.org/api/v1/user/tshedy/mapped-projects)
'Projects'='17', 'Tasks mapped'='245' and 'Tasks validated'='5'

verdy-p:
Note that 'Tasks mapped'='*' and 'Tasks validated'='*' are distinctly reported for each project.
Your global counter in your user profile may include projects that are not listed in your user API (probably those projects that are been deleted, or only drafted, or whose details are no longer available after completion (the history will be only in your edits on the OSM dabase for your "changesets", but not the TM)
Also I wonder about what happens to your personal counters if some task you "mapped" or "validated" was finally invalidated (by someone else or even by you!)

akankshanayak2111:
Hi, I was wondering if I can take up this issue?

bgirardot:
@akankshanayak2111 yes, you can take this one up if you want. I am pretty sure there is a related issue somewhere too. Let me see if I can find it.

bgirardot:
@akankshanayak2111 I think issue #792  might be related

bgirardot:
@akankshanayak2111 I will note that issue #792 is a bit more urgent than this issue. They seem related, you are welcome to work on either one, but #792 would be really nice to get fixed and hopefully in the process you will see what is causing this issue. Please let us know if you have any questions.

ethan-nelson:
#1192 should hopefully reconcile this once it goes live

xamanu:
Is this fixed by now?

xamanu:
Closing because of inactivity.

Issue #792, opened 2017-09-17
User profile - stats, number of mapped/validated tasks in project list is incorrect


rwst:
The **total** number of mapped/validated tasks displayed on top looks right. The project-specific numbers however are way off. Most validated numbers are zero and don't add up to the total by far. Also the mapped numbers look too small.

majkaz:
Even the list of the projects on the profile page is wrong.
For me, only a part of the project list is actually there. The whole Iran-Iraq Earthquake 2017 is completely missing (not in projects, not in maps), some other projects I have participated are missing too. And yes, the numbers for individual projects mentioned there are wrong as well.

bgirardot:
Just for reference, @majkaz has contributed to https://tasks.hotosm.org/project/3833 but that project does not show up in their user profile page.

SDavies-OSM:
Would love this to be bumped up the priority list a bit. The user stats are like a list of personal achievements, this is like a recognition of work done. Human psychology means people being able to see their totals increasing will help inspire them to do more. Conversely, if their contributions aren't showing up, they may get dispirited.

majkaz:
@bgirardot 
I have just noticed additional difference - the number of tasks i can see for myself in one project is different on the user profile (8) and in the project activity table (11) - see attached screenshots.

As unconnected note, the validated count shows only the tasks marked as validated. It might be a good idea to show the invalidated count in the project activity tab as well (in this project, I would see my activity as validator as suspect by the numbers only, because I have invalidated quite a number of tasks). High number of invalidated tasks compared to validated ones might indicate problems - with validator/users/project. 

![2017-12-14_102440](https://user-images.githubusercontent.com/23097386/33985057-b22ea410-e0b9-11e7-8fc0-622a5a867009.png)
![2017-12-14_102510](https://user-images.githubusercontent.com/23097386/33985058-b2944054-e0b9-11e7-9610-f896b9d874d4.png)

bgirardot:
Additional information on this.

On the user's profile page, the listing of projects they have contributed to, the number of tasks mapped in a project listing is actually showing the number of tasks they have mapped that are not yet validated. In other words that individual listing is not showing the number of tasks that a user mapped, but only the number of unvalidated tasks the user has in that project.

You can see below that it shows 0 when in actually that user has mapped 4 tasks, which are all validated.

[...]

PR #1144, opened 2018-07-19
Improved mapped projects by a user query

ethan-nelson:
A performance issue in query monitoring and lingering bug reports (e.g. #792, #917, personal communication from slack somewhere about people unable to load their own profile) led to a first step in possible performance improvements through queries.

This PR includes ~two~ _one_ main change~s~:

~1) (I believe) corrects logic so that we return projects where a user id has either mapped _or_ validated a task, not both. I'm also not sure why we were searching for null user IDs as well. Please double check this--I feel like this is a fix, but I could be thinking about it incorrectly.~
~2)~ Removes project AOI/geometry from results. From what I can see in the javascript and testing, only the centroid is loaded on the frontend (c.f. https://github.com/hotosm/tasking-manager/blob/2f9ee6bb3b2fe42c3b1e24c5da6f6bebf316aced/client/app/services/projectMap.service.js#L162). I can imagine a cool case where someone may zoom in and see the AOI instead of a point, but this really becomes a performance issue now (see below).

Testing `bgirardot`'s profile on the main server results in a time out after about 50 seconds.

Testing `bgirardot`'s profile in my local install, the original query returned about a 3 MB file. This however was not really a correct answer because of the abovementioned issues. 

Testing `bgirardot`'s profile in my local install after implementing 1), the query was a 11 MB file and it was truncated by Firefox (whether that's the server or my browser, I really don't know or care at this stage).

Testing `bgirardot`'s profile in my local install after implementing 2) after implementing 1), the query was 61 kb. Additionally, some very back of the envelope math of adding the validated and mapped counts in his profile seems to reach the numbers displayed at the top of the profile more than the original.

### DIFF
--- a/server/models/dtos/user_dto.py
+++ b/server/models/dtos/user_dto.py
@@ -59,7 +59,6 @@ class MappedProject(Model):
     tasks_validated = IntType(serialized_name='tasksValidated')
     status = StringType()
     centroid = BaseType()
-    aoi = BaseType()
 
 
 class UserMappedProjectsDTO(Model):
--- a/server/models/postgis/user.py
+++ b/server/models/postgis/user.py
@@ -143,7 +143,8 @@ class User(db.Model):
                         tasks t
                   where p.id in (select unnest(projects_mapped) from users where id = {0})
                     and p.id = t.project_id
-                    and (t.mapped_by = {0} or t.validated_by = {0})
+                    and (t.mapped_by = {0} or t.mapped_by is null)
+                    and (t.validated_by = {0} or t.validated_by is null)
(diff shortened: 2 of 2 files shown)


ethan-nelson:
@smit1678 this should be ready for your review now. As mentioned in Slack, I reverted the changes that attempted to fix the correct counts problem (see #1192 for more info on that) and only retained the part that should improve the ability to even load these for a profile. 

**note this should be squashed and merged, not just merged**

ethan-nelson:
PR comment updated to reflect current status of PR.

ethan-nelson:
Putting this into #1192

PR #1148, opened 2018-07-29
Fix organisation tag to display only those with project list

ethan-nelson:
This is a redo of #1053 but with recent develop changes merged so that CI will properly run.


ethan-nelson:
Because the organizations and campaigns are near duplicates of each other in terms of code, I went ahead and extended all these changes to campaigns as well. @bgirardot i think this is ready for another review post #1053 when you get a chance.

willemarcel:
I confirmed the tags that are not part of projects are not listed by the API.

PR #1150, opened 2018-07-31
Replace JOSM Remote Control code with iframe version, which is not af…

twpol:
…fected by browser security in the same way

Note that I've only tested this in Edge so far, but given this is the same approach (although not identical code due to jQuery vs AngularJS) as osm.org itself, which I believe works quite reliably, I do not expect to have issues with other browsers.

### DIFF
--- a/client/app/project/project.controller.js
+++ b/client/app/project/project.controller.js
@@ -1258,22 +1258,22 @@
                         layer_name: encodeURIComponent('Task Boundaries #' + vm.projectData.projectId + '- Do not edit or upload'),
                         data: encodeURIComponent('<?xml version="1.0" encoding="utf8"?><osm generator="JOSM" upload="never" version="0.6"></osm>')
                     }
-                    var isEmptyTaskLayerSuccess = editorService.sendJOSMCmd('http://127.0.0.1:8111/load_data', emptyTaskLayerParams);
-                    if (!isEmptyTaskLayerSuccess) {
-                        //warn that JSOM couldn't be started
-                        vm.editorStartError = 'josm-error';
-                    }
+                    editorService.sendJOSMCmd('http://127.0.0.1:8111/load_data', emptyTaskLayerParams)
+                        .catch(function() {
+                            //warn that JSOM couldn't be started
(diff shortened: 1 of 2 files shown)


ethan-nelson:
Hi @twpol--thanks for your PR! I will have time to test this over the weekend and get back to you. Cheers!

ethan-nelson:
My one question is whether we want to increase the timeout time period to be a bit longer.

nrotstan:
At least on Chrome for Mac, the iframe load event appears to always fire regardless of whether JOSM is running or not. This leads to `sendJOSMCmd` always resolving with success even when communication with JOSM has failed. And since the request is cross-site, the iframe document cannot be inspected to see any actual response from JOSM for further follow-up.

@twpol and @ethan-nelson, was the impetus for switching from ajax to an iframe to support older browsers without CORS support? JOSM Remote Control supports jsonp as a fallback for such cases (there is [example code](https://wiki.openstreetmap.org/wiki/JOSM/RemoteControl#version_command) on the OSM wiki).

I'm happy to do the work to switch back to using ajax with a jsonp fallback for older browsers if that is what drove this change. Thanks!

twpol:
@nrotstan The problem was that the AJAX did not work in anything except Chrome - there were bug reports for Firefox and Safari, and I'd just encountered the same thing in Edge. It's been a few weeks since I did the investigations but I believe the other browsers (Edge, Firefox, Safari) were all blocking the AJAX because it was HTTP and the host page was HTTPS.

So this is not about older browsers at all, but it may still be about CORS being implemented differently in Chrome, Edge, Firefox, and Safari.

The new iframe code comes from OpenStreetMap.org itself (adjusted to support multiple commands), which was tested to work in Chrome, Edge, and Firefox (Safari still stubbornly refuses to work on osm.org or here), and kept the same error handling as there. Does Chrome on Mac exhibit the same "always success" when trying to open JOSM from osm.org without JOSM running?

nrotstan:
Thanks, @twpol, osm.org does not appear to work correctly either for Chrome for Mac (the failure box never appears if JOSM isn't running). It does work for Safari and Firefox.

Issue #1045, opened 2018-02-08
API Error unlocking task - "multiple rows were found for one()"


rustyb:
See -> https://tasks.hotosm.org/project/3885?task=75

If we try to unlock the task we get a 500 error from the DB in the frontend indicating "Tasking API Error...multiple rows were found for one()".

Just flagging in case the error pops up elsewhere.

smit1678:
@rustyb Was this on that specific task, 75? or seeing this on other tasks in this project? I'm seeing that 75 was unlocked. 

![hot tasking manager 2018-02-08 13-54-41](https://user-images.githubusercontent.com/796838/35976309-9f41cb4e-0cd7-11e8-8e90-66b43b0d7874.png)

rustyb:
It was on task 75 only, we waiting for it to be automatically unlocked in the end.

bgirardot:
We have seen this before I think during development, but it was fixed.

We had to run a script from @hunt3ri to correct the state some of them were left it.

I also got a report of this recently too that I will find.

@hunt3ri Do you remember this and the script that you asked David to run to fix the tasks that could not be unlocked? I think we need to run that script again and look for the underlying cause.

bgirardot:
![image 1](https://user-images.githubusercontent.com/265831/35977900-8d9d1954-0ce4-11e8-905d-f881e3e58358.png)

https://tasks.hotosm.org/project/3603

and

https://tasks.hotosm.org/project/4021?task=9#bottom

bgirardot:
Oh, now that I look, these are slightly different errors, one is "no row found" and one is "multiple rows found"

bgirardot:
Here is a Gist of recent errors exactly like this one and similar to it:

https://gist.github.com/bgirardot/c378deb4b6dc725ab71fec1407133724

sunidhiraheja:
@bgirardot Can we see the rows that are returned on that query match?
If they're deleted for this particular instance (on unlocking), maybe we could log them for future occurrences.
I think this will be of great help in debugging.

smit1678:
@sunidhiraheja Did digging into the NewRelic logs and error reporting help here? 

From what I can see, it looks like the function `server.api.validator_apis:unlocktasksaftervalidationapi` is the slow point that is related. The database query returns quickly but the function seems to consistently takes about 3 minutes to respond.

ethan-nelson:
[...]

Issue #992, opened 2017-12-12
Splitting tasks loses previous comments


shiller27:
Hi, 

it seems that when a tile is split, all the comments made on the 'tile of origin' disappear. This used not to happen in TM2. 

I left a lengthy comment on a task I was invalidating and decided to split the tile afterwards. The smaller tiles now show "nothing happened yet" in the comment section, and as far as I can make out my detailed feedback re mistakes/omissions will only be seen by the one user I tagged - or am I missing something?

Thanks!
Sabine

shiller27:
Possibly connected to this is the fact that it is no longer possible to add a comment to a tile one has previously mapped/validated. It just happened to me that the comment I had written for a tile to be invalidated disappeared after I clicked 'invalidate' (not sure was that my mistake or a glitch). However, I couldn't rewrite the comment after invalidation. Again, in TM2  the ability to add comments to a tile was independent of one's mapping/validation involvement in that tile - definitely better!

ethan-nelson:
This should be fixed and is now available on the live instance. Please let me know if you are still running into problems. Cheers!

shiller27:
Brilliant, all working great :) Thanks!

PR #1156, opened 2018-08-11
Keep task history after split and show split actions in history

### LINKED ISSUE #865: Task history lost on split
Tiles are split 4 ways. In the old task manager the edit history and comments would propogate to the four child tiles after a split. In the new task manager this does not happen and all history and comments are lost.

ethan-nelson:
This PR fixes the issue of task history lost during a task split. That's because the old task is deleted during a split and there is presently no transfer mechanism. In this PR, a clone_history function is introduced to loop over the history of the parent task and add task_history events in each of the children tasks. (N.B. this differs from the approach in TM2, where the original task and its history were maintained and children inherited task history events of their parent. We could maybe introduce this here instead to save some DB space for tasks, but it is not possible to recover the history for any task split in TM3 up to the merge of this PR.)

A problem arises however whenever we clone the full history of the parent task. Indeed, we copy every event, _including the lock of the parent task_. That means the user will, by default, also lock each of the child tasks. This problem is confounded by a user not being able to "unlock" a mapping task if it has not been set as being LOCKED_FOR_MAPPING in its task_status. (Keep in mind here that task_history--a table of events tied to a task by id--and the task_status column of a given task are separate.) We must bequeath the task_history value of LOCKED_FOR_MAPPING from the parent task to the child task, and we do so by [setting the value manually](https://github.com/hotosm/tasking-manager/compare/develop...ethan-nelson:retain-history-on-split?expand=1#diff-41e9aae54fefb69558fbbff6e598a0b1R127).

Okay, so we're making progress; we don't really get `more than one row found for one()` errors since our ducks are in a row regarding the status and history. But we need to indicate to users that the split has occurred. Sure, you can click around to the surrounding tasks and see that the history is the same among them and deduce a split was there. But, I like rigor, so a new task state has been included to indicate a state change caused by a split. This allows the frontend to pick up on that fact and include it in the history along with the transferred history:

split history--that sounds like a breakup song

Feedback needed, but please have mercy. Yes there should probably be and likely will be tests added.

Fixes #865, #992. Insights from the db experience added to #1045.

### DIFF
--- a/client/app/project/project.html
+++ b/client/app/project/project.html
@@ -956,6 +956,8 @@
                                                     href="/user/{{ item.actionBy }}">{{ item.actionBy }}</a></span>
                                             <span ng-show="item.action === 'STATE_CHANGE' && item.actionText === 'READY'">{{ 'Marked as ready by' | translate }} <a
                                                     href="/user/{{ item.actionBy }}">{{ item.actionBy }}</a></span>
+                                            <span ng-show="item.action === 'STATE_CHANGE' && item.actionText === 'SPLIT'">{{ 'Split by' | translate }} <a
+                                                    href="/user/{{ item.actionBy }}">{{ item.actionBy }}</a></span>
                                         </div>
                                         <div>
                                             <span ng-bind-html="item.actionText" ng-show="item.action === 'COMMENT'"></span>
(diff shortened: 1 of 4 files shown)


ethan-nelson:
One issue I already see:

Normally, if a user invalidates a task, that task is marked as invalidated--even if another user checks out the task and unlocks it--until someone marks it mapped. This behavior is not retained in splitting, i.e. if someone invalidates a task and then splits it, the children tasks will go back to their non-shaded ("ready to map") state. The latter is perhaps the expected behavior to me in any case, but maybe not to others. Whatever we decide on, it should be consistent between all cases.

ethan-nelson:
Closing in favor of #1178

PR #1160, opened 2018-08-12
Restrict validating to JOSM

### LINKED ISSUE #893: Add option to restrict editing programs allowed during validation
and if you can't why is the option there?

ethan-nelson:
This comes from discussion in #893 and indeed fixes #893. This PR makes the only editor program shown on validating and multi-task validating JOSM.

We split out variables to hold the mapping and validating editor choices because the TM saves user preferences on the preferred map editor program. That interferes with the end goal of this PR by adding other elements in the editor choice menu if the user's preferred editor is not JOSM.

Please close if the discussion in #893 is no longer the preferred approach.

### DIFF
--- a/client/app/project/project.controller.js
+++ b/client/app/project/project.controller.js
@@ -56,7 +56,8 @@
 
         //editor
         vm.editorStartError = '';
-        vm.selectedEditor = 'ideditor';
+        vm.selectedMappingEditor = 'ideditor';
+        vm.selectedValidatingEditor = 'josm';
 
         //interaction
         vm.selectInteraction = null;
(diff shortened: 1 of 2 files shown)


bgirardot:
@ethan-nelson I think the option to restrict the editors for validation on a project by project basis would be best at the moment and commented on that thread. How hard is it to add it as an option to a project?

ethan-nelson:
Hi Blake, that's something that can probably be added as a field to the project settings, and then the javascript front end could hide certain editors depending on that setting. I'll refine the issue ticket a bit more and close this one out for now.

PR #1161, opened 2018-08-12
Add menu item styling

### LINKED ISSUE #751: dropdown items clickable area don't cover full width
When expand a dropdown menu, you can highlight each item when hovering them. However, the clickable area is limited to the text itself.

ethan-nelson:
Fixes #751 for the translation and user drop down menus at the top by adding a missing style definition.


ramyaragupathy:
menu drop
Looks good to me @ethan-nelson. Clickable area is not limited just to the text area.

PR #1166, opened 2018-08-17
Sort contribute list by number of active mappers

### LINKED ISSUE #1151: Add ability to sort Contribute list of projects by number of active mappers
In some cases it would be very useful to see what projects mappers are working on. 

Use cases are several:

As a validator I want to see where the most activity is so I can validate while people are online and mapping.

As a system administrator I would like to know what projects have the most active mappers while troubleshooting any system issues.

As a project manager I would like to know what projects are getting the most mappers so I can understand how to better draw mappers to my projects.

sunidhiraheja:
Fixes #1151

### DIFF
--- a/client/app/contribute/contribute.html
+++ b/client/app/contribute/contribute.html
@@ -152,7 +152,7 @@
                                     <div class="search-results-container">
                                         <div class="project-result__cards">
                                             <a href="/project/{{ project.projectId }}" class="project-result__card {{ project.status | lowercase }}"
-                                                 ng-repeat="project in contributeCtrl.results track by $index">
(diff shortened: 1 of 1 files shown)


ethan-nelson:
Hi Sunidhi, not sure if you saw, but I left a comment on the issue you mentioned regarding this feature. Ideally this would be a selectable feature in a drop down menu of sort types similar to TM2 so we don't lose priority sort by default.

sunidhiraheja:
Hi, just saw it sorry.
So since the DTO has the priority as a string, to show it on the frontend, we lose the enum ordering.
Should I store it as an integer and map it back to the enum label it in the frontend template?

PR #1170, opened 2018-08-26
Spoof task outline gpx file extension for iD

ethan-nelson:
Latest iD release doesn't load GPX files without an extension, so we add a fake one to the URL for now.


bgirardot:
I am not having any luck with this fixing things.

The PR might not be working correctly as my URL does not include the filename= on the end.

But if I edit a url myself to add that in a few different variations it still did not load the purple box :(

bgirardot:
the example link in slack does work not sure why my testing does not.

PR #1178, opened 2018-08-31
Preserve existing task history on task split

nrotstan:
* When splitting a task, copy any prior task history to each new task and add a comment indicating the task has been split.

* Restore Task.clear_task_lock to its prior behavior of simply clearing the task lock so that it's reusable outside of auto-unlock scenarios.

* Move the recently-added auto-unlock specific behavior from Task.clear_task_lock into a new Task.record_auto_unlock that delegates to clear_task_lock and then records the AUTO_UNLOCKED action.


ethan-nelson:
Hey Neil, I will take a look at this in the next two days. Thanks for your patience.

server/services/grid/split_service.py
```diff
@@ -123,6 +123,11 @@ def split_task(split_task_dto: SplitTaskDTO) -> TaskDTOs:
             new_task.project_id = split_task_dto.project_id
             new_task.task_status = TaskStatus.READY.value
             new_task.create()
+            new_task.task_history.extend(original_task.copy_task_history())
+            if new_task.task_history:
+                new_task.clear_task_lock() # since we just copied the lock
+            new_task.set_task_history(TaskAction.COMMENT, split_task_dto.user_id, "Task split")
```
ethan-nelson: Hi Neil, it's taken me a bit to digest your changes. It looks like things are working okay, but I have one request to get your feedback on. Would it be possible to create a new task status state for the split? That would help with database queries (which, could be done by filtering for comments with "Task split", but it would be easier searching for that specific task status integer on a column).

Something like https://github.com/hotosm/tasking-manager/blob/8e07e4ae861b1a80cdf1dffa8ddb7b7cca86f073/server/models/postgis/statuses.py#L34 could work in place of the comment. Then, to display the state on the front end, you could implement: https://github.com/hotosm/tasking-manager/blob/8e07e4ae861b1a80cdf1dffa8ddb7b7cca86f073/client/app/project/project.html#L959-L960.

What are your thoughts on that?

nrotstan: I think that makes a lot of sense, Ethan. I will implement that. Thanks!

ethan-nelson:
Thank you @nrotstan again!!

Boggedy:
Looks good in the current TM interface. Well done to all

PR #1187, opened 2018-09-22
Add reset all bad imagery option

### LINKED ISSUE #1171: Add to Project Admin - Action tab - Reset all bad imagery option
We occasionally get projects where folks do not know how to change imagery or they misunderstand the "Bad Imagery" option.

As a result we end up with projects with dozens of task squares marked bad imagery when in reality they are not bad imagery.

It would be nice to be able to just reset all the bad imagery task squares to 'ready to map'

sunidhiraheja:
Closes #1171

### DIFF
--- a/client/app/admin/edit-project/edit-project.controller.js
+++ b/client/app/admin/edit-project/edit-project.controller.js
@@ -369,6 +369,34 @@
             })
         };
 
+        /**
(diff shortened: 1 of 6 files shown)


ethan-nelson:
I will take a look at this in the next two days--thanks Sunidhi!

ethan-nelson:
Small merge conflict fixed

PR #1192, opened 2018-10-01
Fixes #792 where validation and mapping counts incorrect on profile

### LINKED ISSUE #792: User profile - stats, number of mapped/validated tasks in project list is incorrect

hunt3ri:
Hi folks

Thought I'd try and help out for Hacktoberfest 😄 

This PR should fix issue #792 The original query was somewhat naive and expected at least 1 mapped contribution on each project.  So in cases were users were only validating these counts were missed.

The fixed query looks much more complex, but essentially we're just joining the results from all mapping contributions for a user and all validation contributions.

My local testing showed the query to be performant on large datasets, and fixes cases where I only validated tasks 

image

### DIFF
--- a/server/models/dtos/user_dto.py
+++ b/server/models/dtos/user_dto.py
@@ -64,7 +64,6 @@ class MappedProject(Model):
     tasks_validated = IntType(serialized_name='tasksValidated')
     status = StringType()
     centroid = BaseType()
-    aoi = BaseType()
 
 
 class UserMappedProjectsDTO(Model):
--- a/server/models/postgis/user.py
+++ b/server/models/postgis/user.py
@@ -150,15 +150,36 @@ class User(db.Model):
     @staticmethod
     def get_mapped_projects(user_id: int, preferred_locale: str) -> UserMappedProjectsDTO:
         """ Get all projects a user has mapped on """
-        sql = '''select p.id, p.status, p.default_locale, count(t.mapped_by), count(t.validated_by), st_asgeojson(p.centroid),
+
+        # This query looks scary, but we're really just creating an outer join between the query that gets the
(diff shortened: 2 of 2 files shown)


ethan-nelson:
Hey Iain, thanks for the PR. I'm about to travel for SOTM-US, so it'll be a few days before I can review with DB.

Just out of curiosity, what are your thoughts about this strategy: https://github.com/hotosm/tasking-manager/blob/29b5fdb58f58abdcf7e37876da6cbc4ab8b26bcc/server/models/postgis/user.py I don't really know about the database load using one query versus the other but I can test that out as well when I get back. Cheers!

hunt3ri:
Hey @ethan-nelson I've got a feeling I started there, but the OR query didn't seem to handle cases where there were mapped and validated edits by the user on tasks within the same project.  I should be able to test against my test data, so will give it a try.

hunt3ri:
Hi @ethan-nelson I tested that approach, and while it looks like it SHOULD work it doesn't.  I think the problem is that the counts, for mapped are included if the user has only validated.   For the same test data if you look at the project 617, using the simple query you can see it returns a count of 118 for mapped and validated.  Where the correct answer is actually 0 and 118, and my more complex query calculates

image

So I'm afraid this time complexity wins.  Let me know if you have any more questions.

Thanks

ethan-nelson:
Alright, thank you a bunch for digging into this @hunt3ri ! I will take another look here in the next two days and get back to you with any finishing touches.

ethan-nelson:
Oh and I also added a sort component based on descending id number for _some_ prioritization of more recent projects near the top of the list.

PR #1200, opened 2018-10-04
Add translation repo to CONTRIBUTING

ethan-nelson:
Adding a line about Transifex to clarify the workflow for string translations.


smit1678:
Going to open up a ticket about making this a little more automated with Circle so we're always up to date on translations.

PR #1201, opened 2018-10-05
Allow commenting on task without locking it

nrotstan:
* Add a comment box to tasks that is always present and can be used to comment on a task without first locking it for mapping or validation

* Extract comment box markup into a comment-box.html template to reduce duplication of markup

* Add new server API endpoint for adding a stand-alone comment to a task

standalone_comment


ethan-nelson:
Thank you @nrotstan !!

PR #1202, opened 2018-10-05
Update/translations

smit1678:
This PR updates existing translations from Transifex plus adds the following new translations:

* Vietnamese (vi)
* Tamil (ta)
* Galician (gl)
* Persian Iran (fa_IR)

The new translations haven't been 100% translated yet but have started with the exception of Tamil which is 0% translated.


smit1678:
Made a mistake here. Branched from `master`. 🤦‍♂️  Closing and will fix.

PR #1203, opened 2018-10-05
Update/transifex

smit1678:
This PR updates existing translations from Transifex plus adds the following new translations:

* Vietnamese (vi)
* Tamil (ta)
* Galician (gl)
* Persian Iran (fa_IR)

The new translations haven't been 100% translated yet but have started with the exception of Tamil which is 0% translated.

I didn't rebase before I pulled the last few times locally and so we should squash and merge to keep the history clean.


smit1678:
@ethan-nelson @ramyaragupathy Adding this shortly but in terms of translation process, do we want to offer incomplete or not reviewed translations as options? Or do we want to wait for fully reviewed and fully translated options? Related to #1214 where Finnish is fully translated but not reviewed within Transifex: https://www.transifex.com/hotosm/tasking-manager-3/dashboard/.

ethan-nelson:
As of now it looks like we've been adding languages that do not have any reviews but have been fully translated (French, Dutch, Czech, Japanese, Polish, Portuguese, etc: https://www.transifex.com/hotosm/tasking-manager-3/dashboard/). Ideally it would be good for all strings to be reviewed, however the repo hasn't been too active with reviews--the review line on the graph of the dashboard is pretty flat. Part of the reason may be we only have people requesting to be translators and not reviewers.

smit1678:
Added config, thanks for the reminder on that @ethan-nelson. I also issued a ticket to track the reviewing: https://github.com/hotosm/tasking-manager/issues/1218. Will merge this.

PR #1204, opened 2018-10-06
Enhance readability of task history

nrotstan:
* Change display of task history to a table structure, grouping together consecutive history entries from the same user with visual dividers between groups

* Show icons or color-coded status markers on entries based on status to make the history easier to scan

* Use table structure for Last Activity section of project Activity and Stats page, including color-coded status markers for entries

* Support rendering of Markdown in comments
new_history_layout

### DIFF
--- a/client/app/components/project-chat/project-chat.html
+++ b/client/app/components/project-chat/project-chat.html
@@ -5,7 +5,7 @@
                 <strong>
                     <a href="./user/{{ message.username }}" target="_blank"/>{{ message.username }}</a>
                 </strong> <span class="metadata" am-time-ago="message.timestamp | amUtc | amLocal"></span></p>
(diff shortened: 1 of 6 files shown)


ramyaragupathy:
Current view|Updated view
------------|--------------
image|image

 and the markdown feature for comments works 👍 
image

ethan-nelson:
Great, thank you @nrotstan a bunch! I think this is a very positive change, especially the grouping by user.

ethan-nelson:
It looks like we did lose the activity length "locked for X minutes" where available, but I rather get the major changes pushed through for now so others can check out the changes on the dev server. I'll file an issue with that so we can remember (perhaps another good and small change for Hacktoberfest).

nrotstan:
Thanks for catching that, Ethan!

PR #1206, opened 2018-10-06
Added icons to buttons available when mapping.

jillhkm:


ethan-nelson:
I would guess that we probably can't style the emoticons for coloring like we could for glyphicons. Another alternative could be changing the color of that button...I'm personally not a big fan of the orange for two different buttons in a single action area.

ramyaragupathy:
Yeah @ethan-nelson - agree with the color change here. I've pushed the same here in the layout PR here: https://github.com/hotosm/tasking-manager/pull/1220

jillhkm:
Certainly, I agree the color clash is unappealing. I'll take a look at the icons!

jillhkm:
Okay, I switched to icons. Hopefully, the new look will work better visually.

smit1678:
@jillhkm This is great! 🎉  Thanks for the adjustments. I like the simplicity of the icons a lot better. Let's get these merged into develop so we can have them deployed on staging to be tested. If there is some further feedback on size or specific icon type we can discuss then.

Issue #1168, opened 2018-08-22
Clicking the "Contribute" tab should reload the page with no filters applied


bgirardot:
It is often a UI navigation comfort that you can always click on a navigation element and have it take you to that page, even if you are already on that page. It will generally reload the page so you can start over on something.

The TM should do that with the Contribute link. It should clear all the current searches and filters and display the page as if you had just clicked to land on it for the first time.

also see issue #613, No way to clear searches, this is related. If there was a "clear" button, people like me would not click on the Contribute tab trying to get the page to reload and clear our search / filters.

matizzy:
Hi guys, can I take on this issue?

ethan-nelson:
Hi @matizzy sure, I will mark it as in progress. Please let me know if you have any questions or troubles.

Redarcher9:
Hii, is someone still working on this issue?

matizzy:
@Redarcher9 Hi yes sorry, I am. I had a busy week at work but I'm gonna finish it up today or by tomorrow morning.

matizzy:
https://github.com/hotosm/tasking-manager/pull/1208

I opened a PR for this issue

xamanu:
This works now on the new frontend in TM4.

PR #1208, opened 2018-10-07
pressing contribute will now reset the search params

matizzy:
issue: https://github.com/hotosm/tasking-manager/issues/1168

**Problem**: 
Clicking on contribute only reset the URL back to `/contribute` but did not reset the actual underlying search terms. 

**Solution**:
Add a broadcast event to the click so that the child (contribute) can listen to the press and reinitialize the search. 

**_Note_**
This is my first PR to this project so I am a newbie to it. 

Additionally, when it comes to frontend, I am most a React dev. Last time I touched angular was over 2 years ago. If I made a bad design choice with my approach to this issue I would love feedback.

### DIFF
--- a/client/app/contribute/contribute.controller.js
+++ b/client/app/contribute/contribute.controller.js
@@ -10,39 +10,40 @@
         .controller('contributeController', ['$scope', '$location', 'mapService', 'searchService', 'projectMapService', 'tagService', 'languageService', contributeController]);
 
     function contributeController($scope, $location, mapService, searchService, projectMapService, tagService, languageService) {
-
         var vm = this;
+        
+        function init (){
+            vm.results = [];
+            vm.vectorSource = null;
 
-        vm.results = [];
-        vm.vectorSource = null;
+            // Default to grid view
+            vm.resultsView = 'grid';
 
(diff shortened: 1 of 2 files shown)


matizzy:
Sidenote, note super important, but I made the commit from the wrong account so it shows 2 participants :/

ethan-nelson:
Hey @matizzy thanks for your PR! I too am not an Angular wizard by any means. Here is the approach I would consider taking:

1) create a reset function that sets the search parameters to their default values like https://github.com/hotosm/tasking-manager/blob/develop/client/app/contribute/contribute.controller.js#L27-L37
2) call that reset function and then try running activate or search again

You may need to play around with the exact variables that need to be reset, but it should be ones that are called in the actual search function.

ethan-nelson:
Hi @matizzy I am going to go ahead and close this PR for now since we have not heard from you after the initial filing. We are still very interested in your contribution if you want to follow some of the advice here to get it working, but in the mean time I am trying to clean up the PR queue a bit. If you have any questions, please do not hesitate to get in touch or reopen this PR! Cheers, Ethan

PR #1210, opened 2018-10-07
Remove unused file write in test

lesserj:
Data is written out to a file during the test, probably for local debugging.  This breaks tests on machines that do not have a enelson home directory however.  Removing it fixes the test


lesserj:
Superseded by #1211

PR #1211, opened 2018-10-08
Fix tests

lesserj:
This fixes the couple dozen server side failing or broken tests.   All tests are passing consistently now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,6 +63,7 @@ As the project is open source we have to keep secrets out of the repo.  You will
 * **TM_CONSUMER_KEY** - This is the OAUTH Consumer Key used for authenticating the Tasking Manager App in OSM
 * **TM_CONSUMER_SECRET** - This is the OAUTH Consumer Secret used for authenticating the Tasking Manager App in OSM
 * **TM_SMTP_HOST** - The hostname for the SMTP server that is used to send email alerts
+* **TM_SMTP_PORT** - The port number for the SMTP server that is used to send email alerts
 * **TM_SMTP_USER** - The user for the SMTP server that is used to send email alerts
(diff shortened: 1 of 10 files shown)


ethan-nelson:
Thanks @lesserj for this PR! I am working through it at the moment. I am going to hold off on merging, though, until I can coordinate with our sysadmin that the correct email port is set up in our deployment so that things don't break when I merge :) I will get back to you in a day or two

lesserj:
After a little more investigation, it appears that CicleCI reports tests that fail, but not tests that error out.

ethan-nelson:
Just to make sure, are you talking about errors such as:

```[2018-10-08 22:31:28,535] ERROR in mapping_service: Attempting to unlock a task owned by another user```

that print when running nosetests? I believe those are supposed to be intended behavior the API prints stemming from places like:

https://github.com/hotosm/tasking-manager/blob/e1d50c2c714056ca898ff9beb3f4c8d029cdf181/tests/server/unit/services/test_mapping_service.py#L90-L99

If it's the case that there is e.g. a test that is written wrongly and throws an Exception itself and nose doesn't pick up on that, then we should probably add those bash flags.

Also, I sent a message to the sysadmin to check on our deployment, so once that is confirmed this will be good to merge.

lesserj:
I definitely mean the latter where a mistake in the test throws an exception that isn't expected.  There were I believe 26 errors thrown in the tests that I fixed above when run locally, none of which were reported by CircleCI.

ethan-nelson:
Got the okay on the environmental variables for our deployment, so I'm going to merge and see how it goes. Thank you @lesserj !!

PR #1216, opened 2018-10-08
Split non square tasks

lesserj:
This allows non-square tasks to be split.  Non square tasks are created when trimming a project by the AOI.  This feature continues to track the x,y and zoom of a task, regardless of whether or not that task has been trimmed.  When splitting the task, it calculates what the new x,y and zoom value would be, then uses that result to intersect with the current task.  Resulting tasks will be based on the new x,y,zoom grid, so may not be equal in size, depending on how much area of the existing task falls within each quadrant of the new grid.  If the task does not intersect with one of the quadrants at all, it will not be created, so you may end up with fewer than 4 new tasks when intersecting.

split-non-square-tasks 2

This will only split non square tasks on projects created after this is merged in.  Existing projects will behave the same, without being able to split non-square tasks.

Addresses #1086 

cc @timusumisu


ethan-nelson:
I will be reviewing this PR this week--thanks!

ethan-nelson:
Hi again @lesserj I have not forgotten about this--sorry for the silence. I want to test one more thing before the approval, but we are frozen on PRs for the time being anyway until we get the last cluster of changes merged into master. Thanks for your patience!

ethan-nelson:
Here is an example on the staging server showing the ability to split for edge cases! https://tasks-stage.hotosm.org/project/5532

Thank you again @lesserj  ! :clap:

lesserj:
Awesome.  Thanks for making those changes @ethan-nelson .

PR #1217, opened 2018-10-08
Unlock all expired actions in task history. Fixes #1185

### LINKED ISSUE #1185: Automatically Unlocking Tasks
There are some tasks that are 'automatically unlocking', such as https://tasks.hotosm.org/project/5236?task=1554 and https://tasks.hotosm.org/project/5025?task=1365
auto-unlock

This is causing many contributors to try locking these tasks many times, and projects to not be 'complete'.

Hulios:
Fixes #1185 

Auto unlock wrongly assumed that:
- Only last action in history record is older than 2 hours 
- When task is in locked state, last action is always LOCKED_FOR_VALIDATION or LOCKED_FOR_MAPPING

I've added logic to update all locked history tasks that don't have action_text. Additionally it checks if action is older than 2 hours, so when someone locks task for mapping, he won't be immediately unlocked, like described in issue.

Also acquiring last action is now immune to mixed order of history records. In situation when someone adds comment when task is locked, it won't break auto unlocking.


ethan-nelson:
Hi @Hulios thank you for your PR! I should be able to take a look at this in the next 48 hours.

ethan-nelson:
Hi again @Hulios ! This is good work and should alleviate the issues we have of tasks being stuck in immediate unlocking. There are a few small conflicts with another PR that I will resolve tomorrow and do a final review before merging. Thank you!

server/models/postgis/task.py
```diff
@@ -72,7 +72,7 @@ def delete(self):
         db.session.commit()
 
     @staticmethod
-    def update_task_locked_with_duration(project_id: int, task_id: int, lock_action):
+    def update_task_locked_with_duration(task_id: int, project_id: int, lock_action: TaskStatus, user_id: int):
```
ethan-nelson: I know it would be good to match the order of the calls, but it was easier for now to switch back to the original style given changes in #1223

Issue #1181, opened 2018-09-05
Re-arrange buttons on TM3  map page


FransS1945:
Re arrange buttons on Main page of TM3 

https://tasks.hotosm.org/project/5025?task=89

First button shows "marked as completely mapped."
Exchange this with the "stop mapping" button
Why?  I see often tasks which has been marked as "completely mapped" despite there are still a lot of objects to be mapped. 
Fact is that many mappers can contribute on a task and stop mapping, leaving the task open for other mappers. But there is just 1 mapper who declare a task as "complete mapped" 
So this button should not be positioned on top should I think.
When I took a look to the mappersprofile who declared a task as complete mapped, mostly the are new mappers.
I questioned several mappers about the why, they responded that they just wanted to stop mapping.
It seems that they have no idea what this button means. 
Changing the "stop mapping" button to the first line can solve this. Change colour as well ?

![screen shot 09-05-18 at 02 17 pm](https://user-images.githubusercontent.com/31804031/45092739-a4f95280-b116-11e8-9d50-bd25945f3fae.JPG)

smit1678:
@FransS1945 thanks for opening this ticket. This has been seen before and I think we can improve the layout with some simple changes. 

Adding some initial guidance that I think would be an improvement to the experience here. This is from a brief conversation between @ramyaragupathy and myself about how we could reorganize. 

<img width="474" alt="wireframe cc - minimal wireframing tool 2018-10-01 12-58-25" src="https://user-images.githubusercontent.com/796838/46272135-ea743880-c579-11e8-99d0-fcbf188a68e3.png">

The thinking here is that we can add some more text instructions and then separate the type of actions. For the stop and mark as completely mapped, we order them according to flow. You may need to stop mapping before you mark it completely mapped. For all buttons, we add a simple icon that helps users understand what the action is. 

✋ icon for Stop Mapping (a hand or stop sign icon)
💯 icon for Mark as Complete Mapped (a 100% icon)
❌ for Mark as Bad Imagery (some type of X icon)
🔲 for Split Task (some type of square grid icon)

Other comments: set to a single color and use order and icons to communicate hierarchy and action.

Open for discussion or other ideas for how we could implement these improvements.

goldfndr:
For the Split Task icon, how about  :heavy_division_sign: (heavy division sign)?

* Universal division symbol
* Similar look to how the grid square gets divided

FransS1945:
Yes, for my intuition it looks better.
Thaks,

[...]

PR #1220, opened 2018-10-09
Rearrange buttons for mapping tab

ramyaragupathy:
Per discussion in #1181 - there's already a [PR to add icons](https://github.com/hotosm/tasking-manager/pull/1206). This PR is for modifying the buttons layout.

image


ramyaragupathy:
Still testing the functionality of the buttons locally. Will request a review once the local testing is complete.

ramyaragupathy:
Good for review.

smit1678:
Nice @ramyaragupathy. One quick response to the image. I saw that you added two comment boxes. Should we just keep one to keep it simple for the user? Or did you have some other thinking there? 

tasking-manager 2018-10-10 10-42-10

ramyaragupathy:
Agree that it is simple to keep one comment box. My thinking was to have one comment box next to each action buttons. I understand it is confusing and not useful for the end-user experience. Removing it.

ramyaragupathy:
Screenshot after the merge
image

Blake-Loveland:
What do you think about making the Mark Completely mapped button the same color as a mapped task to help reinforce that color meaning?

smit1678:
To add to the discussion here, I think for now we should stick with the two types of button colors that we have across the site: brown or orange. If we introduce another color, I feel like this might just look more confusing. My suggestion is that if we think orange sticks out too much, we go with brown so all the mapping actions have equal weight.

PR #1222, opened 2018-10-09
Restore the length of events in history view.

### LINKED ISSUE #1219: Include activity time length text in updated task history UI
As of the prior stable TM release, task history entries show the length of time for some events, e.g.

The latest overhaul in #1204 has changed the history view to group successive events by a user so that it is easier to digest the history:

However, the length of the event was not implemented into that.

This issue requests someone re-implement the `X for Y time` string for mapping, validating, auto-unlock mapping, and auto-unlock validating entries in the updated UI. The code diff in the PR mentioned above includes some code that could be repurposed based on the new format.

PierrePIRONIN:
Fixes #1219 by restoring the length of an event.

### DIFF
--- a/client/app/project/project.html
+++ b/client/app/project/project.html
@@ -970,6 +970,10 @@
                                               </td>
                                               <td>
                                                   <span am-time-ago="item.actionDate | amUtc | amLocal"></span>
+                                                  <span ng-show="item.action === 'LOCKED_FOR_MAPPING' && item.actionText"> for {{ item.actionText |  amDurationFormat : 'minute'}} </span>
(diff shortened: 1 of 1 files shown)


PierrePIRONIN:
I think this PR do the trick but I tried to deploy the application locally to ensure my code was good.
But the server is tricky with local database to install, data to fill, etc, have you an environment where the server is deployed where I can check my code on the client ?

ramyaragupathy:
Hey @PierrePIRONIN @ethan-nelson  I got a chance to test this code locally and the changes looks good. Time intervals are similar to the exisitng event specs.
image
image
image

PierrePIRONIN:
Thanks @ramyaragupathy 👍

ethan-nelson:
Hi @PierrePIRONIN  thank you for your help! Regarding the local setup, our documentation could be updated a bit--including with a sample seeded database. I will look into adding some of that info this week in hopes of decreasing the spin up time.

PR #1223, opened 2018-10-10
Add robustness to updating task history

### LINKED ISSUE #1045: API Error unlocking task - "multiple rows were found for one()"

hunt3ri:
Second #Hacktoberfest contribution 😄 Hopefully fixes #1045 and #668 

This PR adds robustness to updating the task history and hopefully resolves #1045 and #668.  Reading thru the issue I agree with @ethan-nelson that the underlying cause of this issue is most likely one or more race conditions.

Stepping thru the code, I think the most likely spot this errors are being generated is update_task_locked_with_duration method, where we only expect to find one row

This PR doesn't seek to resolve the race conditions (as they are notoriously hard to replicate), however, it adds robustness to the update task history code, which should mitigate the error allowing users to continue mapping.

It does this in 3 ways:

* We now supply the user_id when we unlock this task.  If somehow 2 users have managed to lock the same task, we should only attempt to update the history record for that we created.
* If we get a NoResultFound error - the users history record has gone forever, so there's nothing we cand do other than return, this should stop users seeing No row was found for one()
* Passing user_id should hopefully fix the multiple rows issue, however if not and we get MultipleResultsFound error, we'll now delete the oldest task history records until only one remains for the action the user is attempting, on the basis that the latest record was the last thing the user was attempting to do.  This should fix Multiple rows were found for one()

### DIFF
--- a/server/models/postgis/task.py
+++ b/server/models/postgis/task.py
@@ -3,6 +3,7 @@ import datetime
 import geojson
 import json
 from enum import Enum
+from sqlalchemy.orm.exc import NoResultFound, MultipleResultsFound
 from sqlalchemy.orm.session import make_transient
 from geoalchemy2 import Geometry
 from server import db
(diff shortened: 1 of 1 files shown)


ethan-nelson:
Hi Iain, sorry for the delay. I will take a look at this this evening California time...thanks!

hunt3ri:
No worries, @ethan-nelson , wish I was on Calafornia time, I'm currently on cold Autumn Scottish time 😉

ethan-nelson:
Hey, it's been getting down to about 8*C overnight here--not all bright and sunny.

This looks good to go and will be helpful in minimizing the issues that have been persisting for a while. But, I am going to wait to merge this until tomorrow because there these changes are in the vicinity of those in #1217 and I want to make sure everything meshes okay (that PR fixes a recently introduced bug of inability to lock for _any_ time, as opposed to an inability to unlock).

Issue #1152, opened 2018-08-03
Show exact time in "time ago" alt text


ethan-nelson:
The prior version of this software displayed the exact date and time whenever hovering over the "X ago" under a given history element. We should try to implement the same here.

I previously attempted this in https://github.com/hotosm/tasking-manager/pull/1155/files, but the area where this is added has changed a bit. So this issue involves taking the changes in #1155 and adding them to the correct locations in the updated files.

stufinn:
I'd like to give this a shot!

ethan-nelson:
Fixed with #1226

PR #1226, opened 2018-10-12
Re-add exact time to history, chat re: issue #1152

stufinn:
Addresses issue #1152 : Show exact time in "time ago" alt text.

Being really new to contributing, I'm not sure how to preview and test the changes in the changes in the application.

### DIFF
--- a/client/app/components/project-chat/project-chat.html
+++ b/client/app/components/project-chat/project-chat.html
@@ -4,7 +4,7 @@
             <p class="space-bottom--none">
                 <strong>
                     <a href="./user/{{ message.username }}" target="_blank"/>{{ message.username }}</a>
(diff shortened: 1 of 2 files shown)


ethan-nelson:
Hi @stufinn this works great, thanks! We don't really have a way to demo things right now in a centralized location. We've generally relied on people spinning up a local instance and developing things there. I'll make a note to look into options--if you have seen other projects that do have some type of adaptable instance based on a PR, please past it along. I'm also hoping to finish up a video cast this weekend of installing the TM from scratch. Thanks again!

stufinn:
Awesome! @ethan-nelson good to know it works OK.  I'll be sure to keep an eye out for that video cast. This seems like a really cool project, so would be happy to contribute more in the future. Cheers

Issue #1224, opened 2018-10-12
Enable upload of `.osm` files for new projects


ethan-nelson:
Right now we direct users to convert `.osm` files to geoJSON for importing/uploading project geometries. We should explore the possibility of adding `.osm` support natively.

PR #1233, opened 2018-10-13
Added support for importing osm files

Hulios:
Related to #1224 

Using https://github.com/tyrasd/osmtogeojson for converting osm to geojson

Translations for 'Import a GeoJSON, KML, or zipped SHP file.' were not changed. It can be done separately, if this solution is accepted.

### DIFF
--- a/client/app/admin/create-project/create-project.controller.js
+++ b/client/app/admin/create-project/create-project.controller.js
@@ -346,6 +346,12 @@
                             setUploadedFeatures(uploadedFeatures);
                         });
                     }
+                    else if (file.name.substr(-3).toLowerCase() === 'osm') {
+                        // Use the osmtogeojson.js library to read the osm (with GeoJSON as output)
+                        var xml = (new DOMParser()).parseFromString(data, 'text/xml')
+                        uploadedFeatures = geospatialService.getFeaturesFromGeoJSON(osmtogeojson(xml, {flatProperties: true}));
+                        setUploadedFeatures(uploadedFeatures);
+                    }
                 };
                 if (file.name.substr(-4).toLowerCase() === 'json') {
                     fileReader.readAsText(file);
@@ -355,6 +361,9 @@
                 }
                 else if (file.name.substr(-3).toLowerCase() === 'zip') {
                     fileReader.readAsArrayBuffer(file);
+                }
+                else if (file.name.substr(-3).toLowerCase() === 'osm') {
+                    fileReader.readAsText(file);
                 }
                 else {
                     vm.isImportError = true;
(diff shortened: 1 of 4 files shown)


willemarcel:
Thanks again, @Hulios!

I tested importing those XMLs:
https://www.openstreetmap.org/api/0.6/relation/362428 and https://www.openstreetmap.org/api/0.6/way/117898424
I received that error:
screenshot_2018-10-21 hot tasking manager

If the osm file contains elements that are not polygons or has polygons with intersections, it doesn't work.

It only worked when I drew an area on JOSM, so I think it's not so useful to have it working that way. I would prefer to have something like we have on the OSMCha filters: https://osmcha.mapbox.com/filters (it uses the nominatim API to get the polygon of an administrative area). What do you think, @ethan-nelson ?

ethan-nelson:
I think the key here is the OSM file needs to be self-describing, i.e. all nodes that make up the multipolygon or its features should be defined within the .osm file. The example would be more like pulling out a component from overpassAPI with the `>;` syntax. Using geojson.io would also probably require those features, as I believe TM2 also required.; 

I tested it with a simple .osm file of a few nodes in a multipolygon and it worked:

`<osm><relation id='1'><member type='way' ref='1' role='outer'><nd lat='0' lon='0' /><nd lat='0' lon='1' /><nd lat='1' lon='1' /><nd lat='1' lon='0' /><nd lat='0' lon='0' /></member><member type='way' ref='2' role='inner' /><tag k='type' v='multipolygon' /></relation><relation id='2'><member type='way' ref='3' role='outer'><nd lat='1' lon='1' /><nd lat='1' lon='2' /><nd lat='2' lon='2' /><nd /><nd lat='1' lon='1' /></member><tag k='type' v='multipolygon' /></relation></osm>`

ethan-nelson:
I think the key here is the OSM file needs to be self-describing, i.e. all nodes that make up the multipolygon or its features should be defined within the .osm file. The example would be more like pulling out a component from overpassAPI with the `>;` syntax. Using geojson.io would also probably require those features, as I believe TM2 also required.; 

I tested it with a simple .osm file of a few nodes in a multipolygon from the nodejs module's repo and it worked:

`<osm><relation id='1'><member type='way' ref='1' role='outer'><nd lat='0' lon='0' /><nd lat='0' lon='1' /><nd lat='1' lon='1' /><nd lat='1' lon='0' /><nd lat='0' lon='0' /></member><member type='way' ref='2' role='inner' /><tag k='type' v='multipolygon' /></relation><relation id='2'><member type='way' ref='3' role='outer'><nd lat='1' lon='1' /><nd lat='1' lon='2' /><nd lat='2' lon='2' /><nd /><nd lat='1' lon='1' /></member><tag k='type' v='multipolygon' /></relation></osm>`

PR #1236, opened 2018-10-15
Don't jump to the top of the page after clicking on tab header

### LINKED ISSUE #1225: Switching from validate to map causes jump
If you have the validating tab open and select a project that is in a mapping state, the task will direct you that it can't be validated. Here, a user will possibly click the map tab to try and map the task. When you click on the map tab, though, the browser jumps to the very top of the page instead of the top of the contributing window. This is in Firefox 59 on Ubuntu.

Hulios:
Fixes #1225 

Removed href="#"
Removed excessive div


ethan-nelson:
Thank you for all of your contributions @Hulios !

PR #1245, opened 2018-10-19
[WIP] Version 3.0.4 Release

dakotabenjamin:
Marked as WIP because I want to fully summarize changes but there are a lot! Most of the changes are from Hacktoberfest Contributions so let's make sure to attribute properly as well. 

This PR marks a feature freeze. Do not merge any PRs in to develop until after this release has been merged into `master`. The current plan is to merge by the end of next week.

PR #1248, opened 2018-10-22
Support node, way, relation, viewport short codes

nrotstan:
* Add support in task comments for short-codes that generate links to
nodes, ways, relations, and OSM viewports

* Open linked entities from short codes in JOSM if user has chosen it
for their active editor, or in a new Overpass Turbo or OpenStreetMap tab
if not

* Offer user option to preview comment being authored, including markdown and
short-code links

* Display modal error if user has chosen JOSM as their editor but communication
with JOSM fails, and offer user option to stop using JOSM as their chosen
editor

* Supported short codes:
  - `[n/123]` or `[node/123]` for nodes
  - `[w/123]` or `[way/123]` for ways
  - `[r/123]` or `[rel/123]` or `[relation/123]` for relations
  - `[v/14/42.3824/12.2633]` or
    `[view/14/42.3824/12.2633]` or
    `[viewport/14/42.3824/12.2633]` for OSM viewports

> Note: for convenience, the full openstreetmap.org URL of simple OSM viewports
may be used in place of zoom/lat/lon in viewport short codes, e.g.
`[v/https://www.openstreetmap.org/#map=14/42.3824/12.2633]`

* Node, way, and relation short-codes may be combined into a combo code,
such as `[n/123, w/456]`, that generates a single link that loads all
entities together. Individual codes in a combo may be comma or space
separated. Viewport short-codes are not supported in combos

* Fix some minor eslint issues, such as missing semicolons

### DIFF
--- a/client/app/components/project-chat/project-chat.directive.js
+++ b/client/app/components/project-chat/project-chat.directive.js
@@ -90,7 +90,7 @@
             resultsPromise.then(function (data) {
                 vm.messages = data.chat;
                 for (var i = 0; i < vm.messages.length; i++) {
-                    vm.messages[i].message = messageService.formatUserNamesToLink(vm.messages[i].message);
+                    vm.messages[i].message = messageService.formatShortCodes(vm.messages[i].message);
                 }
                 // set the location.hash to the id of the element to scroll to
                 $timeout(function () {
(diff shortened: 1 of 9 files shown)


ethan-nelson:
This is getting a bit messy because of its age--sorry! I think I resolved the last set of conflicts though.

ethan-nelson:
As mentioned in #1245, we lost comment box changes with the rearrangement of the buttons, so the changes made to the separate `comment-box.html` file were not actually included. https://github.com/hotosm/tasking-manager/pull/1248/commits/23211b078061f00068f5ac6f78a8974d4a498944 adds those back in so that the preview button still works.

(We are still missing the ability to comment without task check out as that was also overwritten. Adding that is still in the back of my mind along with this, but things are a bit too messy at the current moment. The code is still there in #1201 when the PR queue does slim up.)

Issue #860, opened 2017-10-23
page title remains the same


pgiraud:
Currently, the title of the page doesn't change during navigation from one step to another. It prevents users from going back to a specific step in the navigation history.
![screenshot from 2017-10-23 08-28-32](https://user-images.githubusercontent.com/319774/31875027-8c3e4f0a-b7cc-11e7-80bc-b106166caaa0.png)

verdy-p:
This priority is raised because it evidently causes difficulty when navigating. It's not difficult to implement in code: each generated page should have an updated title containing the site name "HOT Tasking Manager" and the specific topic

Preferably in the order "(Topic) - HOT Tasking Manager" rather than the reverse, so tabs in browsers will still display a more specific part even if it is shown truncated ("HOT Tasking Manager" is not needed at begining, we already have a site icon at start, and browsers are already grouping pages for the same site when there are multiple sites in the user's history

pgiraud:
> It's not difficult to implement in code

How do you know?

verdy-p:
One line of code only to set a property...

Le 30 nov. 2017 12:46 PM, "Pierre GIRAUD" <notifications@github.com> a
écrit :

pgiraud:
> One line of code only to set a property...

I'd be interested to see what change you would propose. Please provide a patch.

ethan-nelson:
Sure @DorukUlucay , I will mark this as in progress!

DorukUlucay:
I've sent a pr. Please do review.

ethan-nelson:
This has been fixed in #1252

PR #1235, opened 2018-10-14
Add {{ | translate }}

brownben:
Add {{ 'x' | translate }} to UI strings which did not have it.
For Issue #1234

PR #1252, opened 2018-10-24
title changes while browsing #860

DorukUlucay:


willemarcel:
Hi @DorukUlucay 

This sounds very good! Thank you very much! Would you mind replacing the strings you added with `{{'term'| translate }}`? That way people can translate it. See this for a reference: https://github.com/hotosm/tasking-manager/pull/1235/files

DorukUlucay:
Hello @willemarcel 

All strings I added is on a js file. Not in HTML.

So I guess what I must do is call the method behind translate on that strings. Is that so ?

DorukUlucay:
Hello @willemarcel 
Did it. Tested it. Should be ok now.

Issue #796, opened 2017-09-21
Message users when reaching new levels or number of changesets; encourage validation


aawiseman:
I think it would be good to contact users when they reach certain mapping levels like intermediate, expert etc. and also contact them when they reach a certain number of changesets or levels to suggest they try validating. That might encourage more people to validate.

aawiseman:
Hey, just wanted to check on this @bgirardot @ethan-nelson . Thanks. Think it would be a good way to get people to validate and get more involved.

ethan-nelson:
Hey Andrew, I may have someone working on this as an intro to TM dev. Otherwise, since it has been sitting on my to-do list for a while, I will knock it out.

ramyaragupathy:
@ethan-nelson @bgirardot - I'm going to give this a try this week as a starter task

ramyaragupathy:
This is done through #1483

PR #1253, opened 2018-10-24
 Notify users on level upgrade

ramyaragupathy:
Per conversation in https://github.com/hotosm/tasking-manager/issues/796, 
- [x] Added a messaging functionality to `check_and_update_mapper_level` function. I
- [x] Included a template for the level upgrade

Snapshot of the message
image

### To do:
- Resolve circle errors
- Test for 
  - [x] beginner->intemediate
  - [x] intermediate -> advanced
  - [x] No level change situations
- Get the message template reviewed - @ethan-nelson Could you please verify the text here?


ramyaragupathy:
I'm looking at the unit test cases to understand the test failures.

ramyaragupathy:
I've been debugging the unit test cases locally and able to reproduce the same errors on:
- my branch 
- the main `develop` branch and
- other development branches from other open PRs.

While these unit test case failures are reproducible in all these branches, I'm not sure if these are the sole reason for the circle CI failure. Because other successful CI tests also have similar unit tests errors. @dakotabenjamin @ethan-nelson any pointers here?

cc @smit1678

ramyaragupathy:
Error log points to empty username for the test case id. However, I can see a username listed for the id, while debugging. Not sure why this is converted to `None` type in CI env.

ramyaragupathy:
Thanks for the help @ethan-nelson! Like discussed following steps helped.

- Changing the user id
- Importing Message module
- Adding Message in the function input

ethan-nelson:
Thanks for your work on this @ramyaragupathy --especially with the testing issues.

Looking a bit deeper, it seems like this may send an email to the user every time the server does this mapper check (either when they check out a task or when the user DTO is requested). I'm thinking maybe a flag in the users table could keep track of whether a given notification (intermediate, advanced) has been sent?

Besides that, I think things look good to go.

ramyaragupathy:
Sorry about missing this @ethan-nelson adn thanks for flagging! 

I am thinking of the following columns in the user table:
1. Starting Level (String) - Not possibel for the already existing users, but may be we could extract the data based on the joining date?
2. Level Upgrades allowed (Integer)-  Depending on the user level, maximum 2 upgrades allowed (Beginner -> Intermediate, Intermediate -> Advanced)
2. Level upgrade notified (Integer) - Depending on beginner level can be done twice maximum

xamanu:
@ramyaragupathy, this looks great and only a small thing missing. Would be great to get this in. Especially now that we are more looking at the notifications user flow and making sure messages reach the people.

ramyaragupathy:
Worked w/ @willemarcel to implement a solution without introducing any new db changes. However could not get to run unit tests locally - @willemarcel is looking at local setup for unit tests.

Other than this PR is ready to be reviewed @JorgeMartinezG

ramyaragupathy:
Closing this here in favor of https://github.com/hotosm/tasking-manager/pull/1483. This PR has become quite heavy on files touched with a history squash

Issue #1256, opened 2018-10-26
Mapping action icon colors and hover state


smit1678:
The Stop Mapping icon only shows in hover state. It looks like the icon color is the same as the button and so it doesn't show up until you hover. These are from the latest layout changes deployed on tasks-stage.hotosm.org. 

Thinking that the best way forward here is to match the text color. White when not hovering over, and then change to brown or orange when in hover state. But open to other ideas. 

![tasks-icons](https://user-images.githubusercontent.com/796838/47555916-fa770200-d904-11e8-910d-1f83cc72a9eb.gif)

dakotabenjamin:
This should be a simple fix:
https://github.com/hotosm/tasking-manager/blob/develop/client/assets/icons/stop-map.svg?short_path=49a0b68#L9
Change the color to `#000000`

smit1678:
@dakotabenjamin My thinking was that instead of black, it would match the text color and how the text color changes based on state. Black clashes and sticks out quite a bit.

dakotabenjamin:
Merged with #1260 and #1262

PR #1260, opened 2018-10-26
Interactive SVG icons for button action text

ramyaragupathy:
From #1256 - Using SVG as a background img, icon isn't part of the DOM and doesn't let any property manipulation. So adding svg content to html file directly with a comment on the source.

icon


dakotabenjamin:
@ramyaragupathy did you forget to add the IDs to the html elements? I realized there's a mistake after I merged.

Issue #1218, opened 2018-10-09
Review unreviewed translations on Transifex


smit1678:
Currently on Transifex we have 25 languages: https://www.transifex.com/hotosm/tasking-manager-3/dashboard/. Support for reviewing existing translations is need. Below is a list of what's been completed and what's left for review. 

#### Languages that are fully translated but need review of some or all strings: 

* [ ] Chinese (Taiwan) (zh_TW)
* [ ] Czech (cs)
* [ ] Dutch (Netherlands) (nl_NL)
* [ ] French (fr) 
* [ ] German (de)
* [ ] Japanese (ja)
* [ ] Polish (pl)
* [ ] Portuguese (pt)
* [ ] Spanish (es)
* [ ] Ukrainian (uk)

#### Languages that are not fully translated but have strings that need review: 

* [ ] Finnish (fi)
* [ ] Norwegian Bokmål (nb) 
* [ ] Slovenian (sl)
* [ ] Malagasy (mg)
* [ ] Vietnamese (vi) 
* [ ] Portuguese (Brazil) (pt_BR) 
* [ ] Hungarian (hu) 
* [ ] Galician (gl)
* [ ] Russian (ru)
* [ ] Persian (Iran) (fa_IR)

ethan-nelson:
Please ping on this PR with your username if you need access to Transfix and we will make sure permissions are added!

agnesoshiro:
Hi! Happy to help review Japanese strings. My Transifex account username is agnesoshiro. Thank you!

ethan-nelson:
@agnesoshiro added you--thanks!

AndreaGriffiths11:
Hi! Happy to help with Spanish. My Transifex account username is AndreaGriffiths11

Woestijnprinses:
Can I review Dutch  please. My Transifex account username is Woestijnprinses

ethan-nelson:
For those interested in translating, we've updated Transifex with strings that were previously missed, so there is plenty to translate in every language!

hi15:
Hi! Can I review portuguese (br). My Transifex account username is hi15

Leodau:
Hey, i saw you guys still have tons of spanish strings to translate, I would love to get those done!

My Transifex username is **leodau**
:)

xamanu:
Related to #1678

xamanu:
I will close here, in favour of a translation call together with the next round for testing.

PR #1263, opened 2018-10-29
Translate strings to portuguese (Brazil)

hi15:
Translate strings to portuguese (Brazil).

Related to issue #1218


ethan-nelson:
Hi @hi15 we have a decent chunk of these strings already translated on our Transifex site (https://www.transifex.com/hotosm/tasking-manager-3/dashboard/). I am going to go through and manually add any from your file that are untranslated on Transifex, but I would encourage you to create an account on there to help with future translations as that is the mechanism we use for this process. Thank you for your translation help and interest!

Issue #1266, opened 2018-11-03
Error displaying and downloading task boundary GPX


dakotabenjamin:
Multiple users are reporting an error when mapping. The task boundary is not displaying. Attempting to download the boundary as a gpx shows this error:
```
{"Error": "Task as GPX API - unhandled error: object of type 'BaseQuery' has no len()"}
```

This is a critical bug and should be fixed ASAP. In the meantime, we should redeploy an older version of TM3 to tasks.hotosm.org.

PR #1267, opened 2018-11-03
Bugfix for gpx generation error

ethan-nelson:
Re #1266 switching back getting the size of a task query using the object function count instead of len. This is an undo to changes at https://github.com/hotosm/tasking-manager/pull/1211/files#diff-4864ec93a020afee0531607d8a0e7fa4


ethan-nelson:
@ramyaragupathy have you tried testing this again? i've tried downloading a gpx outline on my local install using the link right below the editor drop-down and it works fine, but testing on live iD won't work since the link passed through references the local install at localhost and tries to access it on the iD server.

PR #1271, opened 2018-11-13
Add expert augmented diff and attic query controls

nrotstan:
* Add new control for users in expert mode to each history entry in task
history table for viewing an augmented diff between two entries or attic
query data for a single entry

* Selecting two separate entries will show an augmented diff of the
changes between those entries in Achavi in a new browser tab

* Selecting the same entry twice will perform an attic query and open
the data in either JOSM (if the user has selected it as their editor) or
Overpass Turbo in a new browser tab

* Add new `atticQueryOffsetMinutes` config setting for adjusting history
entry timestamps for the purpose of querying attic data to ensure that
data has had time to become available (defaults to 10 minutes)

* Reorganize some of the projectController bbox code into separate
methods to facilitate easier reuse
attic

### DIFF
--- a/client/app/project/project.controller.js
+++ b/client/app/project/project.controller.js
@@ -79,6 +79,9 @@
         vm.showLicenseModal = false;
         vm.lockingReason = '';
 
+        // Augmented diff or attic query selection
+        vm.selectedItem = null;
+
         //interval timer promise for autorefresh
         var autoRefresh = undefined;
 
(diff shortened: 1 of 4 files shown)


ethan-nelson:
Thanks @nrotstan ! 

@russdeffner do we have an advanced validator training curriculum--or is there a better place we could start collecting information on expert mode features? I think a wiki page in the TM GitHub repo is warranted to for non-HOT users.

PR #1272, opened 2018-11-15
Let users see their invalidated tasks on profile

nrotstan:
> Notes:
> 1. This commit adds a new migration (server) and a new NPM package (client)
> 2. This work only applies to tasks invalidated going forward

* Let users view both tasks they mapped that were marked invalid, as
well as tasks they invalidated themselves, from their profile page

* Invalidated tasks are considered "open" if they have not subsequently
been marked valid and may still require attention, and "closed" if
they been subsequently validated

* If user clicks through to an invalidated task from the new tables,
highlight corresponding invalidation entry in task history on the task
page for easy scanning

* Add task_invalidation_history table for tracking invalidated task
history going forward

* Add `user/:username/invalidated-tasks` API endpoint to server

* Maintain invalidation history on server to track tasks marked invalid
and whether they are later marked valid


ethan-nelson:
Hey Neil, just a quick update on the mostly silence from my end. We're working on redeploying the last release, so there will be a little delay until then before I can get these reviewed.

ethan-nelson:
But thank you!

ethan-nelson:
This is great work Neil--thanks! Beyond the major changes, the enhancement of highlighting a certain history element through the URL parameter is going to be useful for other things as well.

I did make three minor changes on your branch: 1) updating the database downgrade ID since there has been a migration since this was first filed, 2) fixing a conflict in the index file for importing modules due to the same reason, and 3) sanitizing sort arguments that are passed to the service function through the API. This forces `asc` or `desc` direction and one of the two columns included without throwing an error.

nrotstan:
Awesome, Ethan, thank you!

PR #1274, opened 2018-11-15
Make task auto-unlock duration configurable

nrotstan:
* Add `TASK_AUTOUNLOCK_AFTER` configuration setting to server that
allows task auto-unlock delay to be configured (defaults to 2 hours,
preserving prior behavior)

* Include configured delay in task data sent to front-end

* Update front-end task timer to make use of delay duration provided by
server instead of using hard-coded 2 hours

### DIFF
--- a/client/app/project/project.controller.js
+++ b/client/app/project/project.controller.js
@@ -381,7 +381,7 @@
             }
             if (task != null && task.taskId in vm.lockTime) {
                 var lockTime = moment.utc(vm.lockTime[task.taskId]);
-                return lockTime.add(2, 'hours').diff(moment.utc(), 'minutes');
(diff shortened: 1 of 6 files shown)


ramyaragupathy:
Tested for the following configurations.
- [x] 2h30m
- [x] 30m
- [x] 15m
- [x] 7d

In all cases time left is displayed in minutes like 👇 
image

Just thinking if we could get to display units in hours & days if the auto unlock duration exceeds an hour. Like:

- 2hrs and 10 minutes left instead of 130 mins left
- 1 day and 15 hrs left instead of 2340 minutes left

@nrotstan @ethan-nelson - what do you think here?

ethan-nelson:
I think that's a good point, but I'd throw it into a new ticket--I don't see it as an immediate blocker for this.

ethan-nelson:
Also I really like including this in the API data sent to the frontend :+1:

PR #1276, opened 2018-11-16
Allow expert validators to classify mapping issues

nrotstan:
Allow expert validators to classify mapping issues

> Note: this includes a new database migration

* Show validators in expert mode an additional form allowing them to
(optionally) classify the type and number of problems they discover with
mapping work that is leading to invalidation of the task

* Display noted mapping issues in the task history immediately below the
invalidation where they were recorded

* Allow mapping issues fixed by the validator on behalf of the mapper
to be noted when marking a task as validated

* Display in the task history when a task was validated with fixes, and
show the identified list of issues addressed by the validator

* Record identified mapping issues in a `task_mapping_issues` database
table

* Add a new page, accessible from the account-nav dropdown menu, on
which project managers and admins can manage the available mapping-issue
categories

* Seed the mapping-issue categories with a couple of initial categories

* Add server RESTful API endpoints for managing mapping-issue categories

* Add new `mapping_issue_categories` and `task_mapping_issues` database
tables in new migration

* Add support on server for optional inclusion of noted mapping issues
during task validation/invalidation


nrotstan:
Rebased with latest code from develop.

russdeffner:
Hi Neil, sorry, I don't have a way to test this locally at the moment; but how is the list 'managed'? Is it similar to how campaign and organization metadata works now? i.e. if something isn't in the list, the validator can add it? Or something only Admins and Project Managers can do from the new page? My curiosity here is, would it be better to allow Validators to basically enter anything with the risk of the list 'bloating' or risk not having something in the list until they can get PM to add it (and maybe choosing to mark something else instead)?

nrotstan:
Hi Russ, currently new issue categories would need to be added by a PM. I can definitely see the argument both ways. My thought was that it might make sense to have some concensus on the categories, especially in the beginning, and that we could always loosen things up later if desired.

russdeffner:
Ok, thanks for the info Niel; I guess we should have a bit of a think on how this will work in practice. As on optional thing, I think it would be fine to bring in. Thinking we will need very generic `mapping_issue_categories`, such as "Missing Feature(s)", "Feature Geometry", etc. so maybe only a handful; with a bit more variable `task_mapping_issues` - so maybe it's just an estimate of how many missing features (as example) or something like '1/2 buildings not square' or 'roads need more nodes on corners', etc. I could see validators needing to sometimes be very generic, like 'finding lots of misshapen buildings, task needs another go' - although of course this could still be done with the general comments, but just thinking out-loud of situations validators would want to categorize something that isn't easily given a quantitative measure of 'how much work' needs to be done.

xamanu:
Excuse me @nrotstan for the delay here. Let's bring this in. May I ask you to rebase once again, so we can review with the latest `develop`? Thank you!

nrotstan:
@xamanu Freshly rebased and included some additional minor enhancements, such as typeahead support when recording issues; the ability to note issues when marking a cell valid that the validator fixed on behalf of the mapper ("validated with fixes"); and support for archiving used issue categories that should no longer be offered.

PR #1288, opened 2018-11-28
Add matomo tracking code

dakotabenjamin:
Adds tracking js for the new tracking site. We are not removing the piwik tracking code yet.


dakotabenjamin:
This needs work still, please hold off merging

dakotabenjamin:
We are ready for testing- some notes:
- the tracking code control can be found at https://github.com/hotosm/matomo-tracking/
- the privacy policy link goes to hotosm.org/privacy which currently doesn't exist (404)
- I need to set up some continuous integration on the matomo tracking repo to upload updates to s3 without breaking current builds (hence the versioned file).

ethan-nelson:
@dakotabenjamin let me know if I can help with testing in some way. Otherwise we can just throw this new code in and test it out on develop?

dakotabenjamin:
@ethan-nelson please check the client that the popup shows up on the bottom of the page. Make sure you have cookies enabled and script blockers disabled. 

The code update I just pushed incorporates [hotjar](https://hotjar.com) tracking so we can look at pointer movements and clicks

smit1678:
@dakotabenjamin @ethan-nelson I haven't tested this but one thing realizing here is that on smaller screens and mobile is that we have the Mapswipe banner. So we'll have a banner on top of a banner. Perhaps we need to have a class on the Matomo banner to put it at the top? or are we ok with the interaction of first hitting the Matomo banner and then dealing with the mobile banner. Thoughts?

dakotabenjamin:
Can we do that with the z-index? The matomo banner is at index 1001 and the mobile banner is index 4.

Issue #1279, opened 2018-11-21
Task oriented project creation


smit1678:
Proposal based on work presented in TM Working Group: 

Some projects can have targeted purposes based on GIS processing, ML Extractions, or other sources of data. Utilizing that as an input to project creation can insure mapper time is maximized. This also allows project creators to bring external data into the project and individual tasks (similar to the import process features we've used in the past via per task instructions). 

Feature improvement focuses on:

* Clip project to where input task geometry exists
* Maintain task data for passing on to the mapper into either ID (Future) or JOSM (currently just via JOSM plugin)

PRs to test this idea coming soon via @lesserj

willemarcel:
I think this issue is solved with arbitrary tasks & clip to aoi.

PR #1302, opened 2018-12-18
Task datasets

lesserj:
This implements task datasets in Tasking Manager.  Users can create a project by uploading a geojson file with geometries.  Those geometries are split on task boundaries and stored with the task.  There is an included JOSM plugin that can be installed.  When a user clicks start mapping on that task, the task data will be passed to JOSM as a new layer.  The user can then use the task datasets as guides while mapping.  This is useful when geometries are available from another source (such as roads or buildings derived by machine learning) that we don't want to do a bulk import from, but we do want to include as a reference layer while mapping.

When generating the project, any tasks without task data are removed.  When a task is split, the task data is also split and associated with the new tasks.

This addresses #1279.

c.c. @TimuSumisu


ethan-nelson:
Fixing the merge conflict locally didn't work like I wanted it to....I'm going to push the changes to a fresh branch and open up a new PR. Sorry for that hassle @lesserj

ethan-nelson:
Please see #1335

Issue #1280, opened 2018-11-21
Prioritization Framework


smit1678:
Proposal based on work presented in TM Working Group:

Allow project creators to have fine grained control over the prioritization of their projects. This looks to go beyond adding a visual priority area and create the ability to rank and order task grid squares. This also adds a `Select next highest task` button in addition to the `Select random task button`. 

* For any project, a project manager can upload a number of geojson polygons.
* These can all be weighted independently, and reset/recalculated on the fly.
* Any grid that intersects the priority polygons will receive n additive score from what it intersected.
* Mappers can sort tasks by priority, or ask for a random, higher priority grid.

PRs to test this idea coming soon via @lesserj

PR #1303, opened 2018-12-18
Priority datasets

lesserj:
This implements task priorities. It allows project managers to upload geojson polygons to a project.  Each uploaded dataset can have a weight assigned to it.  Datasets are intersected with tasks in the project.  When a priority dataset, or multiple priority datasets intersect with a task, the weights are summed and assigned to that task.

When mappers look at a project that has priorities assigned, there is a table that is sorted from high priority tasks to low priority tasks.  This allows mappers to quickly find and map the highest priority unmapped task. 

Also implements a select random task button that selects the highest available priority task.

This addresses #1280.

c.c. @TimuSumisu


xamanu:
Rerolled in #1485

PR #1308, opened 2018-12-20
Line up master and develop

dakotabenjamin:
@smit1678 @ethan-nelson is there any way we can avoid this?


ethan-nelson:
As far as I know, the only way is to push to master from the command line instead of doing a PR.

ethan-nelson:
Or maybe `rebase and merge`?

Issue #987, opened 2017-12-09
Add restriction option for validation level in project metadata


bgirardot:
It would be nice if project managers could restrict validation to mappers who have intermediate and/or advanced level.

This would allow project managers to prevent very new mappers from validating, but not just restrict it to validators with the actual validation role.

I guess I would also include people with the actual validation role as well, even if they are not intermediate or advanced mappers.

(thank you for the suggestion @russdeffner )

johnwhelan:
JOSM has a number of features and tools that help things that can detect duplicate buildings and the todo list.

Project Managers do not always know much about mapping, I've come across one who invalidated a tile because something had been mapped that wasn't on the project list.  Fine except it had been mapped a year before the project.

We don't have enough experienced validators so even an Id validation is better than nothing sometimes.  Having said that the PM's need a list of things that are not checked when validating with Id so they can get a better understanding of the issues before saying "Just do it" in Id.

ramyaragupathy:
I'm going to give this a try 🔎

russdeffner:
Thank you for working on this @ramyaragupathy, I was just coming to make this suggestion and happy to find that Blake did a while ago and you're working on it!

xamanu:
This has been addressed by @ramyaragupathy in #1313

PR #1313, opened 2019-01-01
Validation metadata update

ramyaragupathy:
From #987 - Allowing option for intermediate or advanced mappers to validate tasks. 

Total validators: 405
Number of beginner mapper validators = 122
Number of intermediate mapper validators = 40
Number of advanced mapper validators = 243

Apart from the 405 validators, this option will enable all intermediate/advanced mappers without validator role to evaluate task data.

Modified the front end to include a check box under `Permissions`

image

Working on the backend changes.

cc @ethan-nelson @russdeffner @bgirardot


ramyaragupathy:
- Added an additional option under permissions to let project managers choose intermediate/advanced mappers to validate tasks. This option can be chosen only if validator level is enforced. If validator role is not enforced, automatically anyone can review a task. So just clubbing it with the validator role required checkbox

validator

- Updated the frontend code to capture the new conditions
- There's DB upgrade to the projects table. A new column `allow_non_beginners` tracks intermediate/advanced mappers role in validation
- Included the different scenarios under karma tests

russdeffner:
@ramyaragupathy - I think we want it to work the other way? Or be independent of each other?

Use case 1: I don't want to have to give everyone the validator role, but want anyone X level or higher to be able.

ramyaragupathy:
@russdeffner - I have listed down the possible scenarios and the validation permission for validators and different mapping levels. Please confirm if my understanding is right here.

Alps please note we have 143 mappers who are beginners but still have the validator role.

EnforceValidatorRole|AllowNonBeginners <br> (Intermediate + Advanced)|Validator|Beginner|Intermediate|Advanced
--------------------|----------------------------------------------|----------|--------|---------|---------
✔️|✔️|✔️|🔴|✔️|✔️
✔️ | 🔴 |✔️|🔴 |🔴 |🔴
🔴 |✔️|✔️|🔴|✔️|✔️
🔴|🔴|✔️|✔️|✔️|✔️

ramyaragupathy:
Per chat w/ @russdeffner - modifying the conditions to allow only validators with intermediate or advanced mapping levels to validate when `allowNonBeginners = true`

EnforceValidator|AllowNonBeginners <br> (Int + Adv)|Validator <br> (Int + Adv)|Validator <br>(Beg.)|Beg.|Int.|Adv.
--------------------|----------------------------------------------|----------|--------|---------|---------|-------
✔️|✔️|✔️|🔴|🔴|✔️|✔️
✔️ | 🔴|✔️|✔️|🔴|🔴|🔴
🔴 |✔️|✔️|🔴|🔴|✔️|✔️
🔴|🔴|✔️|✔️|✔️|✔️|✔️

xamanu:
Rebased on `develop`, adjusted locally the migration to fit the new endpoint, and squashed the commits ([see all here](https://github.com/hotosm/tasking-manager/tree/validation-restriction-rebase)). Testingt the behaviour currently.

I suggest to rename the tiles and descriptions of the project's permission page:

> **Mappers:**
> 
> * Only mappers with experience level INTERMEDIATE or higher can map: If checked, only users with the listed mapper experience level will be able to contribute to this project. If unchecked, anyone can contribute. Go to the Metadata panel to change the mapper experience level for this project.
> 
> **Validators:**
> 
> * Require users to be validators: If checked, only users with the Validator role will be able to validate tasks in this project. If unchecked, anyone can validate or invalidate tasks.
> * Require experience level INTERMEDIATE or ADVANCED: If checked, only users with the experience level of Intermediate and Advanced will be able to validate tasks in this project.

@russdeffner, @ramyaragupathy What do you think?

ramyaragupathy:
@xamanu - agree with your inputs on wording improvement. Modified the html:

image

zlavergne:
@xamanu @ramyaragupathy I believe the `allow_non_beginners` column needs to have `server_default='FALSE'` to allow for backwards compatibility. I was editing a project and was unable to save. The problem was that the project service was sending `allowNonBeginner=null`. I had a similar problem with making `restrict-editor-programs` backwards compatible.

PR #1318, opened 2019-01-10
Update translations, add four templates

ethan-nelson:
Updating translations as they are currently with transifex (`tx pull`)--no manual intervention.

Additionally, added four new languages that were in the Transifex repo but did not have files added yet. It doesn't look like they have any strings translated yet, but the option will show up on the language menu...hopefully advertising we could use some translator help.

@dakotabenjamin @ramyaragupathy @willemarcel or @smit1678 if you don't mind approving so we can get this in for next release. Translation PRs become out of date fairly quickly.

After this next release, I will upload a file file to Transifex with new strings.


ethan-nelson:
Will reopen with updates from new PRs

PR #1317, opened 2019-01-09
Improving docs

willemarcel:
@arunasank had trouble to run Tasking Manager locally because it was missing information about how her could give permissions to her user account to create tasks.

PR #1316, opened 2019-01-09
Fix the README

arunasank:
* Format and fix the README
* Add missing information about retrieving OAuth credentials from OpenStreetMap
* Group related information

cc/ @hotosm/tech

PR #1307, opened 2018-12-20
Add walkthrough, fix typos in migration script

ethan-nelson:
Based on #1183 there were some bugs in the migration script. Additionally, the general process was kind of vague. This fixes the bugs and also adds more text to help a user go through the migration process. This PR is filed in line with changes to the wiki page https://github.com/hotosm/tasking-manager/wiki/Migrating-from-TM2-to-TM3.

Thank you to @wildintellect for @kleper for feedback on this process!

Issue #969, opened 2017-11-23
Allow multiple project managers / organizations


mnoblecourt:
For project managers working as a team, it would be useful that the original project manager (who created the project) can add additional PMs on a project, so that it also appears on their dashboard.
Same request for organizations although it might be more complex.

verdy-p:
for organizations, they should also have their own profile, with one or a few  leaders (users) (allowed to managed this organization profile and the list of members), a list of additional coordinators, a list of sponsored projects (only leaders or coordinators should be allowed to add to/remove from this list), ans probably their own assesment of priorities (instead of the default global priority of projects) for sponsored projects shown on their profile.
Ideally, the TM projects themselves should also list the sponsoring organizations (for now there's a single one: the organization tag when editing projects: if the project manager adds an organization to his project, this should not be automatically added to the list of projects sponsored/followed by the selected organization, instead a notification will be visible in the organization's profile when editing it by the organization coordinators or leaders, saying that they are proposed to endorse it or not, and no one else than the organization's coordinators may remove a project from their list of sponsored/followed projects; and even if a project is "archived" or "drafted" in the global list of projects, it may be "published", and then editable by organization's coordinators/leaders, or workable for mapping/validating by contributors to the organization)
The organization profile would have properties similar to individual projects such as description (including contact means by mail, social networks, or websites), general instructions (including URL references to the OSM wiki), or reports of activities. And of course an organization would have an image (like normal users) generally used as an icon, plus possibly a large banner on at top of their profile.
An organization need be really registered as a legal organization, it may be any working group with some common interest (e.g. user groups of mappers in a region or cities organizing a single mapathon)

mnoblecourt:
Following a discussion with Missing Maps partners: it would actually be interesting to have an organization dashboard showing all the projects for the organization but also overall stats for all projects (% mapped/validated, number of ongoing projects, etc.)
American Red Cross (@dalekunce) has started working on a similar tool and there is interest from Univ Heidelberg to help on that.
Also, possible to expand it to campaigns if the architecture allows? https://github.com/hotosm/tasking-manager/issues/583

bgirardot:
Sure martin, pull requests are always welcome!

xamanu:
There is some progress on several ends, that I think fit into this requirement here and I want to update and verify our thoughts here.

## Multiple organizations

We are currently redesigning the Tasking Manager and one very strong design assumption is to have exactly one organization per project:

[...]

PR #1319, opened 2019-01-10
v3.1.1 release

ethan-nelson:
Prepping for another release of the TM now that the holiday season has ended. Some bugfixes, some enhancements:

* Organization and campaign tags will no longer appear in the search drop down menu if they do not have any projects assigned. This will hopefully remove some of the misspellings, tester, and wrong tags. (#1148)
* Updated analytics code to serve from CDN (#1288)
* Provide option for visitors to opt out of tracking (#1288)
* Add new shortcodes that can be used in comments and messages (e.g. `[n/123]` to link to [OSM node 123](https://www.openstreetmap.org/node/123)); check out #1248 for all the details
* Provide attic and augmented diff options in task history for expert users (#1271)
* Convert the time remaining text on task checkouts from minutes to days, hours, minutes (#1311)
* Removed the fundraising banner (#1314)
* Updated the README (#1310)

Potential additional PRs for merging prior to release (#1316, #1317 for README updates; #1318 for updated translation strings; #1307 for migration script). Note these don't affect the application functionality, so this is ready for testing.

Deployment around Jan 17th?

Shout out to all the contributors for this release!! :tada: @fiennyangeln @kant @nrotstan @dakotabenjamin @ramyaragupathy @ethan-nelson @smit1678 :100:


ethan-nelson:
@russdeffner we could use some testing on this. The full changelog is listed above, but here are the key elements to test on https://tasks-stage.hotosm.org :

1) Ensure the user has expert mode enabled in the profile on the staging server (after logging in, click the user menu in the top right and go to profile; then click the box on the right side of the page under expert mode).

2) On the search page (https://tasks-stage.hotosm.org/contribute), ensure that there are organizations and tags listed on the left side drop down filter menus but also make sure there are not many junk tags like typos or misspellings. We have hidden tags and organizations that are not currently assigned to a project. A tag I added then removed from a project is `Ethan was here`, meaning that should _not_ show up in organizations or tags.

3) Check out a task on some project, and add a few of these bracketed variations to the comment before marking it mapped or unlocking it:

```
[n/1341440928 ]      <- this should reference a node from OSM
@FTA              <- this should mention my profile in the comment
[view/14/42.3824/12.2633]    <- this should reference the osm map at zoom 14 at those coordinates
[n/1341440928 , w/119369552]   <- this should reference both a node and way in OSM
```

Ensure that you "preview" the comment by selecting the button immediately below the comment box before submitting. When you submit, it should show the same thing. Once submitted, try clicking on some of the links to ensure they open proper windows. Then, check out the task, select JOSM as your selected editor, boot JOSM up on your computer, and try clicking the links again. The node and way links should now load things in JOSM.

4) Find a project that has some activity on it. Click on a task that has had that activity. There should now be boxes next to every action. Try double clicking the box next to an activity that has marked a task mapped. An attic query should show what the user has added. Then try clicking boxes next to two different people in the history. This should show the augmented diff of differences between the two. Now, check out the task and select JOSM as your editor. Try the double click action again to check if the attic query loads data in JOSM.

russdeffner:
Sure @ethan-nelson I will send a note to Project Managers soon and will try to test it myself in the next day or two.

Vaomatua:
@smit1678 Good question, it definitely has been after the layout changes. In the past it was very rare for tiles to be marked Bad Imagery, however, in the last month or so there have been dozens. It appears that the only way to go to the Stop Mapping or Completely Mapped buttons is to scroll down, but the Bad Imagery is the one that is visible and it is very large. This is probably part of the overall graphics and user experience.
I would say that with the availability of 5 decent image sources (DG 2, Esri 2, & Bing 1) that Bad Imagery should be an option that users have to look for. In the last 2 or 3 months I have published about 30,000 tiles and less than 10 of those were ones that had issues from all 5 sources. I do have a volunteer now that has taken on looking for tiles marked as Bad Imagery in order to give feedback to the users, but it would be better if we didn't have to do that.
Thanks for the great tool, I look forward to further improvements.

russdeffner:
The Organization and Campaign lists do look better. Still seeing some that have # and not, i.e. duplicates - maybe we say 'no hashtags' in the lists? Also 2 orgs have their name spelled out, and one with name spelled out with acronym in parenthesis; CHAI is example of that plus one incorrect arrangement of spelled out words.

dakotabenjamin:
Can we document the changes from #1248 so users are aware that this is a feature? Any suggestions as to where to put this documentation would be great.

dakotabenjamin:
Are we able to make it so we can select multiple tag/list items in the dropdown? That could be a short term solution

ethan-nelson:
@Vaomatua I believe we have discussed this in a TWG meeting (early December?) but I don't believe we ever opened a ticket on it. I have done so now in #1329

@russdeffner thanks for the feedback on that. As it stands, the PR does not try to modify any of the tags. We could add text on the project creation screen instructing people to avoid certain terms or symbols. For now, the best solution would be to search by those tags that are not desired and correct the tags on those projects.

@dakotabenjamin I've been told (but not really investigated) that the architecture makes it difficult to add more than one tag/campaign. This is probably because it's a single id reference in a column. Also see #969 

Any idea for the best place to put the documentation? We could have a popup that shows syntax examples (I recall we had something like that on TM2). I would lean towards adding that in the next version, though, so it doesn't hold this up.

russdeffner:
@ethan-nelson - cool, that's not too bad to do; I opened up all the "Clinton Access Health Initiative" projects and switched to "Clinton Health Access Initiative (CHAI)" - but the former still shows up in the drop-down now. Does it takes some time, or do we have to do something to 're-run the no project swipe' again?

ethan-nelson:
Hi @russdeffner the database query checks for projects every time there is a search, so a tag should fall off the list in real time whenever all projects are removed. However, if some exist as draft or archived, the org will still show up in the list (for this case, https://tasks-stage.hotosm.org/contribute?difficulty=ALL&organisation=Clinton%20Access%20Health%20Initiative&statuses=DRAFT,ARCHIVED)

ethan-nelson:
Alright since it's lunch, I had a little free time and went through the archived projects (there were 10) and deleted the `Clinton Access Health Initiative` tag. That tag now no longer shows up in the search drop down.

russdeffner:
Sure @ethan-nelson - that's what I thought I did, because none showed up without turning on the draft/archives. Anyway, let's take this to slack as it's more about managing our metadata then the PR.

PR #1321, opened 2019-01-10
Create Select a task by the number to map or validate

FransS1945:


ethan-nelson:
Hi again @FransS1945 I have moved this suggestion to https://github.com/hotosm/tasking-manager/issues/1325 so we can track it there. Thanks again!

PR #1322, opened 2019-01-10
Change matomo site id to TM site

dakotabenjamin:
Let's get this one in for 3.1.1 please. We updated Matomo to account for individual subdomains (tasks., etc) so the site id changed here.

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -271,7 +271,7 @@
 
 <!-- Matomo -->
 <div id="optout-form"></div>
-<script type="application/javascript">var site_id="1"</script>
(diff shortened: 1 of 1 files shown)


dakotabenjamin:
Thanks @ramyaragupathy I updated the tracking script: https://github.com/hotosm/matomo-tracking/blob/master/tracking-v2.js

It will take a few hours for the edge cache to clear but you can test it locally by downloading the file. I will not the tracking was automatic only in the case of hotjar, (which I fixed) but not with Matomo, as evidenced by https://github.com/hotosm/matomo-tracking/blob/master/tracking-v2.js#L95

Issue #1147, opened 2018-07-24
Adding more stats


sunidhiraheja:
This issue is meant to track the development of the `/stats` endpoint.
Here's a list of all features (existing and planned):

Homepage
- [x] Mappers online
- [x] Tasks mapped
- [x] Tasks validated
- [x] Total mappers
- [x] Total number of users who validate
- [ ] Total area mapped and validated
- [ ] Average task completion time
- [x] Number of projects by organization
- [ ] Enable querying by date

User stats
- [x] Mapper level progress indicator
- [x] OSM details
- [x] Contributions to projects (tasks mapped and validated)
- [ ] Total area mapped and validated
- [x] Total time spent mapping

Project stats
- [x] Activity log
- [x] List of contributors with # of tasks mapped and validated
- [x] % mapped and validated
- [ ] Total area mapped and validated
- [ ] Average task completion time
- [ ] Estimate of when a mapping project will be completed based on past contributions [#1083](https://github.com/hotosm/tasking-manager/issues/1083)
- [ ] Graph for mapped and validated tasks [#931](https://github.com/hotosm/tasking-manager/issues/931)

Suggestions welcome!

dalekunce:
Love the list. For users I'd add total _tasks_ mapped and validated. I would also enable date queries for almost all of these that you have listed.

russdeffner:
This is great @sunidhiraheja - thank you!
Regarding User stats:
* What "OSM details" are you thinking? Total changesets, age of account, etc? 
* For 'contribution to projects' I would also include total number of projects contributed to (i.e. mapped and/or validated on X projects).
* And last, would include projects created; maybe only displayed if given PM or Admin role.

giblet22:
In addition to @dalekunce user stats, the number of times a user has been invalidated would be useful for mapper engagement. 

Another would be amount of comments a user has made. This would be lower priority I think because I'm not sure if the payoff is worth it. But shows if users are engaging with others.

bgirardot:
one additional thing that is only slightly related, but I will mention it in case it fits into the development somehow:

list of the individual tasks someone has marked done, and more importantly, marked validated :)

We will often find a new person doing validation, not very well, then we have no way to find all the others ones they might have validated. There is an issue for this already, but seeing the discussion about tasks validate and mapped, I thought I would mention it here.

govvin:
[...]

PR #1172, opened 2018-08-28
Add more homepage stats

sunidhiraheja:
This PR focuses on the list of stats proposed in #1147 for the homepage section.

> Homepage
> - [x] Mappers online
> - [x] Tasks mapped
> - [ ] Tasks validated
> - [x] Total mappers
> - [ ] Total number of users who validate
> - [ ] Total area mapped and validated
> - [ ] Average task completion time
> - [ ] Number of projects by organization
> - [ ] Enable querying by date

- Added tasks validated
- Added total number of users who validate
- Added number of projects by organization

PR #1330, opened 2019-01-22
Pending stats for home-page endpoint

ramyaragupathy:
From #1147 - developing on @sunidhiraheja's work to add more home page stats. Similar to organisation summary, added 
- [x] campaign summary
- [x] Total number of campaigns
- [x] Total number of organisations
- [x] Updated organisation stats logic to [count untagged elements](https://github.com/hotosm/tasking-manager/blob/40ff291ef439f6e04b17a1e7843c21326c40b7e1/server/services/stats_service.py#L219) - current logic increments count by 1 everytime an empty tag is detected. Instead summed up counts from such untagged elements. Followed the same logic for campaigns as well
- [x] Total projects
- [x] stats/home-page -> stats/summary
- [x] ~~Total area from all projects~~
- [x] Total mapped area
- [x] Total validated area

Pending:
- [ ] Average task completion time
- [ ] Enable querying by date

cc @smit1678 @ethan-nelson

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -49,3 +49,5 @@ Unidecode==1.0.23
 urllib3==1.23
 webencodings==0.5.1
 Werkzeug==0.12.2
+newrelic==3.2.0.91
+pyproject==1.3.1
--- a/requirements.txt
+++ b/requirements.txt
@@ -50,4 +50,4 @@ urllib3==1.23
 webencodings==0.5.1
 Werkzeug==0.12.2
 newrelic==3.2.0.91
-pyproject==1.3.1
+pyproj==1.9.6
--- a/client/app/services/stats.service.js
+++ b/client/app/services/stats.service.js
@@ -96,7 +96,7 @@
             console.log('In service');
             return $http({
                 method: 'GET',
-                url: configService.tmAPI + '/stats/home-page'
+                url: configService.tmAPI + '/stats/summary'
             }).then(function successCallback(response) {
                 // this callback will be called asynchronously
                 // when the response is available
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -162,7 +162,7 @@ def init_flask_restful_routes(app):
     api.add_resource(StatsActivityAPI,              '/api/v1/stats/project/<int:project_id>/activity')
     api.add_resource(StatsProjectAPI,               '/api/v1/stats/project/<int:project_id>')
     api.add_resource(StatsUserAPI,                  '/api/v1/stats/user/<string:username>')
-    api.add_resource(HomePageStatsAPI,              '/api/v1/stats/home-page')
+    api.add_resource(HomePageStatsAPI,              '/api/v1/stats/summary')
     api.add_resource(CampaignsTagsAPI,              '/api/v1/tags/campaigns')
     api.add_resource(OrganisationTagsAPI,           '/api/v1/tags/organisations')
     api.add_resource(UserSearchAllAPI,              '/api/v1/user/search-all')
--- a/server/models/dtos/stats_dto.py
+++ b/server/models/dtos/stats_dto.py
@@ -85,6 +85,7 @@ class HomePageStatsDTO(Model):
     tasks_validated = IntType(serialized_name='tasksValidated')
     total_mappers = IntType(serialized_name='totalMappers')
     total_validators = IntType(serialized_name='totalValidators')
+    total_projects = IntType(serialized_name='totalProjects')
     total_area = FloatType(serialized_name='totalArea')
     total_organizations = IntType(serialized_name='totalOrganizations')
     total_campaigns = IntType(serialized_name='totalCampaigns')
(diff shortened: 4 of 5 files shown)


ramyaragupathy:
Current data changes captured in a gist here: https://gist.github.com/ramyaragupathy/67752e782ceabf351a5e59118dc7b49a

smit1678:
> home page stats

Just to clarify, these don't necessarily need to be visualized on the home page, but part of the `/home-page` endpoint: https://tasks.hotosm.org/api/v1/stats/home-page. Should we rename this to add clarity to the purpose of this endpoint?

ramyaragupathy:
Yeah I understand these changes are only for the API endpoint. Do you mean to rename the endpoint here? If so I'd go for `https://tasks.hotosm.org/api/v1/stats/summary-stats`

smit1678:
> Do you mean to rename the endpoint here?

Yes. What about `https://tasks.hotosm.org/api/v1/stats/summary` so it's a little less redundant?

ramyaragupathy:
Similar to the [issue encounteed](https://github.com/hotosm/tasking-manager/pull/1172#issuecomment-422647253) in avergae task completion time, there's trouble setting up the server when I try to compute area for mapped and validated tasks.  There's a 
` tasks = Task.query.filter().all()` to fetch all task data and then I loop through the list to aggregate area counts from mapped and validated tasks. 

Process fails with `SSL SYSCALL error: EOF detected` error. Was this the issue you encountered in average task completion time as well @ethan-nelson ?

ethan-nelson:
Hi @ramyaragupathy the requirements file should be sufficient for indicating new python modules. Regarding the EOF error, I would cancel the process before it would error out when I saw that the RAM was ever increasing. That could be the cause, though, if the RAM usage reaches the max. This may be something we need to index or calculate in a less precise manner (perhaps fraction of total area of AOIs based on fraction of tasks completed).

ethan-nelson:
It looks like CircleCI is trying to install https://pypi.org/project/pyproject/ from the requirements file

requirements.txt
```diff
@@ -43,3 +43,4 @@ urllib3==1.21.1
 webencodings==0.5.1
 Werkzeug==0.12.2
 newrelic==3.2.0.91
+pyproject==1.3.1
```
ethan-nelson: This should be `pyproj` (https://pypi.org/project/pyproj/)

ramyaragupathy: That just did the trick 💯

ramyaragupathy:
### Average task completion time
 Found an `action_text` column in `task_history` table. This seems to stor time spent on a task in `HH:MM:SS` format. However this field is not consistent for time storage when there is a `STATE_CHANGE` to `MAPPED` or `VALIDATED`.

**What we could do?**

- List out all activities for that task in the `task_history` table
- Make an average of all times except for STATE_CHANGE behaviours. But this won't be accurate since we miss out an exact duration for mapping & validation pass.

Recording the query used for computing average duration from task_history table:

```
WITH time_column AS (
SELECT
    CASE 
        WHEN action_text ~ '^0' and action_text LIKE '.*:*.' THEN CAST(action_text AS time) 
    END AS time_taken
    FROM public.task_history
)
SELECT AVG(time_taken) FROM time_column;
```
### Enable querying by date

Again a tricky situation here since we can't have all summary stat values queryable by date. We need to be cosnistent on timestamp storage for task level changes to incorporate this. @ethan-nelson @smit1678 - if you've ideas to go about here, that would be great


### Questions:

Switched to native queries for aggreagte project & task geometries. From our Db I could see we use `4326` projection for storage. `select st_srid(geometry) from public.tasks`

Going with turf geometry function projection(3597), I see our geometry values are closer to results from turf js library. However there are also cases where some of the coordinates just fail due to values failing to project on 3597. For now, I have used 4326 as the default projection. But for the results to match with turf we need to do a `ST_Transfrom` to 3597 befor area is computed.

@ethan-nelson @smit1678 - woudl you be able to throw some light here?

xamanu:
Just FYI: rebased this PR on `develop`.

xamanu:
Hi @ramyaragupathy, do you think we can review this PR already? And then tackle the two more complicated parts in a separate issue/PR? Or do you prefer to work on all at once?

ramyaragupathy:
The leftover items as I mentioned earlier - requires more work on the DB front. I would like to keep them separate from this PR. 

For the existing fix - I want to double check the projection values for geometry, since the area computed using [turfjs](http://turfjs.org) varies from the [postgis area](https://postgis.net/docs/ST_Area.html) function. Per chat, @ethan-nelson is running a check on projections.  

While that's being run you could still test the payload with the current state @xamanu

ramyaragupathy:
Per voice w/ @ethan-nelson @smit1678 - reverting total area for projects since we have area discrepancy between postgis & turf. This we suspect comes from old projects implemented without the projection check. Also this brings in the discussion of data audit.

Issue #1327, opened 2019-01-14
Draft a CloudFormation template for this stack


arunasank:
Currently the tasking manager is configured and set up to run on AWS using Elastic Beanstalk. We want to move the TM configuration to a CloudFormation template, to simplify the creation, updation and deletion of tasking manager stacks using configuration tools like https://github.com/mapbox/cfn-config. This will solve a couple of issues:

* It will make the process of creating or deploying the tasking manager to AWS infrastructure extremely simple, and reduce the dependency on Elastic beanstalk and the console.
* It will create more visibility around the various pieces present in the Tasking Manager infrastructure.
* Adding or removing pieces from the infrastructure in the future will be simpler.

Going to dig into translating the various pieces of the infrastructure as code.

cc/ @hotosm/tech

arunasank:
Hi folks! Quick state of things update:

I have managed to set up the tasking manager, and am currently able to access it using the EC2 public DNS and port, but I am unable to access it via the ELB DNS
* The ELB has access to the EC2 and health checks for the EC2 are passing on the ELB.
* The ELB has a listener forwarding traffic from the HTTP port 80 to port 8000 which is what we expose via gunicorn.
* I got rid of the nginx requirement since requests are proxied via CloudFront and ELB.

## How website looks when accessed via EC2 DNS:
<img width="1266" alt="screen shot 2019-02-06 at 10 24 42 pm" src="https://user-images.githubusercontent.com/3166852/52358748-9369fd00-2a5e-11e9-8241-8bacff9e2b91.png">

## How website looks when accessed via ELB DNS:
<img width="1280" alt="screen shot 2019-02-06 at 10 29 00 pm" src="https://user-images.githubusercontent.com/3166852/52358815-b0063500-2a5e-11e9-9cc5-ccee85874de0.png">

Seems like the fonts/logos/CSS and other resources are unavailable.

* Does anyone have thoughts on this?
* Can someone also shed light on why we run gunicorn on [(core * 2) + 1] workers? 

cc/ @hotosm/tech

PR #1332, opened 2019-01-23
Set up cloudformation for this stack

arunasank:
Fixes https://github.com/hotosm/tasking-manager/issues/1327 when done.

👋 @hotosm/tech! I have started setting up the infrastructure for the tasking manager on CloudFormation. Here's my picture of the architecture:

screen shot 2019-01-23 at 9 18 30 pm

Is this correct? Some things that are still unclear to me:
* How is the knowledge of the RDS postgres URL made available to the EC2? Is this configured through Elastic Beanstalk?
* Who is running the front end? Is each EC2 running both the front-end and the back-end?

## TODO
* Set up the RDS instance in the Cloudformation. In the final production version, we will accept a snapshot ID. For testing, do you all have recommendations on which RDS instance to use?
* Set up a load balancer, and figure out how the client connection flows through website URL to the ELB to the EC2s.
* Currently, we are running this stack on a `t2.large`, which has the following specificications:
  * 2 vCPU
  * 8 GB Memory
Is there a reason we picked this instance? What is most crucial for the tasking manager? Memory or CPU? Are there other instance types we considered and discarded for certain reasons?


dakotabenjamin:
1. The database is connected in software, via environment variables. See https://github.com/hotosm/tasking-manager/blob/master/server/config.py#L27. Then we set that in the EB in the Software configuration along with the other variables. 
2. I'm not sure about the second q, I think you're right though.

willemarcel:
@arunasank awesome work! Each instance serves its own frontend. Although I'm not sure if it's critical, I believe that separating the frontend server from the backend would reduce the load on the backend.

ethan-nelson:
As I understand it on local installs, each EC2 instance is serving the front end, which is really just a few html and js files. Those are copied to the Flask directories that are then served. I need to double check the EBS configuration to make sure this is true in the AWS deployment as well.

cloudformation/tasking-manager.template.js
```diff
@@ -0,0 +1,219 @@
...
+          'python3.6 -m venv ./venv &&',
+          '. ./venv/bin/activate &&',
+          'pip install --upgrade pip &&',
+          'pip install -r requirements.txt &&',
+          'echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf && sudo sysctl -p &&',
+          'sudo -u postgres psql -c "CREATE USER hottm WITH PASSWORD 'hottm';" &&',
+          'sudo -u postgres createdb -T template0 tasking-manager -E UTF8 -O hottm &&',
+          'sudo -u postgres psql -d tasking-manager -c "CREATE EXTENSION postgis;" &&',
+          'export TM_DB="postgresql://hottm:hottm@localhost/tasking-manager" &&',
+          'export TM_CONSUMER_KEY="VLm4AmuigZODSZQSEEdarv8LhFi4NFodc9JbvvEl" &&',
```
dakotabenjamin: @arunasank are these env variables supposed to be here? are they secrets?

arunasank: 👋 dakota, squashed and force pushed to fix this! 🌮

arunasank:
Hi folks! Quick state of things update:

I have managed to set up the tasking manager, and am currently able to access it using the EC2 public DNS and port, but I am unable to access it via the ELB DNS
* The ELB has access to the EC2 and health checks for the EC2 are passing on the ELB.
* The ELB has a listener forwarding traffic from the HTTP port 80 to port 8000 which is what we expose via gunicorn.
* I got rid of the nginx requirement since requests are proxied via CloudFront and ELB.

## How website looks when accessed via EC2 DNS:
screen shot 2019-02-06 at 10 24 42 pm

## How website looks when accessed via ELB DNS:
screen shot 2019-02-06 at 10 29 00 pm

Seems like the fonts/logos/CSS and other resources are unavailable.

* Does anyone have thoughts on this?
* Can someone also shed light on why we run gunicorn on [(core * 2) + 1] workers? 

cc/ @hotosm/tech

arunasank:
Also seeing this error:
```
[2019-02-06 16:38:09,049] CRITICAL in stats_api: Unhandled error: (psycopg2.ProgrammingError) relation "tasks" does not exist
LINE 3: FROM tasks
             ^
 [SQL: 'SELECT count(*) AS count_1 \nFROM (SELECT DISTINCT ON (tasks.locked_by) tasks.id AS tasks_id, tasks.project_id AS tasks_project_id, tasks.x AS tasks_x, tasks.y AS tasks_y, tasks.zoom AS tasks_zoom, tasks.extra_properties AS tasks_extra_properties, tasks.is_square AS tasks_is_square, tasks.geometry AS tasks_geometry, tasks.task_status AS tasks_task_status, tasks.locked_by AS tasks_locked_by, tasks.mapped_by AS tasks_mapped_by, tasks.validated_by AS tasks_validated_by \nFROM tasks \nWHERE tasks.locked_by IS NOT NULL) AS anon_1']
```

I am pointing my server at the Demo RDS, and assuming this is related to that?

dakotabenjamin:
Try pointing it at the staging server instead.

arunasank:
@dakota @ethan-nelson @smit1678 can someone with know-how into the deploy scripts for the tasking manager, take a look at my script: https://github.com/hotosm/tasking-manager/pull/1332/files#diff-7fc300911d0b9b98f085ac720b905092R126, and see if I am missing something vs what's currently being done on circle CI + EBS?

I took a look and can't really say what's different, but when I login after setting up using the staging configuration, it shows me the same error ^.

arunasank:
Hi folks, I have set up scaling policies for the stack to scale up on high CPU utilization. At this point, I am blocked by understanding how to run the tasking manager within each EC2, because of the aforementioned errors. 

I know you all are busy because of the stakeholders meeting, but would love to pair on this with one of you, so we can move ahead!

xamanu:
Love this! Just one question. Is the `cloudformation` directory a kind of standard or could we think about having an `install`, `install scripts`, `scripts`, or similar folder where we can place different scripts and templates for installation?

arunasank:
> Love this! Just one question. Is the `cloudformation` directory a kind of standard or could we think about having an `install`, `install scripts`, `scripts`, or similar folder where we can place different scripts and templates for installation?

It's not a standard for sure. Currently, we store all the scripts in the [UserData](https://github.com/hotosm/tasking-manager/pull/1332/files#diff-7fc300911d0b9b98f085ac720b905092R140) key. If we store the scripts in this repository, we would need to wget it/clone the repository before being able to run it, which is easy in the case of a public repository like this one, but harder for a private repo. 

Let me know if you have suggestions/alternatives!

xamanu:
I'm fine with keeping the script inside of the UserData for cloudformation. I was wondering whether we want to have a `install` or `depoy` directory, where we fit the `cloudformation` in, as well as other ways of installing and deploying the TM. Like for instance the `docker-compose.yml` of #796.

We could have a structure like this one:

* `deploy` / 
  - `ansible`(working on it)
  - `cloudformation` (the content of this pr)
  - `docker` (the proposed docker-compose.yml from #797)
  - `scripts` (for the centos, debian and ubuntu scripts we have in the wiki)

This would group all different ways of installing and deploying it on one place. What do you think?

xamanu:
I just saw the `devops` directory. Probably this is a good place to group all the installation things?

arunasank:
@xamanu Sure, we can standardise that. 👍 The only caveat is that, once the Cloudformation is done, we will not have a dependence on Docker, until we change our architecture to allow for that.

### Other updates

@dakotabenjamin, @ethan-nelson and I synced yesterday, and have a couple of solutions we want to implement to debug the issues above:

* Don't call `gulp run` in the deploy script, since it's a dev environment issue.
* Change the file permissions on all the files, to eliminate a permissions issue
* Change the base url in the `server/config.py` for the `TM_ENV` configuration
* Enable keep-alive settings on the EC2 instances and make sure that the keep-alive timeout is greater than the idle timeout settings of your load balancer.

arunasank:
Okay, I set up all of the above, and I still notice heavy delays in the rendering. I decided to curl the files that were not rendering locally on the instance, and measure the response time, and looks like the files take ages to render even on the local machine:

```
curl -w "@curl-output" -o /dev/null -s 0.0.0.0:8000/assets/img/hot-favicon.ico
    time_namelookup:  0.000
       time_connect:  0.000
    time_appconnect:  0.000
   time_pretransfer:  0.000
      time_redirect:  0.000
 time_starttransfer:  40.615
                    ----------
         time_total:  40.615
```

Compare this with a request to google.com:
```
curl -w "@curl-format.txt" -o /dev/null -s "https://www.google.com"
Warning: Failed to read curl-format.txtroot@ip-172-31-28-153:/home/ubuntu# curl -w "@curl-output" -o /dev/null -s "https://www.google.com"
    time_namelookup:  0.004
       time_connect:  0.006
    time_appconnect:  0.140
   time_pretransfer:  0.140
      time_redirect:  0.000
 time_starttransfer:  0.182
                    ----------
         time_total:  0.182
```
The `time_starttransfer`, also known as the time to first byte is slowww in our case. What's also interesting is that it's slow locally. I don't understand why this is. 

My tasking-manager repo is stored under the root directory: `/tasking-manager`, but permissions have been modified on the directory and all contained files and folder.

The CPU and Memory utilisations on the machine are normal, and since I am not hitting port 80, I am bypassing `nginx` (which redirect requests from port 80 to port 8000). This highlights a couple of things:
1. There's not an issue between the ELB and the server, so much as there is an issue inside the server to retrieve the files.
2. We can get rid of nginx as a proxy - I think the additional proxying is unnecessary, given we have CloudFront and the ELB, and will only cause further delays.

## Instance Type

Env | Name | Memory | CPU | Storage | Network
--| -- | -- | -- | -- | --
Dev | m3.large | 7.5 GiB | 2 vCPUs | 32 GiB SSD | Moderate
Prod | t2.large | 8.0 GiB | 2 vCPUs for a 7h 12m burst | EBS only | Low to Moderate

Note that they are fairly similar, barring the EBS/SSD storage. EBS is generally slower than an SSD, since it's external to the instance. I will try out an experiment with a t* instance and another AMI tomorrow, but I don't think this is it. Will try to explore this in an unbiased manner tomorrow.

## Questions

Have any of you seen something like this earlier? Where querying is super slow even locally?

arunasank:
More weirdness: Results of `ps aux | grep gunicorn`: I see two gunicorn processes running. When I killed one, the other died too. 🙈 

```
root     27084  0.0  0.2  66696 21612 ?        S    17:29   0:01 /tasking-manager/venv/bin/python3.6 /tasking-manager/venv/bin/gunicorn -b 0.0.0.0:8000 -w 1 --timeout 179 manage:application
root     27105  0.0  0.7 186376 58924 ?        S    17:29   0:01 /tasking-manager/venv/bin/python3.6 /tasking-manager/venv/bin/gunicorn -b 0.0.0.0:8000 -w 1 --timeout 179 manage:application
```

dakotabenjamin:
Is gunicorn set up to use asynchronous workers? I was reading [here](https://forums.aws.amazon.com/thread.jspa?messageID=419138) that was their issue. 

https://docs.gunicorn.org/en/stable/design.html

arunasank:
@ethan-nelson @dakotabenjamin I am a little stuck here, trying to understand why local queries to the files are taking as long as they are. Can one of you also try to set up the tasking manager on an EC2 the way you set it up usually. To verify that your set is perfect:

*  Save the following curl configuration to a file, `curl-output`:
```
time_namelookup:  %{time_namelookup}\n
       time_connect:  %{time_connect}\n
    time_appconnect:  %{time_appconnect}\n
   time_pretransfer:  %{time_pretransfer}\n
      time_redirect:  %{time_redirect}\n
 time_starttransfer:  %{time_starttransfer}\n
                    ----------\n
         time_total:  %{time_total}\n
```
* Run `curl -w "@curl-output" -o /dev/null -s 0.0.0.0:8000/assets/img/hot-favicon.ico` to check that the `time_starttransfer` looks good (Like about a couple of milliseconds or so)

I can take over with the ELB and instance level testing, once these localhost queries look good.

dakotabenjamin:
I followed the [script](https://github.com/hotosm/tasking-manager/wiki/Client-Install-Script-for-Ubuntu-16.04 on the same AMI as the one in this config, the only change being I ran the server using the `gunicorn` command. Here's the results:
```
ubuntu@ip-10-138-11-202:~$ curl -w "@curl-output" -o /dev/null -s 0.0.0.0:8000/assets/img/hot-favicon.ico
time_namelookup:  0.000
       time_connect:  0.001
    time_appconnect:  0.000
   time_pretransfer:  0.014
      time_redirect:  0.000
 time_starttransfer:  0.015
                    ----------
         time_total:  0.015
```

arunasank:
Recording my conversation with @dakotabenjamin yesterday. The above behaviour is not consistently reproducible. We also noticed cases where the first curl request works, but when a call is made via the EC2 DNS or ELB DNS, we notice curl delays as recorded further above.

We are investigating why local files are taking a long while to be retrieved by gunicorn.

I think the other curious behaviour is, why we don't see this issue on the Dockerised version of the current production instance of the Tasking Manager on AWS EBS.

arunasank:
Debugged the reasons for the localhost delay with my friend, @niteshjain, today who helped me understand what the issue was. 🙇‍♀️ Thanks again, Nitesh!

The reasons for localhost being overloaded are because of the way gunicorn works - there is a master process, and additional worker processes. The master process listens for requests and passes the details of a request to a worker process. At a minimum it's good to have at least 2 workers therefore. The localhost queries became super fast, consistently, once I increased the worker count to 2. 

We also have a load balancer health check that pings our gunicorn process at regular intervals of time(1/min). So, we should have at least 3 workers.

Further, it looks like the TM is a CPU intensive process, since the number of requests seem related to the number of gunicorn processes running at any given point of time. @dakotabenjamin's suggestion above to use async workers will definitely help. So, also will using high CPU instances. Going to test this out with some high CPU instances.

arunasank:
Switched to a `c3.2xlarge` instance, which is CPU optimised and has 8 CPU cores. Concurrent requests from multiple browsers are still slow. Going to experiment with asynchronous workers, threads, and pseudo threads on gunicorn tomorrow using https://medium.com/building-the-system/gunicorn-3-means-of-concurrency-efbb547674b7.

@dakotabenjamin, I am still seeing some issues with redirects after I auth with OSM. I am using the exact same credentials as what's on Staging on Elastic Beanstalk. Can you investigate this if you have some cycles today? You can use the CloudFormation template to launch a stack (along with the cfn-config commands we exchanged) or use a LaunchTemplate called `tasking-manager-staging-c3.2xlarge-aruna` that I created from our tasking manager instance to launch an instance alone from the EC2 console. Thanks for helping me debug the delays, and for the install script. 🙂

dakotabenjamin:
@arunasank the reason the login redirect isn't working is because you are using the Staging server environment, which assumes the redirect URL will be "tasks-stage.hotosm.org", and in our case, it's the ELB url. so that simply won't work without setting the url properly in the server.py file.

arunasank:
> @arunasank the reason the login redirect isn't working is because you are using the Staging server environment, which assumes the redirect URL will be "tasks-stage.hotosm.org", and in our case, it's the ELB url. so that simply won't work without setting the url properly in the server.py file.

gotcha. Can you update the stack to read this properly, and see if it works? We could export it as an env variable.

dakotabenjamin:
I'm not sure how to change the url in the file without knowing the url beforehand, but I'll see what I can do.

dakotabenjamin:
What's left to complete this PR?

xamanu:
Just a note: let's squash commits here a bit. Wdyt?

arunasank:
> Just a note: let's squash commits here a bit. Wdyt?

Will squash when merging - so there's a single merge commit. : )

cloudformation/tasking-manager.template.js
```diff
@@ -0,0 +1,201 @@
...
+          'echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf',
+          cf.join('', [cf.sub('export TM_DB="postgresql://${MasterUsername}:${MasterPassword}@'), cf.if('UseASnapshot', cf.getAtt('TaskingManagerRDS', 'Endpoint.Address'), cf.ref('RDSUrl')) , '/tm3"']),
+          cf.sub('export TM_CONSUMER_KEY=${OSMConsumerKey} && export TM_CONSUMER_SECRET=${OSMConsumerSecret} && export TM_ENV=${TaskingManagerEnv} && export TM_SECRET=${TaskingManagerSecret} && ./venv/bin/python3.6 manage.py db upgrade && cd client/ && npm install && gulp build && cd ../ && echo "done"'),
```
xamanu: I would suggest to use `python3` instead of `python3.6` so we can smoothly upgrade to non-major versions of python (like the current default one in Debian is 3.7).

cloudformation/tasking-manager.template.js
```diff
@@ -0,0 +1,201 @@
...
...update && sudo apt-get -y upgrade && sudo add-apt-repository ppa:jonathonf/python-3.6 -y && sudo apt-get update && sudo apt-get -y install python3.6 && sudo apt-get -y install python3.6-dev && sudo apt-get -y install python3.6-venv && sudo apt-get -y install curl && curl -sL https://deb.nodesource.com/setup_6.x > install-node6.sh && sudo chmod +x install-node6.sh && sudo ./install-node6.sh && sudo apt-get -y install nodejs && sudo npm install gulp -g && npm i browser-sync --save && sudo apt-get -y install postgresql-9.5 && sudo apt-get -y install libpq-dev && sudo apt-get -y install postgresql-server-dev-9.5 && sudo apt-get -y install libxml2 && sudo apt-get -y install libxml2-dev && sudo apt-get -y install libgeos-3.5.0 && sudo apt-get -y install libgeos-dev && sudo apt-get -y install libproj9 && sudo apt-get -y install libproj-dev && sudo apt-get -y install libgdal1-dev && sudo apt-get -y install libjson-c-dev && sudo apt-get -y install git && git clone --recursive https://github.com/hotosm/tasking-manager.git && cd tasking-manager/ && git reset --hard ${GitSha} && python3.6 -m venv ./venv && . ./venv/bin/activate && pip install --upgrade pip && pip install -r requirements.txt'),
```
xamanu: Is there a specific reason a special repository for the inclusion of Python 3.6 is added here. Wouldn't it be easier to run on the Python 3 version of Debian (which is 3.7 on the current stable version) and the TM runs well with it.

xamanu:
Just a comment: It would be good to rely on an ec2 OS that bases on the latest version of Ubuntu's Long Term Support branch (18.04.2).

dakotabenjamin:
@xamanu I believe the current restriction is finding a supported AMI that does not have EBS optimization.

xamanu:
Can we squash the commits in this PR? This would be nicer to handle for the `machine-learning-cfn` branch.

docker-compose.yml
```diff
@@ -1,45 +1,45 @@
-# To use this file, run `docker-compose up`.
-version: '3'
-
+version: '2'
```
xamanu: It looks like the work  in `develop` made these changes to the `docker-compose.yml` file obsolete.

Dockerfile
```diff
@@ -1,25 +1,23 @@
-FROM python:3-stretch
+FROM python:3.6
```
xamanu: Could we get rid completely of this file in favour of `devops/docker/tasking-manager/Dockerfile`?

arunasank: oh, dropping this!

xamanu: Perhaps we can also delete `requirements.aws.txt`, `.ebignore` and `.ebextensions` within the same clean up commit. What do you think?

cloudformation/tasking-manager.template.js
```diff
@@ -0,0 +1,344 @@
+const cf = require('@mapbox/cloudfriend');
```
xamanu: Is it possible to move this file and directory into our `devops` directory?

arunasank: yep! 👍

Dockerfile
arunasank: yep!

arunasank:
Thanks @dakotabenjamin! Going to quickly rebase here, per @xamanu suggestions for commit break up.

> * Introduce cloudformation setup with this file only `cloudformation/tasking-manager.template.js` 
> * Update Circle CI configuration with this file  `.circleci/config.yml` and removing `devops/ci/circleci_deploy.sh`
> * removed elastic beanstalk setup: remove `Dockerfile` and other EB related files.
> * introduce packaging: `package.json` and `requirements.txt`

dakotabenjamin:
Sounds good! once @xamanu approves I'm happy to hit the button

xamanu:
It would be nice to have the commits groomed again?

xamanu:
Yeay. This is awesome!

PR #1334, opened 2019-01-27
add "cursor: pointer" to buttons and remove unneeded blank lines

willemarcel:
A lot of buttons like those below didn't have the `cursor:pointer` css property

captura de tela de 2019-01-27 13-23-00


ethan-nelson:
Thanks!

PR #1335, opened 2019-01-30
Task datasets

ethan-nelson:
Originally from #1302 

This implements task datasets in Tasking Manager. Users can create a project by uploading a geojson file with geometries. Those geometries are split on task boundaries and stored with the task. There is an included JOSM plugin that can be installed. When a user clicks start mapping on that task, the task data will be passed to JOSM as a new layer. The user can then use the task datasets as guides while mapping. This is useful when geometries are available from another source (such as roads or buildings derived by machine learning) that we don't want to do a bulk import from, but we do want to include as a reference layer while mapping.

When generating the project, any tasks without task data are removed. When a task is split, the task data is also split and associated with the new tasks.

This addresses #1279.

c.c. @TimuSumisu


ethan-nelson:
As of now, the version of turf has been upgraded for these changes and is now in the `turf/turf` namespace. However, I need to read changelogs and test further to ensure there aren't any features that are not backwards compatible with the new version since this is an upgrade of that library across the full TM app.

ethan-nelson:
Just an update: talked to Jacob offline last week and he is in the process of updating the JOSM plugin to fix the issue.

xamanu:
Hey, thanks for this PR. This is going to be very useful!

I was wondering, whether this could also cover and include the work/experience of @zlavergne's import functionality - so pinging him here, to ask him for his input. It would be great if this PR could also cover his use case.

As far as I know he is using the .osm format instead of .geojson for giving to the tasks. @ethan-nelson @lesserj, would be be difficult to abstract the format we are using here? What do you think about this?

zlavergne:
@ethan-nelson @lesserj I'm interested what you two think. The main difference of my feature is that it splits the data based on a task when the user clicks "Start Editor". So there's no need to break the data up prior to project creation. This retains the ability to split tasks without needing to reprocess the data. 

For my feature, project creation is the same as before. You just upload a file or files in the Project Editor. Due to this, there are still tasks where there isn't provided data. Which it sounds like you avoid with this feature.

I think it would be great to try to merge our workflows. Let me know what you think. I think abstracting my data extraction would probably make it more applicable to your workflow.

I'm also really interested in the JOSM plugin!

xamanu:
This PR is having some conflicts. @ethan-nelson can you please rebase on `develop` and check issues. Thanks.

xamanu:
This has been unable to catch up with the current developments. Closing here for now. Feel free to reopen. Thank you.

PR #1289, opened 2018-12-06
Task annotations

geohacker:
Work in progress to address https://github.com/hotosm/tasking-manager/issues/1284

POST /project/<id>/task_annotations/<type>
example /project/1/task_annotations/ml
post body:
```
{
    'projectId': 1,
    'tasks': [
        {
            'taskId': 1,
            'annotationSource': 'https://some-url.com/ml',
            'annotationType': 'ml',
            'properties': {
                'building_area_ml_pred': 0.7,
                'building_area_osm': 0.8
            }
        },
        {
            'taskId': 2,
            'annotationSource': 'https://some-url.com/ml',
            'annotationType': 'ml',
            'properties': {
                'building_area_ml_pred': 0.4,
                'building_area_osm': 0.2
            }
        }
    ]
 }
```

GET /project/<id>/task_annotations/<type>
example /project/1/task_annotations/ml
response:

```
{
    "projectId": 1,
    "annotationType": "ml",
    "tasks": [
        {
            "taskId": 1,
            "annotationType": "ml",
            "annotationSource": null,
            "properties": {
                "building_area_ml_pred": 0,
                "building_area_osm": 0.8
            }
        },
        {
            "taskId": 2,
            "annotationType": "ml",
            "annotationSource": null,
            "properties": {
                "building_area_ml_pred": 0,
                "building_area_osm": 0.2
            }
        }
    ]
}
```

cc @ethan-nelson @wronk @batpad @smit1678

Issue #1284, opened 2018-11-26
ML prediction storage in TM


geohacker:
Related to #1264. 

@smit1678 @ethan-nelson - @wronk @batpad and I are thinking a bit more about integrating the ML results to the front-end. Currently, we aren't storing the predictions to the ML db and we think we should store this per task. So this would involve:

1. Changes to the Tasks model or a new model
Tasks already have an `extra_properties (varchar)` column which @batpad and I suspect is carried over from some previous work Sanjay did but is unused. We could turn this column into a jsonb type and store ml results as key value pairs for look up and updates. The other idea and may be slightly cleaner is to create a new model altogether and call it MLPredictions and store `task (id), ml_prediction (json)` and then use this for lookup.  This would allow us to keep Tasks model independent, and also expand later. @ethan-nelson do you have a suggestion or preference?

2. API endpoint to POST project/tasks geojson 
Depending on 1, we can figure out where this endpoint should go.

3. Front-end visualisation
We'll use the [current structure](https://github.com/hotosm/tasking-manager/issues/1264#issuecomment-440191005), and add another button to visualise ML results and pull the attributes to the front-end.

wronk:
## Updated plan
After getting some community feedback, our updated plan is to create a separate "Task Properties" table within the TM database. We'll also build API endpoints to interface with that new table.

The advantage of this strategy (pointed out by a number of people) is that this approach is useful for any generic task annotation (where ML-derived information is only one example). We could also use this for task prioritization (e.g., based on humanitarian needs or search engine hits) or other task-relevant info that can be stored in JSON format.

## Task properties table
**Columns**
* Project ID
* Task ID
* Annotation type (enum) to help front-end parsing of task properties
* Annotation source (str? Optional)
* Timestamp of last update
* Properties (jsonb) with key/value pairs containing information about a task

**API endpoints**
* POST requests to add/update task properties
    * Eventually (or immediately? cc @ethan-nelson), this should make use of an API key to limit access
* GET requests to fetch task properties
    * Will be called from TM's front-end
    * Once fetched, properties can be added to a project's geojson file for visualization

lesserj:
Take a look at #1302.  It allows projects to be created by geojson data.  Geojson data is split by task boundaries and stored per task.  Tasks without data are discarded.  There is no visualization component in Tasking Manager, but it does include a plugin for JOSM that allows the data to be passed to JOSM when clicking the start mapping button.

xamanu:
[...]

PR #1349, opened 2019-02-15
Task Annotation API 

geohacker:
This PR introduces the Task Annotation API:

* Introduces API end points to allow registered services to post JSON annotations
* And for the client to fetch annotations to render on the map
* Currently the map visualization [is hardcoded](https://github.com/hotosm/tasking-manager/pull/1289/files#diff-744e6c34c74ef9f175a41528b267f851R829) on `ml` type annotation and `building_area_diff` as the property, assuming the annotations are for building area prediction from an ML model. 
* The map allows to visualize annotations, and shows a legend.
* The API is [documented here](https://tasks-demo.hotosm.org/api-docs/swagger-ui/index.html?url=https://tasks-demo.hotosm.org/api/docs#!/project-admin/post_api_v1_project_project_id_task_annotations_annotation_type), and you can read the background [here](https://github.com/hotosm/tasking-manager/issues/1284) and [here](https://github.com/hotosm/tasking-manager/pull/1289).  
* Example of a [payload JSON is here](https://github.com/hotosm/tasking-manager/pull/1289#issuecomment-448905552).
* See a [demo task here](https://tasks-demo.hotosm.org/project/4662).

This PR also includes @ethan-nelson's application token work. 

cc @wronk @smit1678


ethan-nelson:
sorry--wasn't clear that this is ready for review. i will do so this week!

ethan-nelson:
Finishing up some testing still (I think there is a bug in the application key generation that doesn't allow deletion). We are also hopefully doing a develop->master merge soon, which will allow for a clean slate with some new changes. 

(If anyone else is testing this in the interim using the develop branch, you will need to change L14 of migrations/versions/05e1ecf9953a_.py to `3ee58dee57c9`.)

ethan-nelson:
More updates. I used the online GitHub conflict resolution tool to update the `client/package.json` file--there was one package missing between tm4ml and develop, so that is now updated.

I've also isolated the issue with the inability to delete keys: the angular function was not defined properly. In addition, I found an error thrown on the sort parameter so I went ahead and removed that. Here is the diff for now--trying to figure out the easiest way to add a new commit.

```
diff --git a/client/app/profile/application-key.controller.js b/client/app/profile/application-key.controller.js
index 5f9f585..326e26d 100644
--- a/client/app/profile/application-key.controller.js
+++ b/client/app/profile/application-key.controller.js
@@ -69,7 +69,7 @@
          * Delete a user's application key
          * @param applicationKey
          */
-        function deleteApplicationKey(applicationKey){
+        vm.deleteApplicationKey = function(applicationKey){
             var resultsPromise = applicationKeyService.deleteApplicationKey(applicationKey);
             resultsPromise.then(function (data) {
                 vm.deleteResult = data
diff --git a/client/app/profile/application-key.html b/client/app/profile/application-key.html
index 50e3589..b04aec7 100644
--- a/client/app/profile/application-key.html
+++ b/client/app/profile/application-key.html
@@ -29,7 +29,7 @@
                                         <th><span>{{ 'Revoke key' | translate }}</span></th>
                                     </thead>
                                     <tbody>
-                                        <tr ng-repeat="item in appKeyCtrl.applications | orderBy: 'createdDate'">
+                                        <tr ng-repeat="item in appKeyCtrl.applications">
                                             <td>{{ item.createdDate }}</td>
                                             <td>{{ item.applicationkey }}</td>
                                             <td><a ng-click="appKeyCtrl.deleteApplicationKey(item.applicationkey)"><i>X</i></a></td>
```

xamanu:
We have to reroll this on the latest `develop`.

xamanu:
Rebased on latest `develop`, fixed conflicts on the way.

client/app/project/project.html
```diff
@@ -1053,7 +1053,14 @@ <h6>{{ 'Options' | translate }}</h6>
                             </a>
                         </div>
                     </div>
-                    <div class="ol-control legend-control">
+                    <div ng-show="projectCtrl.projectData.annotations">
```
xamanu: Would it make sense to separate the general Task Annotation work and the one that addresses the specific application as an ML layer on the project page?

xamanu:
CI fails because of deployment logic. I opened an issue #1488 to spinn up a new ml playground instance and deprecate the auto deployment on this branch. This shall fix the issue of the CI.

ethan-nelson:
Thanks @xamanu

xamanu:
Perhaps it makes sense to separate the backend work (#1577), that seems already to be at a good shape and could find a place into the main tasking manager code base. While the front-end needs much more work, still (#1578, #1579).

Would you be ok with closing this PR in favor of the follow ups, @geohacker?

geohacker:
> Would you be ok with closing this PR in favor of the follow ups, @geohacker?

@xamanu sure thing. So it looks like to me that:
* backend code is now moved to #1577 - it'll be great to either merge this, or I can continue working on the improvements coming up over there.
* there's some ML visualisation as part of this PR, which I think is valuable to show in the short-term but it's your call whether to surface that or wait for #1578 and #1579.

xamanu:
Great! Closing here.

> backend code is now moved to #1577 - it'll be great to either merge this, or I can continue working on the improvements coming up over there.

Yes, let's bring this into the code base.

> there's some ML visualisation as part of this PR, which I think is valuable to show in the short-term but it's your call whether to surface that or wait for #1578 and #1579.

Yes, we will do a PR with the front-end example. We can use this to deploy on the [tasks-assisted](https://tasks-assisted.hotosm.org) to have a running example. And then work on #1578 and #1579 for a general implementation that could become part of the main code base.

PR #1366, opened 2019-03-02
Add a few locales, update translations

ethan-nelson:
Three languages were added to the server config already but didn't have corresponding locale files (DA, LT, SI)
One language had corresponding locale files but wasn't in the config (BN)

We have quite a few translations (including new strings) since the last update, too, so I have synced changes from Transifex.


ethan-nelson:
Make sure you `gulp build` in the `client/` directory to test this

ethan-nelson:
@xamanu or @smit1678 or @dakotabenjamin would you mind checking this over in the next day or two? I'd like to make sure this gets merged before we do the next version deploy.

xamanu:
Excuse me @ethan-nelson for taking that long. This compiles fine and new languages show up. Thanks for the PR!

Issue #1153, opened 2018-08-06
Refactor project page refresh


ethan-nelson:
The basic idea here is to create a new endpoint for project refreshes that contains the project metadata and _only a portion_ of the task information after initial loading.

### Background 

As a user sits on a project's page--mapping, validating, or watching--their browser is polling the `/project/#` endpoint [every 10 seconds](https://github.com/hotosm/tasking-manager/blob/1fdba13050a3f8dc3b7a16547858cdf74e3aef4a/client/app/project/project.controller.js#L130). This [API endpoint](https://github.com/hotosm/tasking-manager/blob/1fdba13050a3f8dc3b7a16547858cdf74e3aef4a/server/api/project_apis.py#L13-L75) returns to the user's browser the full project metadata and information about every task: geometry, status, history ([example here](https://tasks.hotosm.org/api/v1/project/3000)). 

While this information is necessary to initially populate the project page, it is not necessarily needed for every refresh cycle. We really only need information to keep the map animated with task statuses (as users lock, unlock, mark mapped, mark bad imagery, validate, or invalidate tasks; or when priority areas are modified), though there is a caveat related to task splitting that will be addressed below.

### Work required

This will touch on both backend and frontend code.

On the backend, a new API endpoint will be required (something like `/project/#/status-update` or a better verb there). This endpoint either needs a new project service and project model function to provide this, or function calls added to the [existing project service](https://github.com/hotosm/tasking-manager/blob/1fdba13050a3f8dc3b7a16547858cdf74e3aef4a/server/services/project_service.py#L46) and [model function](https://github.com/hotosm/tasking-manager/blob/1fdba13050a3f8dc3b7a16547858cdf74e3aef4a/server/models/postgis/project.py#L377-L384) to not include coordinates.

For the frontend, the current refresh mode will be utilized initially (and perhaps every few minutes just for safety's sake), while the update endpoint will be used in between at a similar frequency to the current refresh cycle. The js will need a new function to poll the new point, read the update version of the json, and update the task status on the map based on id number of each feature (since geometry is not included). Furthermore, **it is important to note** the number of tasks will need to be retained and compared against every subsequent update. This is because tasks can be split, altering their geometry, ids, and count. If the frontend notices the number of tasks has changed, it should revert back to the original endpoint to update the geometry of the task features along with everything else.

### Refactor benefits

[...]

PR #1368, opened 2019-03-03
Task diff update

ethan-nelson:
Please see #1153 for full background

This PR implements an "abbreviated" project API endpoint that does not contain geometry for the individual tasks and updates to the frontend to handle this. This behavior is only used for routine refreshing (the ~10 second idling refresh when a user is viewing the project screen without any interaction).

For routine refreshing, this is sufficient to change the states of tasks on a user's screen as they change state in the database. The core exception to this is when tasks are split, which is handled on the frontend by checking the length of the task layer currently loaded versus that sent by the server; if they are not equal, more tasks exist on the server (precisely 3 if a square is split) and it falls back to the original refresh behavior.

For all other refreshing, i.e. when a user performs an action on a task, the original refresh behavior is retained for that user at that point. This is a pretty conservative approach as it's not probably necessary, but I rather minimize the number of things changed at once. We can always phase in other places where we add this optimization.

Note this is complementary to possibly other methods of streaming updates between the client and the server, assuming those can fall back to this if that method is not enabled.

Beyond whether it works or not (I have tested this locally with multiple users doing different actions including splitting), I'm looking for feedback on the frontend code and how it can/whether it should be refactored.


ethan-nelson:
Update: I did refactor some of the frontend code after I had slept overnight. The changes are now much more minimal. Now, we substitute the function for the 10 second refresh with a new one that tries to do the abbreviated update, and if there's any issue, that function falls back on calling the original. Hopefully this makes it much more straightforward.

ethan-nelson:
This is ready for review if someone wants to assign themselves @xamanu @ramyaragupathy @willemarcel

xamanu:
I would like to review it. There seems to be a conflict with the latest version of the `develop` branch. Can you please rebase and check, @ethan-nelson? Thanks.

ethan-nelson:
Done

xamanu:
Rerolled and moved this PR to a local branch in #1514

PR #1393, opened 2019-03-17
Update Dockerfile 

xamanu:
Update Dockerfile:

* to Debian stable (Stretch)
* some smaller typo changes for better readability
* Consolidate port to `5000`
* Prepare for development setup use of the container
* Add `.dockerignore`
* Use `/usr/src/app` instead of `/src`

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,22 @@
+# Basic stuff
+Dockerfile*
+docker-compose*
+.dockerignore
+.git
+.gitignore
+
+# Python
+venv/
+*pyc
+
+# Dev artifacts
+client/node_modules/
+client/assets/styles/css
+shippable/
+
+# Programms that like to leave stuff
+.DS_Store
+.vscode
+.idea/
+.elasticbeanstalk/
+venv_aws/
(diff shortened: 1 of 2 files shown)


arunasank:
@xamanu, thanks for this PR! With the move to Cloudformation, we'll not be needing this file anymore for our deploys. Do you think we should still be keeping the file around?

xamanu:
It is still going to be nice to provide a Dockerfile of the Tasking Manager for an easy testing and deployment. Maybe we want to move it into the `devops` folder, once we stop using it directly.

xamanu:
By today, the python images based on Debian Jessie are not building anymore. This means that people building the Tasking Manager on Docker will not be able to do so.

Raising the priority here to critical.

xamanu:
Changed port back to `8000`, thus not to introduce any breaking changes

ethan-nelson:
trying to determine why #1394 builds while this one does not

xamanu:
Because #1394 was rebased on `develop` after #1413 went it. This one hasn't been rebased, but was now and builds.

xamanu:
I'm going to reroll this PR.

PR #1400, opened 2019-03-19
Show 404 error on using in-valid message id.

uwaiszaki:
The Pr handles
1-> Invalid message id.
2-> Any other error loading the message. And
3-> Add error status in getMessage service.

### DIFF
--- a/client/app/message/message.controller.js
+++ b/client/app/message/message.controller.js
@@ -14,6 +14,7 @@
         vm.message = {};
         vm.showDeleteMessageModal = false;
         vm.errorRetrievingMessages = false;
+        vm.errorMessage = [];
 
         activate();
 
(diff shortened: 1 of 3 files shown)


xamanu:
Thanks for this PR! I'm wondering whether we wanted to show a `404` here, or whether it would be more convenient to move people to login and redirect them afterwards to the url they were trying to hit at. In case it is not their account's message, I would present an `403`. What to you think?

uwaiszaki:
@xamanu The redirecting part is already working if they are not logged in. In case they are logged in
, I think for using an invalid message id we should show "not found" instead of showing "not allowed or blocked".

uwaiszaki:
@xamanu @JorgeMartinezG I have implemented the requested changes. Please give ur reviews

JorgeMartinezG:
Thank you again for this contribution @uwaiszaki. I found this approach correct. @xamanu do you have additional opinions?

uwaiszaki:
@xamanu It's working fine with my setup. Can u please check it again.

uwaiszaki:
tmpull1400

PR #1401, opened 2019-03-20
Add Mark as read feature in inbox

uwaiszaki:


uwaiszaki:
@xamanu To add Mark as read should I use a patch request or a simple get request(getMessage) as on using get request it is marking the message as read.

uwaiszaki:
@xamanu @ethan-nelson @ramyaragupathy @smit1678  Now messages are grouped by their "read" status.
How should I add sorting according to date?

uwaiszaki:
I have added a patch request in the MessageAPI to change read status without visiting the message page.

xamanu:
@uwaiszaki, may you please reroll this on the lastest `develop` and I suggest to use #1423 as a base, too.

uwaiszaki:
@xamanu this will add "mark as read" feature, in continuation of #1423.

xamanu:
Can you please rebase, @uwaiszaki?

server/api/messaging/message_apis.py
```diff
@@ -238,6 +238,50 @@ def delete(self, message_id):
             current_app.logger.critical(error_msg)
             return {"error": error_msg}, 500
 
+    @tm.pm_only(False)
+    @token_auth.login_required
+    def patch(self, message_id):
```
uwaiszaki: @xamanu I have used a patch request here.

uwaiszaki:
@xamanu I have rebased the pr with develop branch, please give your reviews.

xamanu:
Just reviewed. Works well, but I'm not sure whether this is the desired behaviour:

image

I would prefer to reuse the checkbox on the beginning and out of the "Delete all selected" I would combine it another action "Mark as read all selected messages"

path2-53-7-8-2

@willemarcel @russdeffner What do you think?

willemarcel:
I agree with you, @xamanu . We could make the unread messages bold to distinguish it from the read ones.

willemarcel:
Other thing: could we put the filter line outside the messages table?

uwaiszaki:
@willemarcel @xamanu I think it would be really confusing for the user if the checkbox is not marked after marking it as read as the same checkbox will be used for deleting multiple messages. Can we show the read messages on another page? ex. like the second page in the Gmail inbox

uwaiszaki:
This currently adds API and services to mark/unmark multiple messages as read. Placement of delete and mark buttons as a settings menu is remaining.

xamanu:
I don't think the read status should be indicated by a checked checkbox. Also the redesign separates the status indicator (line on the left of each message) with the function to mark it as read.

xamanu:
As per conversation with @uwaiszaki targeting this for the redesign and shall be implemented with React.

Issue #1197, opened 2018-10-03
Add ability to generate auth token in user profile


ethan-nelson:
Right now, there is no way for a user to generate an authentication token for API except to carve it out of the HTTP headers after logging in. This is a request to add an authentication token generation ability in the user profile. It would ideally be a button you click that either opens a modal with the key or populates a disabled text box near it with the key.

This will require creating an API endpoint:

1. That receives a user authentication token;
2. If valid, [generates another user authentication token](https://github.com/hotosm/tasking-manager/blob/develop/server/services/users/authentication_service.py#L124-L134); and
3. Returns the new authentication token.

Additionally, the front end will then need:

1. A new promise to access that endpoint;
2. A button to activate the promise; and
3. Populate either a model or a disabled text box with the received auth token.

ethan-nelson:
I split the up into two components because the API endpoint could feasibly be added first separately, then the front end added later by another person or PR.

jtj9817:
I would like to try to work on this issue!

ethan-nelson:
Great, I will mark it in progress!

uwaiszaki:
@ethan-nelson I would like to work on this.

ethan-nelson:
Hi, we've actually transformed this a bit in #1349. Within that is the ability for an application based auth token to be generated. The issue with exposing the ability to generate regular user auths to web users is that a compromised account could be compromised in perpetuity by a malicious actor generating a new auth token before the previous one expired. Instead, the application based auth token allows certain users to generate a new class of tokens for auth that can be revoked at any time instead.

ethan-nelson:
In fact, I'm going to go ahead and close this issue since I consider the original ability no longer preferable. Thanks for the interest in this one, though!

PR #1407, opened 2019-03-24
[WIP] Add ability to generate auth token in user profile

uwaiszaki:
Backend->
1) This creates a Token API that receives a user authentication token by authorization headers.
2) If valid, generates another user authentication token. and
    returns the new authentication token.


xamanu:
@uwaiszaki, sorry, our `develop` branch was rerolled. May I ask you to rebase this PR on the current `develop` branch, please. Thanks! :)

xamanu:
This PR needs to be rebased on latest `develop`.

JorgeMartinezG:
Hi @uwaiszaki . Thanks for contributing again. I have some questions: 

1. Could you please explain the purpose of generating a new token if the one which is read from headers is already valid?
2. Is possible to modify the API endpoint to different value like **/api/v1/auth/validate**? the reason is that **/auth/token** is reserved for token exchange between application and authorization servers when performing the authorization_code protocol for oauth2.

This one is more a comment, rather than a question:
3. I think that for APIs the server should keep consistency among data returned. With this PR, If the token is valid the format is string. When there is an error, an int value is returned. One possible option is to always return a json response. Something like:
```
Token ok --> {"status": 200, "token": "access_token"}
Not found --> {"status": 404, "error": "Token not found"}
Not valid --> {"status": 400, "error": "Invalid token"}
```

Again, thanks for this PR. Let me know your opinion! 😄

ethan-nelson:
Please see the discussion at https://github.com/hotosm/tasking-manager/issues/1197#issuecomment-476454764 for more info--there has been a change of thinking and I don't think we want to enable this.

PR #1416, opened 2019-03-27
Improve Dockerfile

xamanu:
This PR is a predecessor of the #1393.

**Update Dockerfile:**

* Include front end packages in the container.
* some smaller typo changes for better readability and sticking to best practices in Docker containers
* Prepare for development setup use of the container (Add .dockerignore).

This PR is not introducing any breaking changes.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,25 +1,49 @@
+# Base on Python on Debian stable
 FROM python:3-stretch
 
-# Install dependencies for shapely
-RUN apt-get update \
- && apt-get upgrade -y \
- && apt-get install -y libgeos-dev \
- && rm -rf /var/lib/apt/lists/*
-
-
-# Uncomment and set with valid connection string for use locally
-#ENV TM_DB=postgresql://user:pass@host/db
-
-WORKDIR /src
-
-# Add and install Python modules
-ADD requirements.txt .
-RUN pip install --no-cache-dir -r requirements.txt
-
-ADD . .
-
-# Expose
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -1,25 +1,49 @@
...
-#ENV TM_DB=postgresql://user:pass@host/db
+# Install dependencies
+RUN apt-get update \
+  && apt-get upgrade -y \
+  && apt-get install -y libgeos-dev \
+  && apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
-WORKDIR /src
+RUN mkdir /usr/src/app
+WORKDIR /usr/src/app
```
ethan-nelson: For changing the working directory, I think we need to make a few more modifications (thanks to experience in #1414).

* The directory listed in https://github.com/hotosm/tasking-manager/blob/develop/.ebextensions/aws.config#L12
* Possibly locations where `/app` is mentioned in here? https://github.com/hotosm/tasking-manager/blob/develop/circle.yml

xamanu: * Right, especially for the `aws.config` file! I removed there the `cd /app` command entirely, because the entrypoint should have been defined through the `WORKDIR` anyway.
* I think for the circle it doesn't really matter (it is a separate definition), but if we change to best practices here, I agree we should also do there. I updated the `circle.yml`.

xamanu:
Reviewed and approved in #1419

PR #1426, opened 2019-03-28
Update python dependencies in the AWS Docker container

xamanu:


dakotabenjamin:
It's failing because Circle is building on node 10 for this PR. I can still merge it in after a review and see if develop deploys.

PR #1433, opened 2019-04-01
Add frontend building in Dockerfile

xamanu:
This one should (hopefully) fix the current issues we are facing with the Elastic Beanstalk deployment on the `develop` branch.


xamanu:
Thanks to @JorgeMartinezG for spotting the `client` line in the `.ebignore` file!

xamanu:
This needs more investigation.

PR #1437, opened 2019-04-02
Fix Elastic Beanstalk deployment

xamanu:
This PR fixes the issues we were having with the AWS Elastic Beanstalk deployment. It has been tested [in another branch](https://github.com/hotosm/tasking-manager/commits/modify/Dockerfile), which contains some temporal configuration on top of this PR to get the deployment testing going.

Please make sure the (new) environment variables are set for the staging and live site. Particularly these two:

* `TM_APP_BASE_URL`=`https://tasks.hotosm.org` (or similar)
* `TM_LOG_DIR`=`/var/log/tasking-manager-logs`

### DIFF
--- a/.ebextensions/environment.config
+++ b/.ebextensions/environment.config
@@ -1,7 +1,3 @@
-# Environment settings
-option_settings:
-  - option_name: TM_ENV
-    value: Staging
 files:
   "/tmp/45_nginx_https_rw.sh":
     owner: root
--- a/.ebextensions/https.config
+++ b/.ebextensions/https.config
@@ -1,7 +1,3 @@
-# Environment settings
-option_settings:
-  - option_name: TASKING_MANAGER_ENV
-    value: Staging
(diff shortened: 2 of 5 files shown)


xamanu:
JFYI, I prepared the environment variables for our deployment for `taskingmanager-stage` and `taskingmanager-experimental` (`tm3-db-stretch`).

dakotabenjamin:
I see you already updated the envvars for staging, can we safely do it to the other environments without breaking things?

xamanu:
Yes, there are no conflicts. Basically we are required to add:

* `TM_APP_BASE_URL`
* `TM_LOG_DIR`

And then it is recommended, but not necessary to add the new way of doing the data base connection (instead of `TM_DB`):

* `POSTGRES_DB`
* `POSTGRES_ENDPOINT`
* `POSTGRES_PASSWORD`
* `POSTGRES_USER`

In a first step we can just add them. And once it we deployed the `TM_DB` can be removed.

Issue #1435, opened 2019-04-01
Blocked User Can Comment


russdeffner:
Currently a user that is blocked (i.e. Read_Only) can still comment on projects. This should be changed to ensure that a blocked user can not perform any activity on a project.

uwaiszaki:
@russdeffner I would like to work on this.

xamanu:
This has been addressed within release 3.2.1 for the backend. But the front end should hide the possibility to add comments. This is work in progress in #1438.

xamanu:
#1438 and https://github.com/hotosm/tasking-manager/pull/1431 went into `develop` for the 3.4 release.

PR #1438, opened 2019-04-04
Block read_only user comments in project chat

uwaiszaki:
Fix for #1435


ethan-nelson:
Hi @uwaiszaki thanks for your contribution on this! I think the approach may need to be altered a bit.

Right now, since the changes are on the frontend, someone that is blocked could use the API to add project chat. So, first, we should add a check at https://github.com/hotosm/tasking-manager/blob/develop/server/services/messaging/chat_service.py#L18 that ensures the chat member does not have a READ_ONLY role attached to their user object.

Additionally, I think it would be more straightforward on the frontend to just `ng-hide` the chat box at https://github.com/hotosm/tasking-manager/blob/develop/client/app/components/project-chat/project-chat.html#L18-L39 if a user's role is read only so that the user has a direct signal they cannot participate. That could be done by using https://github.com/hotosm/tasking-manager/wiki/Restricting-features-to-certain-users.

Let me know your thoughts. Also, if you need any guidance, I'm happy to assist as well. Cheers!

xamanu:
A backend quick fix has been applied. Let's continue on this one here to implement it properly as started by @uwaiszaki.

JorgeMartinezG:
Thanks @uwaiszaki for this contribution 👍 

Updating this PR, these changes disallow read_only users to view chat messages. I think the only restriction would be to post messages. 

I think one quick fix on the frontend side might be a `ng-show` directive which checks for user role:
https://github.com/hotosm/tasking-manager/blob/develop/client/app/components/project-chat/project-chat.html#L18

@xamanu @uwaiszaki let me know your opinions 🤓

xamanu:
I think we want the interface to appear like a user can not comment -  instead of currently presenting an error message when trying to comment. But we still want to show comments to read_only users. Comments are also visible to logged out users.

uwaiszaki:
@xamanu @JorgeMartinezG Should I just hide the comment box or showing this message is fine?
tmpull1438

JorgeMartinezG:
I think hiding comment box and message button is OK. The issue with displaying message is that it would become necessary to translate into mulitple languages @uwaiszaki

xamanu:
Yes, I also think to hide the text box and button is just fine. To display the error is a bit strong. Alternatively we could present a simple text, not red, without the word "Error" that gives feedback like for example "You can not comment".

uwaiszaki:
Thanks, @JorgeMartinezG @xamanu for your reviews. I have updated the pr.
Now the comment box is disabled with a simple text message "You can not comment".

JorgeMartinezG:
I consider this PR ready to be merged. @xamanu do you have additional comments?

xamanu:
Rerolled this and [merged into `develop`](https://github.com/hotosm/tasking-manager/commit/52260a0287bc7b6712f8c8220982f0b2db20ca66). Thanks. Closing here.

PR #1451, opened 2019-04-09
Include a few missing profile translate strings

ethan-nelson:
Catching a few strings on the profile page that weren't translated.

### DIFF
--- a/client/app/profile/profile.html
+++ b/client/app/profile/profile.html
@@ -60,7 +60,7 @@
                             </div>
                             <div>
                                 <h4>Stats</h4>
-                                <span>Time Spent Mapping (TM3 only)</span>
+                                <span>{{ 'Time Spent Mapping (TM3 only)' | translate }}</span>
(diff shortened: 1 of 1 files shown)


client/app/profile/profile.html
```diff
@@ -60,7 +60,7 @@ <h4>{{ profileCtrl.username }}'s {{ 'Progress' | translate }}</h4>
                             </div>
                             <div>
                                 <h4>Stats</h4>
-                                <span>Time Spent Mapping (TM3 only)</span>
+                                <span>{{ 'Time Spent Mapping (TM3 only)' | translate }}</span>
```
xamanu: I'm wondering whether "TM3 only" is still something we want to print here?

xamanu: I suggest to rephrase more generally: `Time Spent Mapping (on this instance)`

ethan-nelson:
Agreed--updated!

ethan-nelson:
I will work on updating strings in Transifex for translation.

PR #1453, opened 2019-04-09
Data dump for local testing

ramyaragupathy:
Uploaded @willemarcel's local DB dump. Added a section to [Devops wiki](https://github.com/hotosm/tasking-manager/wiki/Dev-Ops#use-an-existing-data-dump) on steps to  build from dump.

cc @xamanu @dakotabenjamin @ethan-nelson


xamanu:
Great you got this started!

I'm just wondering about the wording of the files, as we might include more than just tasks in the future. What about using `/tests/database/tm.sql`?

xamanu:
And please use the new environment variables for the documentation:

`POSTGRES_DB`, `POSTGRES_ENDPOINT`, ...

JorgeMartinezG:
I executed migration using the file. I was able to visualize records from this dump using **/api/v1/project/search** endpoint.

hunt3ri:
As per prompting from @willemarcel I'm noting my suggestion here that to make it easy for people to start hacking on the project it should be feasible to either load the dump file from within the docker-compose file, or if not write a simple script for people to run to load the dump file into a local developement environment.

That way we have a one-liner in the README to let people start running the project immediately.

xamanu:
Autoloading and having a command would be great. Could be part of the Makefile #1492

Issue #1457, opened 2019-04-09
Disable http access to the tasking manager


arunasank:
Ref: https://github.com/hotosm/tasking-manager/issues/1395

Currently we allow both https and http access to the tasking manager. Lets modify the ELB listeners to disable HTTP access to the tasking maanger, or use nginx to do this.

cc/ @dakotabenjamin

arunasank:
This works, because the application load balancer allows for a way to redirect http requests to https. 🎉. Merged https://github.com/hotosm/tasking-manager/pull/1463.

PR #1463, opened 2019-04-09
Allow only HTTPS access to the tasking manager

arunasank:
Currently, the tasking manager does not enforce the use of https. This PR enforces https use - the stack is currently set up at https://tasks-secure.hotosm.org/. You can see that `http://tasks-secure.hotosm.org/` does not work, while `https://tasks-secure.hotosm.org/` does. 

Fixes https://github.com/hotosm/tasking-manager/issues/1457

Screen Shot 2019-04-09 at 3 47 12 PM
Screen Shot 2019-04-09 at 3 47 23 PM

cc/ @dakotabenjamin @willemarcel @xamanu @smit1678 @russdeffner @ethan-nelson

### DIFF
--- a/cloudformation/tasking-manager.template.js
+++ b/cloudformation/tasking-manager.template.js
@@ -202,18 +202,6 @@ const Resources = {
       Protocol: 'HTTPS'
     }
   },
-  TaskingManagerLoadBalancerHTTPListener: {
-    Type: 'AWS::ElasticLoadBalancingV2::Listener',
-    Properties: {
-      DefaultActions: [{
-        Type: 'forward',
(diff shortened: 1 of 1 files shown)


arunasank:
Don't worry about the failing tests - this is a merge into `cloudformation`, and the tests will pass after the `cloudformation` branch has been rebased against `develop`. 😊

arunasank:
Thanks for testing this @willemarcel. It seems like we redirect from HTTP -> HTTPS on our production instance, and want the same behaviour here.

arunasank:
Redirects work now - since the ALB (as opposed to the classic load balancer we were using earlier) allows for requests to be redirected between source protocols. 🎉 

We tested this and it works!

PR #1466, opened 2019-04-10
Aligning master and develop

xamanu:
This is a proposal for re-aligning `master` and `develop` branch. It is meant to be brought in as a minor-minor release 3.2.1.

What do you think?


dakotabenjamin:
After merging this PR, a subsequent Release PR will be made?

What do you think if we simply swept these two releases into 1 (calling it 3.2.0). I haven't finished the release notes yet.

xamanu:
Yes, a subsequent PR can then be made from `develop` to `master`. Yes, as the release notes haven't been published I think we can include it in 3.2.0/.

Issue #1444, opened 2019-04-08
Set up Circle CI for automated deployments with CloudFormation


arunasank:
Ref: https://github.com/hotosm/tasking-manager/issues/1395

To complete the new setup, we must:
* [x] Migrate to Circle v2
* [x] Set up automated deployments for the Tasking Manager with circle ci + cloudformation. From https://github.com/mapbox/cfn-config

>  
```
-f, --force             perform a create/update/delete command without any
                            prompting, accepting all defaults
```

So, we can use preset templates on S3 + the `-f` option on circle ci to do automated deployments much like we were doing with Elastic Beanstalk.

arunasank:
cc/ @dakotabenjamin @hotosm/tech

dakotabenjamin:
From our slack convo: I had some confusion about the software deployment strategy that was cleared up (thanks @arunasank ) which I will describe here to illustrate some problems that need to be addressed. 

Using cfn-config as the deployment strategy, there is a GitSha parameter which will be different each time deployment happens. This hash corresponds to the commit being deployed and ensures a change in the LaunchConfiguration (because the UserData clones the Tasking Manager repo based on the commit). When this happens, the LaunchConfiguration will be replaces thus replacing all the EC2 instances with the new UserData and the new code. 

The problem therein is that there will not be 100% uptime during a deployment. There are ways around this see #issue ), but with in-place or rolling updates we have to address a more pressing issue around the database. When pushing software updates that update the database (i.e. a new alembic revision) there is a risk of mismatch between the running instances and the database. I'm not exactly sure what this failure will look like, but in the past with elastic beanstalk there have been issues where alembic update has failed due version mismatch. In any case we should avoid having multiple versions of tasking manager connected to a single database.

dakotabenjamin:
Looking for some advice on how to properly store the environment variables below:
```
      - run: 
          name: Create config file
          command: |
[...]

PR #1469, opened 2019-04-10
[WIP] Update CircleCI config for cfn strategy

dakotabenjamin:
This configuration uses CircleCI version 2.1 to leverage some new features such as reuable commands. It also sets up a workflow that simplifies the deployment process. 

It's a WIP because I need to figure out the best way to store and pull the configuration env files. I also do not think I have written a robust rds snapshot cycle and it should be iterated on. 

lastly, I just remembered that I can use some circleci commandline tools to test the config files locally so I need to do that as well (in the morning). 

cc @arunasank ref #1444


.circleci/config.yaml
```diff
@@ -0,0 +1,120 @@
+version: 2.1
+commands:
+  install-cloudfriend:
+    description: "Install mapbox/cloudfriend"
```
arunasank: 🐛 nitpick 🐛 : some of the descriptions need to be changed.

.circleci/config.yaml
```diff
@@ -0,0 +1,120 @@
...
+      - checkout
+      - setup_remote_docker
+      - install-cloudfriend
+      - run: 
+          name: Remove last snapshot and backup database
+          command: |
+            # Delete old snapshot
+            aws rds delete-db-snapshot tm3-staging-deployment-snapshot-latest
+            # create new aws rds snapshot
+            aws rds create-db-snapshot --db-snapshot-identifier tm3-staging-deployment-snapshot-latest --db-instance-identifier tasking-manager-staging
```
arunasank: Should we create a snapshot, before deleting the old one? This would mean some difficulties with naming, but would be safe in the scenario that the snapshot deletion was successful but the snapshot creation was not.

.circleci/config.yaml
```diff
@@ -0,0 +1,120 @@
...
+            aws rds create-db-snapshot --db-snapshot-identifier tm3-staging-deployment-snapshot-latest --db-instance-identifier tasking-manager-staging 
+      - run: 
+          name: Create config file
+          command: |
+            printf '{"GitSha":"$CIRCLE_SHA1", "Environment":"staging", "DBSnapshot":"$DB_SNAPSHOT_STAGING", "MasterUsername":"$MASTER_USERNAME_STAGING", "MasterPassword":"$MASTER_PASSWORD_STAGING", "OSMConsumerKey":"$OSM_CONSUMER_KEY_STAGING","OSMConsumerSecret":"$OSM_CONSUMER_SECRET_STAGING", "TaskingManagerSecret":"$TASKINGMANAGER_SECRET_STAGING", "TaskingManagerEnv":"$TASKINGMANAGER_ENV_STAGING", "TaskingManagerSMTPHost":"$TASKINGMANAGER_SMTP_HOST_STAGING", "TaskingManagerSMTPPassword":"$TASKINGMANAGER_SMTP_PASSWORD_STAGING", "TaskingManagerSMTPUser":"$TASKINGMANAGER_SMTP_USER_STAGING", "DatabaseSize":"$DATABASE_SIZE","ELBSubnets":"$ELB_SUBNETS_STAGING", "SSLCertificateIdentifier":"$SSL_CERTIFICATE_ID_STAGING","RDSUrl":"$RDS_URL_STAGING"}' > cfn-config-production.json
+      - deploy:
+          name: Deploy to Production
+          command: cfn-config update staging cfn-config-staging.json -f -c hot-cfn-config -t hot-cfn-config
```
arunasank: The contents of the `cfn-config-staging.json` should be copied over to s3://hot-cfn-config/tasking-manager/<tasking-manager-production-stack-name>-<tasking-manager-production-stack-region>.json`, since cfn-config reads this value from S3.

In the `cfn-config update .....` command, we should replace `cfn-config-staging.json` with the path to the cloudformation template.

.circleci/config.yaml
```diff
@@ -0,0 +1,120 @@
...
+      - checkout
+      - setup_remote_docker
+      - install-cloudfriend
+      - run: 
+          name: Remove last snapshot and backup database
+          command: |
+            # Delete old snapshot
+            aws rds delete-db-snapshot tm3-prod-deployment-snapshot-latest
+            # create new aws rds snapshot
+            aws rds create-db-snapshot --db-snapshot-identifier tm3-prod-deployment-snapshot-latest --db-instance-identifier tasking-manager-production
```
arunasank: same comments as above.

.circleci/config.yaml
```diff
@@ -0,0 +1,120 @@
...
+            aws rds delete-db-snapshot tm3-prod-deployment-snapshot-latest
+            # create new aws rds snapshot
+            aws rds create-db-snapshot --db-snapshot-identifier tm3-prod-deployment-snapshot-latest --db-instance-identifier tasking-manager-production 
+      - run: 
+          name: Create config file
+          command: |
+            printf '{"GitSha":"$CIRCLE_SHA1", "Environment":"production", "DBSnapshot":"$DB_SNAPSHOT", "MasterUsername":"$MASTER_USERNAME", "MasterPassword":"$MASTER_PASSWORD", "OSMConsumerKey":"$OSM_CONSUMER_KEY","OSMConsumerSecret":"$OSM_CONSUMER_SECRET", "TaskingManagerSecret":"$TASKINGMANAGER_SECRET", "TaskingManagerEnv":"$TASKINGMANAGER_ENV", "TaskingManagerSMTPHost":"$TASKINGMANAGER_SMTP_HOST", "TaskingManagerSMTPPassword":"$TASKINGMANAGER_SMTP_PASSWORD", "TaskingManagerSMTPUser":"$TASKINGMANAGER_SMTP_USER", "DatabaseSize":"$DATABASE_SIZE","ELBSubnets":"$ELB_SUBNETS", "SSLCertificateIdentifier":"$SSL_CERTIFICATE_ID","RDSUrl":"$RDS_URL"}' > cfn-config-production.json
+      - deploy:
+          name: Deploy to Production
+          command: cfn-config update production cfn-config-production.json -f -c hot-cfn-config -t hot-cfn-config
```
arunasank: same comments as above.

.circleci/config.yaml
```diff
@@ -0,0 +1,120 @@
+version: 2.1
+commands:
+  install-cloudfriend:
+    description: "Install mapbox/cloudfriend"
+    steps:
+      - run: npm install -g @mapbox/cfn-config
```
arunasank: Looks like the npm install below (https://github.com/hotosm/tasking-manager/pull/1469/files#diff-19f445bfad95516224bbe1795cd87b36R34) will not install `@mapbox/cloudfriend`, since we have more than one `package.json` (One in the root directory and one in the `client/` directory). Any thoughts around how we can make this less messy? 😬 Otherwise, for now, we'll have to explicitly run npm install in the root TM directory as well.

.circleci/config.yaml
dakotabenjamin: This can be fixed by copying the old one to a temp name, then doing the deployment, then safely deleting the temp old one.

dakotabenjamin:
@arunasank I'm made some changes according to your suggestions. 

I've somewhat tested the jobs using the [local circleci cli](https://circleci.com/docs/2.0/local-cli/) but it doesn't support CircleCI 2.1 features like orbs. It would be a good way to test deployments otherwise (@xamanu might be interested in delving deeper)

I'm still a little lost on implementing the "Create config file" step for the deployment so for now I suggest we put them in CircleCI's project Environment Variables for now. 


You will notice I implemented some "orbs" which are basically premade jobs that anyone can use. They are quite helpful and a very cool feature. We will have to add the AWS_DEFAULT_REGION variable to the CircleCI env variables as well.

.circleci/config.yaml
dakotabenjamin: It's not quite less messy but the changes I just pushed I think are a little better way of doing things.

.circleci/config.yaml
arunasank: Awesome. 👍

.circleci/config.yml
```diff
@@ -0,0 +1,156 @@
...
+            fi
+            # create new aws rds snapshot
+            aws rds create-db-snapshot --db-snapshot-identifier tm3-staging-deployment-snapshot-latest --db-instance-identifier tasking-manager-staging 
+      - run: 
+          name: Create config file
+          command: |
+            printf '{\"GitSha\":\"$CIRCLE_SHA1\", \"Environment\":\"staging\", \"DBSnapshot\":\"$DB_SNAPSHOT_STAGING\", \"MasterUsername\":\"$MASTER_USERNAME_STAGING\", \"MasterPassword\":\"$MASTER_PASSWORD_STAGING\", \"OSMConsumerKey\":\"$OSM_CONSUMER_KEY_STAGING\",\"OSMConsumerSecret\":\"$OSM_CONSUMER_SECRET_STAGING\", \"TaskingManagerSecret\":\"$TASKINGMANAGER_SECRET_STAGING\", \"TaskingManagerEnv\":\"$TASKINGMANAGER_ENV_STAGING\", \"TaskingManagerSMTPHost\":\"$TASKINGMANAGER_SMTP_HOST_STAGING\", \"TaskingManagerSMTPPassword\":\"$TASKINGMANAGER_SMTP_PASSWORD_STAGING\", \"TaskingManagerSMTPUser\":\"$TASKINGMANAGER_SMTP_USER_STAGING\", \"DatabaseSize\":\"$DATABASE_SIZE\",\"ELBSubnets\":\"$ELB_SUBNETS_STAGING\", \"SSLCertificateIdentifier\":\"$SSL_CERTIFICATE_ID_STAGING\",\"RDSUrl\":\"$RDS_URL_STAGING\"}' > $CIRCLE_WORKING_DIRECTORY/cfn-config-staging.json
+      - aws-s3/copy:
```
arunasank: Have you checked the contents of the file on S3? Does it look like a JSON string (`\"<string>\"`) or a JSON ("<string>"). We want the latter.

arunasank:
@dakotabenjamin the changes look great to me! Just had one minor comment above. In order to test, should I merge a test branch with `develop` to test the changes (You mentioned merging `cloudformation` with `develop`, which I am a little wary about)? Or have you already deployed the changes to test them?

.circleci/config.yml
dakotabenjamin: Those `\"` should be parsed by `printf` into regular `"`

dakotabenjamin:
So I need to create a `tasking-manager-staging` stack (the circleci script does not create stacks, only updates) but it's set up to deploy from the `cloudformation` branch.

Issue #1456, opened 2019-04-09
Change ASG UpdatePolicy, to cycle EC2 instances


arunasank:
Ref: https://github.com/hotosm/tasking-manager/issues/1395

We don't have an update policy configured on our TM cloudformation template at the moment, but let's configure one to:
* Cycle EC2 instances when the LaunchConfig is updated
* To maintain 100% health on the ASG - this means that the EC2 instances will be cycled without causing any downtime to the TM service.

More info on https://github.com/hotosm/tasking-manager/issues/1444#issuecomment-481085332

cc/ @dakotabenjamin

arunasank:
Starting work on this today.

arunasank:
We fixed this!

PR #1473, opened 2019-04-10
Cycle ASG instances in a sustainable way

arunasank:
Fixes https://github.com/hotosm/tasking-manager/issues/1456

This PR:
1. Cycles ASG instances when a change is made to the LaunchConfiguration
2. Maintains 100% healthy instances

cc/ @dakotabenjamin


arunasank:
I tested this on a new stack, and it works:

Screen Shot 2019-04-10 at 1 13 20 PM

cloudformation/tasking-manager.template.js
```diff
@@ -118,11 +124,14 @@ const Resources = {
...
           'echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf',
```
arunasank: @dakotabenjamin i know you really like the install all deps at once format, but it makes diffs to the script difficult to follow. Would you mind too much if we moved more to one command per line?

arunasank:
Don't worry about the tests failing, because this is a merge with the `cloudformation` branch. There's a comment there about fixing the failing test: https://github.com/hotosm/tasking-manager/pull/1332#commitcomment-32804493

Thanks for flagging @ramyaragupathy 🍏

cloudformation/tasking-manager.template.js
dakotabenjamin: Agreed

PR #1344, opened 2019-02-12
Project stats endpoint related

ramyaragupathy:
### /api/v1/stats/project/{project_id}

_[Sample payload](https://gist.github.com/ramyaragupathy/3e0b535170a97ddc4a3df04c83001f93)_

- [x] Total Mappers
- [x] Total Tasks
- [x] List all Changeset Comments
- [x] Percent of bad Imagery
- [x] Total Comments
- [x] Entities to Map
- [x] Due Date for project completion
- [x] Project Author/created by
- [x] Project Priority
- [x] mapperLevelEnforced
- [x] validatorLevelEnforced
- [x] Average mapping  time
- [x] Average validation time
- [x] Total mapping time
- [x] Total validation time
- [x] Total time spent on the project

### /api/v1/stats/project/{project_id}/activity

_[Sample Payload](https://gist.github.com/ramyaragupathy/4d91b1b64606e885ed2e1e2376d52ba1)_

- [x] Include a task id in activity stub

### /api/v1/stats/project/{project_id}/contributions

_[Sample Payload](https://gist.github.com/ramyaragupathy/f50e37ceaac49e0e23574bc5250635c0)_

For each user add:
- [x] Time spent mapping
- [x] Time spent validating
- [x] Total Time spent

PR #1477, opened 2019-04-11
Expand project activity query

ethan-nelson:
The DTO to handle task history included the history id and task_id
so those could be displayed on the project activity page. However,
those did not make it into the query that is used to populate the
DTO when the activity page is loaded.

### DIFF
--- a/server/services/stats_service.py
+++ b/server/services/stats_service.py
@@ -114,7 +114,8 @@ class StatsService:
     def get_latest_activity(project_id: int, page: int) -> ProjectActivityDTO:
         """ Gets all the activity on a project """
 
-        results = db.session.query(TaskHistory.action, TaskHistory.action_date, TaskHistory.action_text, User.username) \
+        results = db.session.query(TaskHistory.action, TaskHistory.action_date, TaskHistory.action_text,
+            TaskHistory.id, TaskHistory.task_id, User.username) \
             .join(User).filter(TaskHistory.project_id == project_id, TaskHistory.action != 'COMMENT')\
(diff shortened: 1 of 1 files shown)


server/services/stats_service.py
```diff
@@ -114,7 +114,8 @@ def _set_counters_after_invalidated(task_id: int, project: Project, user: User):
     def get_latest_activity(project_id: int, page: int) -> ProjectActivityDTO:
         """ Gets all the activity on a project """
 
-        results = db.session.query(TaskHistory.action, TaskHistory.action_date, TaskHistory.action_text, User.username) \
+        results = db.session.query(TaskHistory.action, TaskHistory.action_date, TaskHistory.action_text,
```
xamanu: Thanks, @ethan-nelson. Checking this PR, it seems has been fixed in the meanwhile by @ramyaragupathy and is in [current `develop`](https://github.com/hotosm/tasking-manager/blob/develop/server/services/stats_service.py#L128). coming from the PR #1344 (see [diff](https://github.com/hotosm/tasking-manager/pull/1344/files#diff-52db8babc6b474fbcba619e6b9a0cdfaR128)).

```
        results = db.session.query(
                TaskHistory.id, TaskHistory.task_id, TaskHistory.action, TaskHistory.action_date,
                TaskHistory.action_text, User.username
            ).join(User).filter(
                TaskHistory.project_id == project_id,
                TaskHistory.action != 'COMMENT'
            ).order_by(
                TaskHistory.action_date.desc()
            ).paginate(page, 10, True)
```

Please close the PR in case you can confirm.

PR #1480, opened 2019-04-11
Improvement of docker-compose

JorgeMartinezG:
This PR enables users to create instances of the tasking-manager project using docker. Users are able to execute both client and server tests within the container.

### DIFF
--- a/devops/docker/tasking-manager/Dockerfile
+++ b/devops/docker/tasking-manager/Dockerfile
@@ -1,31 +1,37 @@
-# Base on Python on Debian stable
 FROM python:3-stretch
 
-EXPOSE 5000
+RUN mkdir -p /usr/src/app
+WORKDIR /usr/src/app
 
-# Install dependencies
+# Install dependencies for shapely
 RUN apt-get update \
   && apt-get upgrade -y \
   && apt-get install -y libgeos-dev \
   && apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
-RUN curl -sL https://deb.nodesource.com/setup_9.x | bash - \ 
+# Upgrade pip
+RUN pip install --upgrade pip
+
+ARG branch=develop
+RUN git clone --depth=1 git://github.com/hotosm/tasking-manager.git \
+	--branch $branch /usr/src/app
+
+## API/SERVER CONFIGURATION
+
+# Add and install Python modules
+RUN pip install --no-cache-dir -r requirements.txt
+
+## CLIENT CONFIGURATION
+
+# Install nodejs.
+RUN curl -sL https://deb.nodesource.com/setup_9.x | bash - \
   && apt-get install -y nodejs
 
-RUN npm i gulp-cli -g && npm link gulp
+RUN npm install -g gulp-cli karma karma-jasmine karma-chrome-launcher
 
-# Add code base of the Tasking Manager
-ARG defined_branch=develop
-RUN git clone --single-branch --branch $defined_branch https://github.com/hotosm/tasking-manager /usr/src/app
-WORKDIR /usr/src/app
+RUN cd client && npm link gulp && \
+	npm install closure-util --save && \
+	npm install && gulp build
 
-# Add and install Python modules for back-end
-RUN pip install -r requirements.txt
-
-# Build front-end
-RUN cd client && \
-  npm install browser-sync closure-util openlayers --save && \
-  npm install && gulp build
-
-# Serve application (be aware this runs on port 5000)
-CMD python manage.py runserver -h 0.0.0.0
+EXPOSE 5000
+CMD ["python", "manage.py", "runserver", "-h", "0.0.0.0"]
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -1,0 +1,18 @@
+# To use this file, run `docker-compose up`.
+version: '3'
+
+services:
+
+  # Main application
+  app:
+    build: ./devops/docker/tasking-manager
+    volumes:
+      - .:/usr/src/app
+    ports:
+      - "3000:3000"
+
+  # Migration service
+  migration:
+    build: ./devops/docker/tasking-manager
+    volumes:
+      - .:/usr/src/app
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,11 +2,13 @@
 version: '3'
 
 services:
-
   # Main application
   app:
-    build: devops/docker/tasking-manager
+    image: hotosm/tasking-manager
+    container_name: app
     env_file: tasking-manager.env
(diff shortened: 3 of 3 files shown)


xamanu:
Thanks, @JorgeMartinezG!

We habe had some issues with our deployment when changing the `Dockerfile` in the repository's root. Which let us move it, for now, the respective Dockerfile into the `devops/docker/tasking-manager`. Unfortunately our CI would only fail after merging to `develop` cause only this will run tests on Elastic Beanstalk and show the problems. Once the move to Cloudformation has been done, the main Dockerfile can then be replaced, because no Elastic Beanstalk building will happen anymore.

What do you think? Can we apply for now your improvements to the `devops/docker/tasking-manager/Dockerfike`? Shall we wait with this PR, until we moved on with our deployment?

JorgeMartinezG:
@xamanu  Sure. I will move the file to ./devops/docker/tasking-manager and modify the docker-compose. Do you think that I should keep the PR with a single commit? Or include multiple commits within a PR?

xamanu:
I guess one commit is best for such a change.

devops/docker/tasking-manager/Dockerfile
```diff
@@ -1,31 +1,44 @@
...
 
-# Install dependencies
+# Install dependencies for shapely
 RUN apt-get update \
   && apt-get upgrade -y \
   && apt-get install -y libgeos-dev \
   && apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
-RUN curl -sL https://deb.nodesource.com/setup_9.x | bash - \ 
+RUN mkdir -p /usr/src/app
```
xamanu: For better readability, I suggest to move this line and the next on top of the `## SERVER CONFIGURATION` block.

devops/docker/tasking-manager/Dockerfile
```diff
@@ -1,31 +1,44 @@
...
+WORKDIR /usr/src/app
+
+# Upgrade pip
+RUN pip install --upgrade pip
+
+ARG branch=develop
```
xamanu: Same here and next two lines. I think this is more general and doesn't fit under `## SERVER CONFIGURATION`

devops/docker/tasking-manager/Dockerfile
```diff
@@ -1,31 +1,44 @@
...
-# Add and install Python modules for back-end
-RUN pip install -r requirements.txt
+Run gulp build
 
-# Build front-end
-RUN cd client && \
-  npm install browser-sync closure-util openlayers --save && \
-  npm install && gulp build
+## Command configuration
+WORKDIR /usr/src/app/
```
xamanu: Duplicate with line 13.

docker-compose.yml
```diff
@@ -2,47 +2,45 @@
...
+    image: hotosm/tasking-manager
+    container_name: app
     env_file: tasking-manager.env
+    environment:
+      - POSTGRES_ENDPOINT=postgresql
     depends_on:
       - postgresql
     ports:
       - "5000:5000"
+      - "3000:3000"
```
xamanu: Why `3000` instead of `5000`?

devops/docker/tasking-manager/Dockerfile
JorgeMartinezG: Check

docker-compose.yml
JorgeMartinezG: Port 3000 should be included with the purpose of client development using gulp. I moved it to the override file.

devops/docker/tasking-manager/Dockerfile
JorgeMartinezG: Check

devops/docker/tasking-manager/Dockerfile
JorgeMartinezG: Check

devops/docker/tasking-manager/Dockerfile
```diff
@@ -1,31 +1,44 @@
...
 
-# Add code base of the Tasking Manager
-ARG defined_branch=develop
-RUN git clone --single-branch --branch $defined_branch https://github.com/hotosm/tasking-manager /usr/src/app
-WORKDIR /usr/src/app
+RUN npm install closure-util --save && npm install
 
-# Add and install Python modules for back-end
-RUN pip install -r requirements.txt
+Run gulp build
```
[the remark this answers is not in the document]
JorgeMartinezG: Check

devops/docker/tasking-manager/Dockerfile
```diff
@@ -1,31 +1,44 @@
-# Base on Python on Debian stable
 FROM python:3-stretch
+MAINTAINER Tasking Manager development team
```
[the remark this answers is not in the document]
JorgeMartinezG: Check

JorgeMartinezG:
@xamanu Applied changes and rebased into single commit. Let me know your opinions

docker-compose.yml
xamanu: Cool. Can we use `3000` as the one we expose from the Dockerfile by default? Or is there a reason we want to use `5000`?

xamanu:
Thanks, @JorgeMartinezG! I think we are almost there. Lets go quickly into the two open questions regarding the best port to use and whether we can avoid the double building.

docker-compose.yml
JorgeMartinezG: **EXPOSE** will enable communication between containers that belong to different docker networks. I think you are referring to establish communication between the container and the host machine. For this case, **PUBLISH** will be the command. IMO, publishing the 3000 for development would be ok when creating the docker instances. For production, it would involve security issues. Let me know your opinion.

docker-compose.override.yml
```diff
@@ -0,0 +1,16 @@
...
+
+  # Main application
+  app:
+    build: ./devops/docker/tasking-manager
+    volumes:
+      - .:/usr/src/app
+
+  # Migration service
+  migration:
+    build: ./devops/docker/tasking-manager
```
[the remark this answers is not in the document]
xamanu: ok, shall we factor this out into a separate issue then?

docker-compose.yml
xamanu: I just want to be consistent. And using `5000` at some places and `3000` on others will most likely cause problems and confusion. Especially for new ones. Is there a reason we want to use both?

xamanu: Never mind, I was still looking at yesterdays PR proposal. The new changes make sense of course.

docker-compose.override.yml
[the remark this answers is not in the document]
xamanu: Created an issue for that: https://github.com/hotosm/tasking-manager/issues/1496

PR #1481, opened 2019-04-12
Fixes #602 - Making fields in metadata panel required

### LINKED ISSUE #602: Project Creation - Metadata fields mandatory
Based on stakeholder feedback, all of the fields in the meta data panel, with the exception of campaign, should be mandatory. We need those fields to be used so we get the maximum benefit from them.

fitoria:
This fixes issue #602

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,9 @@ client/node_modules/
 client/assets/styles/css
 client/app/taskingmanager.config.js
 client/package-lock.json
+client/yarn.lock
+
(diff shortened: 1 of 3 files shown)


xamanu:
Thanks @fitoria. It works for the mapping types, but unfortunately I get the error `An organisation tag is missing`, even an organization is specified. Some debugging needed here.

image

ethan-nelson:
Yeah, this may need some special handling due to the `<tags-input>` field.

PR #1486, opened 2019-04-13
Fix posting message when commenting

xamanu:
When posting a comment on a project an error message is presented that the comment was not added.

image

The log file presented:

```
CRITICAL: Chat PUT - unhandled error: name 'tm' is not defined [in /home/user/Code/HOT/tasking-manager/server/api/messaging/project_chat_apis.py:71]
```

This PR includes a necessary dependency declaration in order to fix this error.


russdeffner:
oops, I filed #1516 before seeing this - should be fixed with the deployment

xamanu:
Proposing to bring this as a hotfix into a minor release: #1520

PR #1500, opened 2019-04-19
Revert circleci version to 2

dakotabenjamin:
Note I added a a DB name to the RDS resource.


dakotabenjamin:
It's failing on cache restoration- something you can't test locally for. Probably something to do with the change in base image which the user isn't root anymore. Thought I fixed that by setting the folder permissions.

PR #1507, opened 2019-04-22
Reset bad imagery counter

### LINKED ISSUE #1505: Reset bad imagery count on task reset
From https://github.com/hotosm/tasking-manager/issues/1470#issuecomment-485311160 - On reseting task status, `tasks_mapped` and `tasks_validated` counts are set to zero. Similarly `tasks_bad_imagery` should be initialised to zero, otherwise it will lead to incorrect escalated numbers on bad imagery count.

ramyaragupathy:
Fixes #1505 - Updated code to reset bad imagery counter in `projects` table on task reset


ethan-nelson:
Good catch!

Issue #1506, opened 2019-04-22
Improve project deletion behaviour


ramyaragupathy:
From https://github.com/hotosm/tasking-manager/issues/1470#issuecomment-485311160 - when a project manager tries to delete a project that has some contributions(task history, project discussions) then we show a message informing the user about contributions.

![](https://camo.githubusercontent.com/82b9302195888ad75c545c49dc4eb1a79416c9a8/68747470733a2f2f70617065722d6174746163686d656e74732e64726f70626f782e636f6d2f735f414445414635393235413042383344464331313446394430313444414139423142313745313646463144423035313144444133443731423430303833394633445f313535353330373738343232365f696d6167652e706e67).

From @russdeffner's [comment](https://github.com/hotosm/tasking-manager/issues/1460#issuecomment-483125894)
> this is true; however you can ‘trick’ the TM into thinking no mapping was done by using the unmark/reset tasks to ready state buttons in edit mode and then delete the project

naturally project managers will try task reset and then do a project deletion. However if there has been project chat in the discussion tab, then the deletion wouldn't still get through, since chat is counted as a contribution too. That seems right, however our frontend error message for deletion should highlight what other contributions are remaining yet, rather than just a blanket word of contributions. It is sort of misleading:

- after the task reset or 
- for a project that has zero task history but an active discussion thread

ramyaragupathy:
hey @wille @ethan-nelson - python question, I am looking at how `db.session` handles deletion of projects. 
- Project ID is referred in multiple other tables (Tasks, task history, task invalidation history, project info, project chat). 
- When there is no entry in the project chat table - deletion gets through fine. If there are chat records for the project id then the deletion fails with `foreign key constraint` error. 

```
{
    "error": "Project GET - unhandled error: (psycopg2.IntegrityError) update or delete on table \"projects\" violates foreign key constraint \"project_chat_project_id_fkey\" on table \"project_chat\"\nDETAIL:  Key (id)=(5496) is still referenced from table \"project_chat\".\n [SQL: 'DELETE FROM projects WHERE projects.id = %(id)s'] [parameters: {'id': 5496}]"
}
```
- This is understandable since the project id is referred to in `project_chat` table. Going by this logic, this should also throw errors for `tasks` and `task_history` table where the project id is referred as foreign key right? How does db.session handles foreign key constraint in task related tables, but doesn’t do so for `project_chat` table?

ramyaragupathy:
[...]

PR #1508, opened 2019-04-22
Define cascade relation for project chat

ramyaragupathy:
Fixes https://github.com/hotosm/tasking-manager/issues/1506 - Now project deletion will go through on task reset irrespective of whether there is discussion/comment within project. This is handled by defining a cascading relationship for `project_chat` table similar to how `project_info` and other tables relations are stated in the code.

**How to test?**
- Pick a test project that has mapping and discussions/comments
- Check deletion
- Deletion should throw an error message about contributions
- Now reset all tasks
- Try deletion again and this time it should get through without any error


JorgeMartinezG:
Tested locally in my local setup. Created an arbitrary project and followed the steps defined in the pull request getting expected results

Screen Shot 2019-04-25 at 3 12 18 PM

PR #1512, opened 2019-04-23
Minor changes to the test database

xamanu:


arunasank:
Testing this out currently. 🏃‍♀️

willemarcel:
@xamanu maybe it would be better if we import that dump and export it with the ` --no-owner` option on the `pg_dump` command, so it will not require to set a user or superuser to restore the dump correctly.

willemarcel:
@xamanu @arunasank I pushed the change I mentioned in the last comment to https://github.com/hotosm/tasking-manager/pull/1513

xamanu:
Cool. Closing here.

PR #1513, opened 2019-04-23
No owner on test SQL dump

willemarcel:
Replaces https://github.com/hotosm/tasking-manager/pull/1512


arunasank:
This works. I loaded this script on tasks-assisted.hotosm.org, and the dump loaded perfectly. 👍

PR #1517, opened 2019-04-24
Bugfix/bugs found

fitoria:
Fixes a bug on TaskInvalidationHistory.record_invalidation and on projectMapService


xamanu:
@fitoria Can you give some instructions how this bug can be reproduced and the behaviour of this code tested?

Issue #1491, opened 2019-04-16
Getting errors when executing python tests in local instance


JorgeMartinezG:
Running the command ```python -m unittest discover tests/server``` on a fresh installation of the project, I receive this 4 errors. I also include the Traceback for each error for better debugging:

```
======================================================================
ERROR: test_gpx_xml_file_generated_correctly (integration.services.test_mapping_service.TestMappingService)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/unittest/mock.py", line 1204, in patched
    return func(*args, **keywargs)
  File "/usr/src/app/tests/server/integration/services/test_mapping_service.py", line 54, in test_gpx_xml_file_generated_correctly
    gpx_xml = MappingService.generate_gpx(1, '1,2', timestamp)
  File "/usr/src/app/server/services/mapping_service.py", line 182, in generate_gpx
    if not tasks or tasks.count() == 0:
TypeError: count() takes exactly one argument (0 given)

======================================================================
ERROR: test_osm_xml_file_generated_correctly (integration.services.test_mapping_service.TestMappingService)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/unittest/mock.py", line 1204, in patched
    return func(*args, **keywargs)
  File "/usr/src/app/tests/server/integration/services/test_mapping_service.py", line 93, in test_osm_xml_file_generated_correctly
    osm_xml = MappingService.generate_osm_xml(1, '1,2')
  File "/usr/src/app/server/services/mapping_service.py", line 213, in generate_osm_xml
    if not tasks or tasks.count() == 0:
TypeError: count() takes exactly one argument (0 given)

======================================================================
ERROR: test_invalidate_all_sets_counters_correctly (integration.services.test_validation_service.TestValidationService)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/usr/src/app/tests/server/integration/services/test_validation_service.py", line 56, in test_invalidate_all_sets_counters_correctly
    ValidatorService.invalidate_all_tasks(self.test_project.id, self.test_user.id)
  File "/usr/src/app/server/services/validator_service.py", line 247, in invalidate_all_tasks
    task.unlock_task(user_id, new_state=TaskStatus.INVALIDATED)
  File "/usr/src/app/server/models/postgis/task.py", line 535, in unlock_task
[...]

PR #1519, opened 2019-04-25
Multiple fixes to python local tests

JorgeMartinezG:
This PR includes implementations and test modifications with the purpose of running tests within a local setup. Each of the changes are documented and explained on issue #1491 

**Note.** The assigned reviewer **must** verify this PR using their own setup of Tasking Manager or with docker.

```python -m unittest discover tests/server/```

### DIFF
--- a/server/config.py
+++ b/server/config.py
@@ -7,7 +7,7 @@ class EnvironmentConfig:
     """ Most settings can be defined through environment variables. """
 
     # Load configuration from file
-    load_dotenv(os.path.join(os.path.dirname(__file__), 'tasking-manager.env'))
+    load_dotenv(os.path.normpath(os.path.join(os.path.dirname(__file__), '..', 'tasking-manager.env')))
 
     # The base url the application is reachable
     APP_BASE_URL = os.getenv('TM_APP_BASE_URL', 'http://127.0.0.1:5000')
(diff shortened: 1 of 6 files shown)


xamanu:
Thanks @JorgeMartinezG! I went over code and tested it. I worked well with the Docker setup, However for it to run the tests properly I had to fix in the `config.py ` the configuration path in line 10. I added a commit to it and pushed here. Feel free to squash commit messages on merge.

JorgeMartinezG:
@xamanu history has been rebased in order to keep a single commit.

ramyaragupathy:
Per chat w/ @JorgeMartinezG - this might be due to version mismatch in postgis in both our machines. I want to verify the version information from our server and then set up my test environment accordingly. Working w/ @arunasank to get this info from our server boxes.

ramyaragupathy:
Ran this w/ docker setup to replicate server env. Noting a couple of issues - 
- one related to env variables and how it defaults to specific values that interrupts with the DB connection. Ticketing this separately after confirming w/ @xamanu and @JorgeMartinezG 
- second the DB upgrade did not get through well, that is it never ended even after an hour or so.
Again need to double check on a different env

ramyaragupathy:
> Tested and works fine. Ran into similar issues with the postgis version, as @ramyaragupathy, but I think these should not stop this from going in.

Agree @xamanu - As such I don't see any issues in the code flow and any docker related issues I will branch out discussion in a new ticket. We're good to hit 🍏

PR #1535, opened 2019-05-02
Only recycle snapshots, but do not use them during CloudFormation updates

arunasank:
cc/ @dakotabenjamin


dakotabenjamin:
I have to fix something with the temp snapshots not deleting and then I'll test this but it looks good!

PR #1537, opened 2019-05-02
Remove redundant code and fix temp snapshot deletion

dakotabenjamin:
This fixes points 2 and 4 in https://github.com/hotosm/tasking-manager/issues/1444#issuecomment-488684858

cc @arunasank 

tested via local circleci execution

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -131,24 +131,19 @@ jobs:
           # create new aws rds snapshot
           aws rds create-db-snapshot --db-snapshot-identifier tm3-staging-deployment-snapshot-latest --db-instance-identifier $RDS_ID
           sleep 2m
-    - run:
-        name: Create config file
-        command: |
-          touch $CIRCLE_WORKING_DIRECTORY/cfn-config-staging.json
-          echo $JSON_CONFIG > $CIRCLE_WORKING_DIRECTORY/cfn-config-staging.json
-    - run:
-        name: S3 Copy $CIRCLE_WORKING_DIRECTORY/cfn-config-staging.json -> s3://hot-cfn-config/tasking-manager/tasking-manager-staging-us-east-1.cfn.json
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -209,24 +204,19 @@ jobs:
...
-        name: S3 Copy $CIRCLE_WORKING_DIRECTORY/cfn-config-production.json -> s3://hot-cfn-config/tasking-manager/tasking-manager-production-us-east-1.cfn.json
-        command: aws s3 cp $CIRCLE_WORKING_DIRECTORY/cfn-config-production.json s3://hot-cfn-config/tasking-manager/tasking-manager-production-us-east-1.cfn.json
     - deploy:
         name: Deploy to production
         command: echo $JSON_CONFIG & cfn-config update production $CIRCLE_WORKING_DIRECTORY/cloudformation/tasking-manager.template.js -f -c hot-cfn-config -t hot-cfn-config -r $AWS_REGION -p "$JSON_CONFIG"
     - run:
         name: Cleanup
+        when: always
         command: |
           DESCRIBE_SNAPSHOT=`aws rds describe-db-snapshots --query "DBSnapshots[? DBSnapshotIdentifier=='tm3-production-deployment-snapshot-temp']" --output text`
```
arunasank: Unrelated to the fix, and more of a readability preference - Can we use the simpler version of the command like `aws rds describe-db-snapshots ----db-snapshot-identifier tm3-production-deployment-snapshot-temp` instead of the query format we are using here?

dakotabenjamin: the command produces an error if a snapshot isn't found, which was causing the whole script to fail

arunasank: @dakotabenjamin ah, gotcha. I tested to see that it works if we also pass the `db-instance-identifier`. Your call though.

Issue #1467, opened 2019-04-10
Reduce instance bootstrapping time


arunasank:
Currently, we are using an `m3.medium`, and instance takes ~20 minutes to bootstrap. Let's bring this down to < 5m. This will look like testing on instances with better network, since the biggest time sink is retrieving and installing packages from the OS package manager.

cc/ @dakotabenjamin

dakotabenjamin:
Can we build a custom AMI with the packages already installed?

arunasank:
Yes, but it's a bit more complex to build, but also maintain. Let's maybe look at it at the end, after we exhaust all the simple options? 😅

arunasank:
Ref: https://github.com/hotosm/tasking-manager/issues/1395

arunasank:
Moved to a `c5d.large`, and the time is down to 6.667 minutes. I don't think we can go below this, unless we want to spend $$$. However, this machine is bigger than the `m3.medium`. I want to load test it to see how many users it can support, and see if this delay is in fact acceptable.

xamanu:
I'm not very much aware of the AWS scaling options and prices. 

In general, I think I would like to explore the of building an own AMI. Is it that hard? Can we reuse other efforts? Just looking around quickly, something like [packer](https://www.packer.io/) could be interesting, possibly better to provisioned by some well-used solution (i personally like and used Ansible in the past).

I would like to have a reproducible setup. This would make it testable and speed up our (deployment) processes.

arunasank:
I built a custom AMI using Packer. This reduces the instance bootstrapping time to ~2m48s. is this good enough? Current bottlenecks to making this faster:
* The AMI installs system level dependencies, but TM dependencies are installed during the instance bootstrap, since the state of the `requirements.txt` to be used for installing python dependencies and the state of the `package.json` to install the node dependencies are based on the `GitSha`, which is only available at the time the instance is bootstrapped.

**Solution**
* To build a fresh AMI as part of CI, whenever a merge is made to `develop` or `master`, which uses the `HEAD` commit in the `develop` or `master` branch to install TM dependencies. For all other branches, we will use a common AMI that configured TM dependencies during bootstrap. 

Thoughts @xamanu @smit1678 @dakotabenjamin ?

xamanu:
For production - live - `master`, I think, it is worth to build an AMI through the CI. This is what we mostly going to use when there is a need for scaling.

[...]

PR #1546, opened 2019-05-07
Create an AMI to reduce bootstrapping time

arunasank:
Fixes https://github.com/hotosm/tasking-manager/issues/1467 when done.

Copying over the last comment from there:

> I built a custom AMI using Packer. This reduces the instance bootstrapping time to ~2m48s. is this good enough? Current bottlenecks to making this faster:
> * The AMI installs system level dependencies, but TM dependencies are installed during the instance bootstrap, since the state of the `requirements.txt` to be used for installing python dependencies and the state of the `package.json` to install the node dependencies are based on the `GitSha`, which is only available at the time the instance is bootstrapped.
>
> **Solution**
> * To build a fresh AMI as part of CI, whenever a merge is made to `develop` or `master`, which uses the `HEAD` commit in the `develop` or `master` branch to install TM dependencies. For all other branches, we will use a common AMI that configured TM dependencies during bootstrap. 
> 
> Thoughts @xamanu @smit1678 @dakotabenjamin ?


arunasank:
👋 @xamanu and @dakotabenjamin. The things remaining to be done before this PR gets merged are:
* * Migrate more things from the `UserData` script on the CloudFormation template, and use it in the packr AMI script. That is, we will move all the lines of the script until the migration of the environment variables to the `packr` script.
* Make sure that Circle CI is set up to run packr to generate an AMI, and then use it in the CloudFormation template on **merges with `master` alone.**

xamanu:
@eternaltyro, @dakotabenjamin, can this be closed? Or is this still relevant?

PR #1487, opened 2019-04-13
Restrict editor programs for mapping and validation stages

xamanu:
Rerolling the PR #1421 of @zlavergne on a local branch (to ease up collaboration) and rebased on latest `develop` resolving some minor conflicts.

> Restrict editor programs for mapping and validation stages
> 
> >**This PR includes a db migration**
> 
> Related Issue: #893
> 
> * Added list of editors in the `Settings` menu of a project-edit page where a project manager can select permitted editors for each editing stage 
> Screen Shot 2019-03-21 at 10 37 50 AM
> 
> * Updated the editor dropdown for to populate only with the permitted editors
> Screen Shot 2019-03-25 at 12 28 35 PM

PR #1547, opened 2019-05-07
Fix old content migrations' database state

xamanu:
PR together with @ramyaragupathy.

In some old migrations we spotted some lines that were causing errors today, like on #1487:

`sqlalchemy.exc.ProgrammingError: (psycopg2.ProgrammingError) column projects.mapping_editors does not exist`

This was caused by running content migrations on the database in these files, but not looking at the respective state of the database (schema) but rather pulling in the latest state from code with:

```
projects = Project.__table__	
tasks = Task.__table__
messages = Message.__table__
``` 

This caused errors for missing columns that have been created in later migrations. We replaced it with direct queries to the database which will return in result sets with the database schema as it is at the moment of running the migration.


xamanu:
There is a follow up issue, which you might want to test together #1548.

xamanu:
Cool. Please squash your commits, too. Thanks.

ethan-nelson:
Can't run the tests locally, but it makes sense to me: we shouldn't rely on the structure presently defined in the models file, but rather on the structure in the database prior to the migration.

migrations/versions/0a6b82b55983_.py
```diff
@@ -78,9 +78,15 @@ def upgrade():
 
             # Only process messages from projects that still exist
             if project_existence[project_id]:
-                op.execute(
-                    messages.update().where(messages.c.id == message.id)
-                        .values(message_type=message_type, project_id=project_id, task_id=task_id))
+                if task_id is None:
```
nrotstan: This check and change to 'null' is also needed for message_type and project_id.

ramyaragupathy:
Thanks for the review @nrotstan. I noticed this failing on another branch and was digging into what was going wrong. This helps!

xamanu:
Tested again today with @ramyaragupathy. And moving in.

PR #1548, opened 2019-05-07
Fix missing migrations

xamanu:
Lately it was a bit uncomfortable to produce migrations as there were some lines around the `idx_project_info composite` table, which can be fixed by avoiding the white space in the name and renaming the index table to `idx_project_info_composite`. In addition this PR catches up on other migrations that doesn't seem to be run before.


xamanu:
Included a check for the `textsearch_idx` index. Thanks for spotting @ramyaragupathy!

xamanu:
Cool. This is ready. I'm just waiting for #1547, for merging in, as the migrations are pointing to the last migration of #1547.

PR #1559, opened 2019-05-09
Remove unused Dockerfiles

xamanu:
Some cleanup here.

This Dockerfile was part of the EC2 setup but won't be needed for the Cloudformation setup anymore. The general Dockerfile is located in `devops/docker/tasking-manager/Dockerfile`.


xamanu:
Thanks, @arunasank! Porbably we can clean up the `win` folder and files in it. However I suggest doing this in a separate PR.

PR #1566, opened 2019-05-10
Optimize gulp watch 

fitoria:
Optimize gulp watch to only track files inside our project and not the whole node_modules dir, it uses less CPU this way.


JorgeMartinezG:
@fitoria does this happen with specific operating system?

Issue #1572, opened 2019-05-14
A few bugs on develop 


zlavergne:
I've found a few bugs in `develop` that may block the upcoming release:

Backwards compatibility: #1313 [(comment explaining bug)](https://github.com/hotosm/tasking-manager/pull/1313#issuecomment-492356639) 

TasksAsGPX API broken: #1519 [(comment about bug)](https://github.com/hotosm/tasking-manager/pull/1519#commitcomment-33535346)

ramyaragupathy:
Thanks for noticing and fixing these bugs @zlavergne 🙇‍♀ I have tried downgrade through `python manage.py db downgrade` when I was working on this and I didn't face any issue. Will you please share steps to reproduce this issue?

zlavergne:
Steps 2 reproduce:
1. upgrade db to `a8a7537985c0`
2. Edit a project
3. Try to save and you'll get an error
* console shows the response is `"{'allowNonBeginners': ConversionError(\"This field is required.\")}"`
* the request body shows that `allowNonBeginners: null`

If you go to Permissions and check then uncheck "Require experience level INTERMEDIATE or ADVANCED" it forces `vm.allowNonBeginners` to `true` then `false` which makes it not null and then you can save.

xamanu:
Many thanks @zlavergne for spotting and fixing the bugs.

PR #1573, opened 2019-05-14
Fixed the following bugs from #1572

zlavergne:
* backwards compatibility
* broken TasksAsGPX API


zlavergne:
That's weird. I did the same thing on `develop` and got the same error. I wonder where the difference could be??

My SQLAlchemy modules:
```
Flask-SQLAlchemy@2.2
SQLAlchemy@1.3.0
```

Issue #1504, opened 2019-04-22
Restrict project edit function to project author


ramyaragupathy:
From https://github.com/hotosm/tasking-manager/issues/1470#issuecomment-485311160, at present we allow a project to be deleted by anyone who has a project manager role. Even if someone has not authored a project, they could edit/delete it, provided they have a project manager status. 

I feel this should be restricted to just one project manager who is the project author.

russdeffner:
I would agree, as long as Admins could still edit/delete any project and that more than one project manager can be assigned. We often have a need in disaster response for a team to edit projects while original creator is sleeping/away

xamanu:
Yes, in #1591 there is planned to have admins of the same organization that can edit the projects and special teams can be created that can be assigned with the project management role. 

In this issue here we will immediately fix that all project managers can edit and restrict it to the author.

xamanu:
This has been part of release 3.3.2. Closing here. Thanks.

PR #1574, opened 2019-05-14
[WIP] Restrict project edit only to author and admin

JorgeMartinezG:
This pull request is a fix for #1504 . For now the restriction is done in the client side. Projects are limited to authors and admins to edit or delete actions.

I want to iterate through this Pull request and know your opinions @ethan-nelson @xamanu. Would be necessary to make the restriction also on the server side also?


ethan-nelson:
If we are going to restrict editing, I would say one definite need would be the ability to delegate to or add other project managers the edit capability. Some activations rely on teams of PMs (that may not be admins).

xamanu:
Yesterday, due to emergency (#1687) a partial advancement of this has been applied. Please review this PR to build on top of the lastest changes again.

Issue #1264, opened 2018-10-30
Machine learning additions to TM


wronk:
## Vision for updates
Development Seed is working with CosmiQ Works and @smit1678 to bring some new machine learning tools to the TM interface. We want to deploy ML within TM inform mappers about the project they're working on and boost their efficiency. We're starting with estimating total building footprint per task square, and you can see version 0 in our [blog post](https://medium.com/devseed/weaving-automation-into-the-mapping-workflow-adding-ai-to-the-tasking-manager-52172ed2ece8) on the topic. Basically, it takes an ML estimate of building footprint (in sq. meters), building area from OSM, and displays the difference using color coded task squares.

## Goals
Our main focus here, is to make a model agnostic pipeline for deploying ML. You should be able to use any type of per-tile ML model (e.g., road length estimator, green area estimator, etc.) on a task-by-task basis and display that information in the TM. We expect that mappers will be faster and have a better experience within TM when they roughly know the difficulty of each task and can see at a glance how much remaining infrastructure in their project is missing from OSM.

**In this round of work, we will focus on:**
1. Initiating and running an ML model on the cloud (AWS Lambda) from TM's backend to calculate per-tile statistics
1. Storing and accessing per-task predictions on TM's backed as well as visualizing this information on the frontend

**Future work will focus on:**
1. Updating the project creation interface so that project owners can run this ML pipeline themselves
2. Expanding to other infrastructure estimates (e.g., roads)

## Design and suggestions
**We want to hear from the community on plans concerning the project visualization.** Specifically, what do you think is the best way to display information like an estimate of missing building area? Again, see the blog post demo ☝️for our first iteration. We have some ideas for the next design iteration, but we'd like to hear from mappers on their preferences.

**Rough design specs:**
1. We'll need to display a gradient of information. With missing buildings for example, that might be (1) no missing building area (2) <1000 sq. m. of missing building area and (3) >= 1000 sq. m. of building area.
1. If possible, we should incorporate this information into the default screen rather than create a new toggle-able layer like we did in the linked demo.
1. Avoid adding new colors -- there are already 7 states with 6 different colors.
1. Obviously, be as intuitive as possible -- the more complex this is, the harder it is to explain to new mappers.

smit1678:
> If possible, we should incorporate this information into the default screen rather than create a new toggle-able layer like we did in the linked demo.

[...]

PR #1577, opened 2019-05-15
Task Annotation Backend

fitoria:
This is the backend work from PR #1349 as work in progress to solve issue #1264


JorgeMartinezG:
@fitoria could you share a workflow for testing? I tested this branch on local server and made POST and GET requests to `ApplicationAPI` endpoint.

xamanu:
I think this has only an API, and there is no way to test further (correct me @fitoria if I'm wrong). As far as the API works well, and you agree with the code and way of implementation it is all good.

Issue #1580, opened 2019-05-15
Prepare Task Annotations and complexity assisted ml-layer


xamanu:
We want to bring up the work done so far for the machine learning integration (#1264). First steps:

* [x] Rebase on latest `develop` and resolve minor conflicts for PR #1349
* [x] Separate backend and front-end code into different branches. Squash the commits.
* [x] Create a PR for the backend only (#1577)
* [x] Create a PR for the frontend example code, which then can be deployed on the `tasks-assisted` testing instance.
* [x]  Create the example project on the `tasks-assisted` instance.
* [x] Fix some bugs

PR #1582, opened 2019-05-16
Complexity measures showcase

fitoria:
this is a PR for the front and backend work for task annotations to be deployed on tasks-assisted testing instance, includes two minor bug fixes part of issue #1580


xamanu:
Two things are missing:

* Text representation of the calculation (x buildings on OSM; y buildings detected; ...). I saw this one before on some other testing instances.
* The legend on the map is jumping to the side then selecting a task, while the ml layer is enabled.

xamanu:
Changed the title to `frontend` only, as this is the main goal for this PR. Of course the PR is based on the backend PR #1577

xamanu:
This PR is superseded by #1673

PR #1595, opened 2019-05-24
Modify CloudFormation to use spot instances

arunasank:
Modified the CloudFormation template to use spot instances, using the `MixedInstancesPolicy` in the `AutoScalingGroup` resource. In this scenario, the stack will always attempt to use spot instances from a pre-defined set of instance pools, and only use on-demand instances when there is no available spot capacity.

cc/ @dakotabenjamin @smit1678 @xamanu


arunasank:
## TODO

* Benchmark the two new instances added here, to see that they perform at the same level as the `c5d.large`.

Issue #1589, opened 2019-05-21
Enable different matomo sites to be used based on config


dakotabenjamin:
In order for tasks-assisted and teachosm to be tracked via matomo without diverging code, I am trying to make the tracker configurable for multiple site IDs. 

The matomo code is found in [client/index.html](https://github.com/hotosm/tasking-manager/blob/1cea6188b26c95f9c233f3e9f415a12f403eafa2/client/index.html#L280-L284):
```
<!-- Matomo -->
<div id="optout-form"></div>
<script type="application/javascript">var site_id="9"</script>
<script type="application/javascript" src="https://cdn.hotosm.org/tracking-v2.js"></script>
 <!-- End Matomo Code -->
```

What I want to do is have an environment variable in `tasking-manager.env` that can set the `site_id` in that line above. I do not know if the client configuration can work in the same way the server configuration does, however, and could use some advice.

xamanu:
This might be related to https://github.com/hotosm/tasking-manager/issues/1383, as there is a need to also introduce variables from the `tasking-manager.env` to have effects on the frontend client. @willemarcel, do you have a suggestion?

dakotabenjamin:
Yes, I agree- in fact the whole matomo code should be optional for those who do not need it

xamanu:
I'm wondering @dakotabenjamin, whether this is critical priority, because this needs to be fixed before the next release?!

dakotabenjamin:
reiterating from chat, but no it's not a priority. 

I think I have a solution, which is to have CircleCI overwrite `client/taskingmanager.config.json` to include variables set in the CircleCI project. 

1. Add the matomo site id as a variable in circleci
2. in the `devops/cloudformation/tasking-manager.template.js` file, edit the UserData value in the LaunchConfiguration to install `jq` and then edit the config file to include the site id

xamanu:
I would like to add here:

* Configurable matomo endpoint (move the hard coded url into the configuration file)
* if no endpoint is provided hide the tracking script entirely.

Do you think we can handle this here, or should it be a follow up issue, @dakotabenjamin?

dakotabenjamin:
We can handle this here, I'll add it to the PR

dakotabenjamin:
This needs to be reimplemented in the redesign

PR #1625, opened 2019-05-28
Add/custom matomo

dakotabenjamin:
This PR updates the matomo tracking code and cloudformation/circleci deployment:
1. Allows for different sites with matomo site IDs to be tracked separately
2. Removes hotjar

We had to update the [matomo tracking script](https://github.com/hotosm/matomo-tracking/blob/master/tracking-v3.js) to load it after the site_id variable was set. There is a small risk of losing views if the user leaves the site before the page is fully loaded. 

Thanks @JorgeMartinezG and @willemarcel for helping me with the angularjs stuff!!


xamanu:
Cool stuff!

Just for the record: This PR is related to issue https://github.com/hotosm/tasking-manager/issues/1589. And please squash the commits into one before merging.

dakotabenjamin:
Also need to roll https://github.com/hotosm/tasking-manager/issues/1783 into this

dakotabenjamin:
Closing this because we need to refactor with the redesign

PR #1647, opened 2019-05-29
Regular email updates for project managers (fixes #1377)

### LINKED ISSUE #1377: Give project managers regular updates on their projects via email
Keep the project managers updated. For example, send them stats of each project with current % mapped and validated, the change over the past week or month etc, # of invalidated tasks and so on.

fitoria:


xamanu:
Rebased and cleaned commits today.

I have some remarks on this, which i wrote in https://github.com/hotosm/tasking-manager/issues/1377#issuecomment-499087752. Basically to implement the surroundings that a user and admins needs to deal with such messages:

* Settings page to opt out of all emails and a way opt-out per project.
*  Link in the email footer to the "email preferences".
*  Per default, disable (or send to log) automatic messages. An let opt-in on certain deployments, like live. We probably don't want to send emails to project managers from the staging or a local developers' machines imp

Aadesh-Baral:
Reworked on #1935.

PR #1648, opened 2019-05-29
Encouraging emails for mappers (fixes #1378)

### LINKED ISSUE #1378: Give mappers feedback to their contribution and remind them to come back via email
Often mappers don't return to map after the first time. After a mapathon or event, send an automated thank you for mapping and a reminder to keep going.

fitoria:
This includes work for #1377 too.


xamanu:
Rebased and cleaned commits today.

I have some remarks on this, which i wrote in https://github.com/hotosm/tasking-manager/issues/1378#issuecomment-499089348. Basically not to send weekly mails to mappers but when a project reaches certain percentages of being mapped and validated.

Aadesh-Baral:
It is implemented in #5262 due to large no of merge conflicts and some implementation changes as mentioned in the comment above.

PR #1650, opened 2019-05-30
Rebase teachosm with master

arunasank:
cc/ @xamanu @dakotabenjamin @ramyaragupathy @willemarcel @smit1678 

Going to update the teachosm stack with this branch, after migrating the data, and share the link here.


xamanu:
This doesn't look very clean. I think we should just directly rebase `teachosm` on `master`. Doing this in coordination with @arunasank. Closing here.

PR #1652, opened 2019-05-30
Deploy version 3.3.0 to tasks.teachosm.org

xamanu:
Follow up to #1650 and #1651
Pinging @arunasank


xamanu:
It doesn't make sense to handle this deployment through a PR.

PR #1653, opened 2019-05-30
Task annotations display on task detail

xamanu:
Follow up to #1577.

The base to allow Task Annotations to be displayed on the task detail.


xamanu:
Superseded by #1653

Issue #1654, opened 2019-05-30
Tooltip in Messages - Inbox show extra text '| translate'


wireguy:
Hover mouse over an item in the inbox to get the tooltip popup and notice the text has the extra words  | translate.   Also, should the tooltip text be  'quoted'

![image](https://user-images.githubusercontent.com/47227325/58632772-a5dda000-82b4-11e9-88a9-48c0cba6f3bd.png)

xamanu:
The information used in the `title` attribute of the `td` element is used to generate the title, but a tooltip should probably not be appearing. Other tables in the tm usually do headers through the `tr` explicitly. @nrotstan, what do you suggest to address this?

xamanu:
Not an issue anymore with the new frontend of TM4.

PR #1658, opened 2019-05-31
Fix for #1654 - Add missing {{ }} around translate lines

wireguy:
Tooltip text not getting translated, and displaying incorrectly in English.
fix for #1654 

Also, 1654 only mentions Inbox, but a code search also found the issue in Profile.

**NOTE:** fist pull request, and I don't have a dev environment setup to test.


xamanu:
Checked this and the changes introduced in this PR would block the affected pages from loading. Unfortunately it is not that strait forward. I think there shouldn't be tooltips on the tables at all. Currently this information is used to generate the title row of the table.

xamanu:
That doesn't seem to be an easy fix. Please propose a working solution, or lets move on with TM4.

PR #1664, opened 2019-06-03
Version 3.3.1 Release

### LINKED ISSUE #1659: Project Admin - dashboard load fails
Per chat w/ @xamanu @russdeffner - a few project admins are noticing a project loading error on the dashboard with `504 - Timeout` error code.

Preliminary investigation shows this might be because of a nested loop situation.

This line here https://github.com/hotosm/tasking-manager/blob/develop/server/models/postgis/project.py#L348, which runs project summary for each project, is slowing down the process.  Project summary has been [updated](https://github.com/hotosm/tasking-manager/pull/1344/) in this recent release to do average time computations:
```
....

 "totalValidationTime": "22:47:45.990745",
  "totalTimeSpent": "11:33:46.122691",
  "averageMappingTime": "0:00:33.326294",
  "averageValidationTime": "0:08:17.363636",
  "status": "PUBLISHED"
```
Internally this fetches related tasks from the project table, iterates over the results and adds up time spent for each of the task. Now imagine this happening for 466 projects from @russdeffner  - it is a loop within a loop

```
all projects for a user -> for each project:
    fetch tasks: ->  iterate over task list and add up time:
```
This is a classic case of a nested loop that increases the time complexity of a program (O(n²)). When an admin has more projects, this complexity is enhanced resulting in a timeout error.

**How to solve?**
1. Remove time calculations from the project summary. Include time estimates in a different endpoint
2. Retain time calculations within the project summary, but sum up time values from the column for validation & mapping. A limitation is the field type (varchar). 
3. Retain time calculation with project summary, add a centroid field to the project dto. Now for project fetch within the admin dashboard, invoke project dto function instead of the project summary function. Not able to tell immediately if this will impact other endpoints. But @russdeffner & I tried it briefly this evening and the frontend map rendering fails.

**Next actions:**
- Try solution 2
- If solution 2 fails, look at solution 1 (separate out time calculations)

### LINKED ISSUE #1668: Invalidated tasks API fetch error
Noticing invalidated-tasks fetch failing with error code: 500 
`user/username/invalidated-tasks?asValidator=false&closed=false&page=1&pageSize=10&sortBy=updatedDate&sortDirection=desc`

More details from error response:

`Error: "Invalidated Tasks API - unhandled error: Can't resolve label reference for ORDER BY / GROUP BY. Textual SQL expression 'updated_date desc' should be explicitly declared as text('updated_date desc')"`

Pushing a fix

### LINKED ISSUE #1676: Project Admin - project deletion fails
Project deletion fails with error message:
```
Project GET - unhandled error: (psycopg2.IntegrityError) update or delete on table "projects" violates foreign key constraint "fk_message_projects" on table "messages"
DETAIL:  Key (id)=(3643) is still referenced from table "messages".
```

Sending a fix @xamanu

ramyaragupathy:
Fixes #1659 - using Postgres timestamp function in

- project/summary (this affects dashboard)
- stats/user/{username}
- stats/project/{id}/contributions
- user-id/{userid}

Fixes #1668 - using `text()` for `order_by` clause

Fixes #1676 - Define cascade relation for Messages table

### DIFF
--- a/server/models/postgis/message.py
+++ b/server/models/postgis/message.py
@@ -33,7 +33,7 @@ class Message(db.Model):
     # Relationships
     from_user = db.relationship(User, foreign_keys=[from_user_id])
     to_user = db.relationship(User, foreign_keys=[to_user_id], backref='messages')
-    project=db.relationship(Project, foreign_keys=[project_id])
+    project = db.relationship(Project, foreign_keys=[project_id], backref='messages', cascade='all,delete')
 
 
     @classmethod
--- a/server/models/postgis/project.py
+++ b/server/models/postgis/project.py
@@ -394,8 +394,6 @@ class Project(db.Model):
                 TaskHistory.action == 'LOCKED_FOR_VALIDATION',
                 TaskHistory.project_id == self.id
             ).distinct(TaskHistory.user_id).count()
-        current_app.logger.debug(unique_mappers)
-        current_app.logger.debug(unique_validators)
         summary.total_tasks = self.total_tasks
         summary.total_comments = db.session.query(ProjectChat).filter(ProjectChat.project_id == self.id).count()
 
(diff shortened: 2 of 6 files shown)


xamanu:
Combined into this PR the other fix (#1669)  that we want to include in the minor bugfix release.

PR #1667, opened 2019-06-03
Task annotation backend work

fitoria:


xamanu:
This is replaced by #1670

PR #1689, opened 2019-06-10
Compute tasks percentages based on sql response

### LINKED ISSUE #1527: Incorrect completion stats
Hello,

I am working on a project which is 102% mapped whereas [a task](https://tasks.hotosm.org/project/5847?task=335#bottom) is not mapped.

image

JorgeMartinezG:
this PR fixes #1527 .

Instead of computing percentages from model values, this PR creates a query to group and count all tasks from a project by the status that it has. Then, these count values are used for percentage computation of mapped and validated.

**how to test**.
1. Create a project.
2. Map some tasks and validate them.
3. go to project dashboard.
4. Verify percentages


xamanu:
I think this became a bit bigger. Closing here in favour of #1696

PR #1691, opened 2019-06-11
Do not apply cascade deletion on message model

JorgeMartinezG:
This Pr removes the cascade deletion on message model.

**How to test**
1. Apply required migration.
2. Alter your user as admin in your local instance.
3. Create a toy project. 
4. In _Questions and comments_ tab, reference yourself and send messages.
5. Then go to your inbox. The messages will appear.
6. Delete messages.
7. Go back to the project view. Also, you can verify executing a query to get the project from projects table.

### DIFF
--- a/migrations/versions/b25f088d40c2_.py
+++ b/migrations/versions/b25f088d40c2_.py
@@ -1,0 +1,24 @@
+"""empty message
+
+Revision ID: b25f088d40c2
+Revises: fcd9cebaa79c
+Create Date: 2019-06-11 12:31:41.697842
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = 'b25f088d40c2'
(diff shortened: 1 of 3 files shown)


xamanu:
Today talked with @ramyaragupathy and we were thinking about aiming at not deleting the messages when deleting a project gets deleted. And also not delete projects when messages get deleted. Messages with a deleted project should appear in the user's inbox but without a link, but a message clarifying the deleted project.

xamanu:
Changed base. please make sure this PR is based on and made for `master`. Thanks.

xamanu:
Rebased this on `master` and added a the composite foreign key condition to the `message` table. Seems to work here.

willemarcel:
It worked here. I had some problems applying the migration, but I think it was because I had to downgrade from other PR revision. Starting with an empty db, I could apply the migrations and test the message deletion.

PR #1700, opened 2019-06-14
Add markdown support to task annotations

geohacker:
_This is part of the ml-enabler work we're doing over at https://github.com/hotosm/ml-enabler/issues/9_

This PR adds support to submit markdown/string information at the task annotation level. One way TM would make use of this information is:
1. TM requests data from ml-enabler for a project geojson
2. ml-enabler aggregates predictions per task, and also prepares an info in markdown that's useful to be displayed in TM. For example this could have, information about the model, link to satellite imagery tile, link to vector/geometry API/tile
3. TM front-end would render this info when a user clicks the task.

cc @xamanu @batpad @wronk


server/models/postgis/task_annotation.py
```diff
@@ -12,17 +12,19 @@ class TaskAnnotation(db.Model):
...
     annotation_source = db.Column(db.String)
+    annotation_markdown = db.Column(db.String)
     updated_timestamp = db.Column(db.DateTime, nullable=False, default=timestamp)
     properties = db.Column(db.JSON, nullable=False)
 
     __table_args__ = (
         db.ForeignKeyConstraint([task_id, project_id], ['tasks.id', 'tasks.project_id'], name='fk_task_annotations'), db.Index('idx_task_annotations_composite', 'task_id', 'project_id'), {})
 
-    def __init__(self, task_id, project_id, annotation_type, annotation_source, properties):
+    def __init__(self, task_id, project_id, annotation_type, annotation_source, annotation_markdown, properties):
```
fitoria: Can this be refactored as:

__init__(self, task_id, project_id, annotation_type, properties, annotation_source=None, annotation_markdown=None): 

To keep the requried parameters first and the nullable as optionals with None as default value?

server/services/task_annotations_service.py
```diff
@@ -13,7 +13,8 @@ def add_or_update_annotation(annotation, project_id, annotation_type):
 
         task_id = annotation['taskId']
         source = annotation['annotationSource'] if 'annotationSource' in annotation.keys() else None
-        task_annotation = TaskAnnotation(task_id, project_id, annotation_type, source, annotation['properties'])
+        markdown = annotation['annotationMarkdown'] if 'annotationMarkdown' in annotation.keys() else None
```
fitoria: Can this be changed to:

markdown = annotation.get('annotationMarkdown', None)

xamanu:
Ah, please make sure the test run through. Thanks.

geohacker:
> Please check the tests run through. Thanks.

Sorry my bad! Just fixed.

PR #1703, opened 2019-06-18
Adding link to user manual on teachosm.org

xamanu:


Nick-Tallguy:
....er - shouldn't this read learnosm.org ?

PR #1704, opened 2019-06-18
Fix db migration with messages relating to non-existing tasks

xamanu:
This should fix the failing db upgrade during deployment by setting in the database all `task_id` to `NULL` for messages, where the task doesn't exist anymore.


xamanu:
After @ramyaragupathy review I would schedule this for merging right after the project restoration activity.

xamanu:
Thanks for the reviews. Moved into #1711 for easier deployment.

Issue #1687, opened 2019-06-10
Projects inaccessible


ramyaragupathy:
Following projects which have been active with contributions are no longer accessible:

- 5375, 5762, 6067, 5773, 5740 & 1945

xamanu:
As per communication with @ramyaragupathy I'm reporting here the current status:

The error seems to be caused by deletion of messages, that would cascade in the database and delete the associated project because of a foreign key connection of the messages table to the project with its `id`.

For now a [quick fix has been introduced](https://github.com/hotosm/tasking-manager/pull/1690), which impedes users to delete messages for the time being.

The envisioned solution is aiming at not deleting the messages when deleting a project gets deleted. And also not delete projects when messages get deleted. Messages with a deleted project should appear in the user's inbox but without a link, but a message clarifying the deleted project.

xamanu:
Today we were able to provide a fix for the deletion of messages and it is currently deploying to live. Next step is to bring back the projects.

And then there are some follow up issues in order to improve the infrastructure to react better to such emergencies: #1693, #1694, #1695

rhinton17:
Validation of project #2742 is underway but is also inaccessible at this time.

lusdavo:
Project #[5907](https://tasks.hotosm.org/project/5907), 100% mapped and about 94% validated is as well inaccessible.

xamanu:
We were able to fix this issue and we have a solution to bring back the inaccessible projects. For this we will schedule a downtime this week.

xamanu:
The project restoration happened last Friday and seems to work well. Thanks for all this good work. Closing here.

PR #1705, opened 2019-06-19
Scripts to restore projects

ramyaragupathy:
This PR includes @JorgeMartinezG's script to restore deleted projects #1687


xamanu:
Thanks for the reviews. Moved into #1711

PR #1706, opened 2019-06-19
Introduce command to extract strings for translations, make them required for any changes and update documentation

xamanu:
* Adds a script to extract strings for translations into json files
* Add `make` command to call the script
* Adjust link to renamed transifex project
* Update `en.json` file, which is the source for the translation and it is getting pulled from `develop` automatically now. This will make sure people are always translating for `develop` and before the release we have to pull the files down and include them.
* Complete documentation
* Added test for unupdated `en.json` file (comparing it with the code base)


ethan-nelson:
back in town :+1: will review this weekend

devops/docker/tasking-manager/Dockerfile
```diff
@@ -12,7 +12,7 @@ RUN apt-get update \
 # Upgrade pip
 RUN pip install --upgrade pip
 
-ARG branch=develop
+ARG branch=feature/1678-translations
```
xamanu: This needs to be changed back to `develp` before merging in. But should stay as it is for testing purposes.

server/services/translation_service.py
```diff
@@ -0,0 +1,58 @@
...
+
+    @staticmethod
+    def _get_translatables():
+
+        # Source files containing translation strings
+        files = glob2.glob('client/app/**')
+        files.append('client/index.html')
+
+        # Regular expression to detect translation strings
+        translate_string = re.compile('\{\{\s+[\'\"]([^\|]*)[\'\"]\s+\|\s+translate\s+\}\}')
```
ethan-nelson: My main suggestion would be to change `+`s to `*`s here as there seem to be a few translatable strings without the prettified spacing. I guess ideally we would be consistent across all the strings about the spacing, so I can also see the case to not include that--we'll just need to fix some of those inconsistencies (e.g. [[1](https://github.com/hotosm/tasking-manager/blob/14c4d5885ad912c688391e15ede24cd7e6f31602/client/app/about/faq.html#L19)][[2](https://github.com/hotosm/tasking-manager/blob/659ee5b9960ced26b04042c88d393ca5a5c3da17/client/app/about/about.html#L15)] and about 40 others).

xamanu: Thanks, @ethan-nelson! Adjusted as you suggested.

PR #1707, opened 2019-06-19
Avoid providing standard changeset comment

xamanu:
Fixes #1663

This PR requires instances to set their own changeset comment. This is very important because they should never be the same between instances.

### DIFF
--- a/manage.py
+++ b/manage.py
@@ -10,21 +10,21 @@ from server.services.users.user_service import UserService
 
 import os
 import warnings
- 
+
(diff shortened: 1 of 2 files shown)


xamanu:
Oh, yes, changed the issue id. I think `None` is better. Anyway the TM won't start without the `TM_DEFAULT_CHANGESET_COMMENT` being set. This PR includes a check and fails if not set.

xamanu:
Included into #1711

PR #1709, opened 2019-06-19
Update translations

xamanu:


xamanu:
Thanks for your reviews. Moving into #1711.

PR #1710, opened 2019-06-19
Change ASG Policy to Target Request Count threshold (#1701)

### LINKED ISSUE #1701: Production site is not autoscaling
A mapathon coordinator flagged an issue that TM was running slowly, because autoscaling is not occurring. 

The [current scaling policy](https://github.com/hotosm/tasking-manager/blob/develop/devops/cloudformation/tasking-manager.template.js#L129) will autoscale when the Autoscaling group Average CPU breaches an 85% usage threshold. 

- [x] Check CloudWatch metrics during last week's mapathons to see what the threshold should be, or what other metrics may be more effective to use as an autoscaling trigger
- [x] Change cloudformation template resource "TaskingManagerScaleUp" with improved metrics and values.

dakotabenjamin:
Resolves #1701 

This PR would change the Autoscaling Policy from monitoring CPU to trigger new instances to using Target Requests (see https://github.com/hotosm/tasking-manager/issues/1701#issuecomment-503731829)

I'm open to turning this into a hotfix @xamanu

### DIFF
--- a/devops/cloudformation/tasking-manager.template.js
+++ b/devops/cloudformation/tasking-manager.template.js
@@ -127,9 +127,9 @@ const Resources = {
         AutoScalingGroupName: cf.ref('TaskingManagerASG'),
         PolicyType: 'TargetTrackingScaling',
         TargetTrackingConfiguration: {
-          TargetValue: 85,
+          TargetValue: 400,
           PredefinedMetricSpecification: {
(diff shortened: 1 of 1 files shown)


xamanu:
Yeah, please target for the branch `deployment/hot-tasking-manager`.

xamanu:
Included this one into #1711 closing here.

PR #1713, opened 2019-06-23
Add Korean to the dropdown menu for language

depth221:
I checked that there is Korean in client/locale directory and .tx/config file, but not in server/config.py. I'm seemed to that's why the dropdown menu in Tasking Manager does not include Korean.


xamanu:
Many thanks @depth221! Just rerolled your commit and submitted it for the next release in #1717.

Issue #1731, opened 2019-06-28
Prevent SQL Injection when executing custom SQL


dodobas:
When using simple Python string formatting methods input values are not escaped which opens up the possibility of SQL injection attacks. For example: https://github.com/hotosm/tasking-manager/blob/master/server/models/postgis/task.py#L653

SQLAlchemy provides a way to safely pass values to a custom SQL - https://docs.sqlalchemy.org/en/13/core/tutorial.html#using-textual-sql

PR #1732, opened 2019-06-28
Prevent SQL Injection when executing custom SQL

dodobas:
Fixes: https://github.com/hotosm/tasking-manager/issues/1731


xamanu:
Thanks a lot for this. Rerolled in #1738.

PR #1733, opened 2019-06-28
Enforces PM or Admin role on the users that will get a project transfer

fitoria:


xamanu:
When trying to transfer the project i get an error:
image

The logs show: 
`[2019-07-02 16:25:55,977] CRITICAL in project_admin_api: Project GET - unhandled error: Invalid owner_id`

server/services/project_admin_service.py
```diff
@@ -229,7 +229,11 @@ def transfer_project_to(project_id: int, author_id: int, username: str):
         project = Project.get(project_id)
         if project.author_id == author_id:
```
xamanu: Shouldn't this be `!=`?

Issue #675, opened 2017-08-16
Add option for AND operator for the terms in the project search


bgirardot-np:
It looks like right now the search terms put in the full text search are 'or' so "American Samoa" returns anything with american or samoa.

The search terms should be 'and' so anything with american and samoa.

hunt3ri:
I'd be careful before implementing this, as it means ALL  terms in the description need to be present which in local testing was extremely frustrating for users, as they often got no results from the text search.  Possibly quoting the string would treat it as a whole rather than two separate words?

bgirardot:
@hunt3ri I think the quoted string sounds like a good compromise.

Any chance we could support the word "and" as an operator? e.g., "missing maps" and laos ?

bgirardot:
This issue can be partially addressed in server/services/project_search_service.py at line 132

fiennyangeln:
Hi @bgirardot may I take this issue? i will try looking at the code first

fiennyangeln:
@bgirardot I think the text search need some database trigger to work. Thats why it doesnt work on local dev, including mine. However, when I tried searching for the database trigger in the tm database, there is a lot of trigger in it that I could not differentiate.

bgirardot:
@fiennyangeln Does this information help at all? Especially toward the bottom where it lists the trigger and the procedure that needs to be executed.

tm3=> \d+ project_info
                                    Table "public.project_info"
        Column         |          Type          | Modifiers | Storage  | Stats target | Description 
-----------------------+------------------------+-----------+----------+--------------+-------------
 project_id            | integer                | not null  | plain    |              | 
 locale                | character varying(10)  | not null  | extended |              | 
 name                  | character varying(512) |           | extended |              | 
 short_description     | character varying      |           | extended |              | 
 description           | character varying      |           | extended |              | 
 instructions          | character varying      |           | extended |              | 
 project_id_str        | character varying      |           | extended |              | 
 text_searchable       | tsvector               |           | extended |              | 
 per_task_instructions | character varying      |           | extended |              | 
Indexes:
    "project_info_pkey" PRIMARY KEY, btree (project_id, locale)
    "idx_project_info composite" btree (locale, project_id)
    "textsearch_idx" gin (text_searchable)
Foreign-key constraints:
    "project_info_project_id_fkey" FOREIGN KEY (project_id) REFERENCES projects(id)
Triggers:
[...]

PR #1735, opened 2019-06-28
Include automatic trigger for indexing text_searchable

### LINKED ISSUE #1723: Project search by number is not consistent
While searching for projects by a specific number, there are certaincases where the results are not returned. This behavior seems random though

> entering '6179' into the search box delivers no match - searching for 'chad' delivers project 6179 as the second match

project_search

JorgeMartinezG:
This PR fixes #1723 . Current database status with latest migration file does not have automatic trigger to index information into the `text_searchable` column from project_info table. This disallows project search since text_searchable column is NULL

The commit enables and automatic trigger which indexes before an insert or update in project info table.

**how to test**:
1. Create a fresh database.
2. Run migrations
```python manage.py db upgrade```
3. Downgrade
```python manage.py db downgrade```
4. Create a project with description and short description.
5. Verify that text_searchable column is null.
```SELECT text_searchable FROM project_info```
6. Update migrations
```python manage.py db upgrade```
7. Edit project_info again
8. Verify that information is indexed
```SELECT text_searchable FROM project_info```

NOTE: This trigger will work also for previous project

### DIFF
--- a/migrations/versions/451f6bd05a19_.py
+++ b/migrations/versions/451f6bd05a19_.py
@@ -1,0 +1,24 @@
+"""empty message
+
+Revision ID: 451f6bd05a19
+Revises: a43b9748ceee
+Create Date: 2019-07-02 17:03:55.567940
+
+"""
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = '451f6bd05a19'
+down_revision = 'a43b9748ceee'
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by xamanu** [comment]
Factored that out into its own migration. However the migration throws the following error:

```
psycopg2.ProgrammingError: trigger "tsvectorupdate" for relation "project_info" already exists
```

JorgeMartinezG:
@xamanu does it happen with a new database or a database dump? I found that the db dump has already the trigger. However, when you run migrations from the beginning, the trigger does not appear

xamanu:
I tested it with a live dump from june13, right before the project restoration and version 3.3.2

JorgeMartinezG:
Looks like this issue has been hanging around since #675

PR #1737, opened 2019-07-02
Refine packer template for tasking manager

eternaltyro:
Ever so gently, refine the packer template for the instance hosting
tasking manager.

- Parameterize instance type, region and SSH username
- Use a source_ami_filter to pick up the latest Ubuntu 16.04 AMI
  as source image instead of hard-coding an AMI ID
- Add image tags
- Re-organize the instance bootstrap commands; consolidate some of
  them to make it easier to read.
- Add inline_shebang for provisioning script

TODO:

- Figure out steps to prepare for a move to Debian 9 or 10.
- Figure out if NodeJS can be upgraded to version later than 6
- Figure out if in the interim Ubuntu version could be bumped to
  18.04 (Bionic)
- Find out ways to use default Python instead of external apt sources
- Find out if browser-sync and gulp can be added to package.json and
  installed using just npm install.


xamanu:
Shall we make a hotfix release out of this? Or wait for the next release, which might take some time?

dakotabenjamin:
Correct me if I'm wrong @eternaltyro but this PR is not active/superseded by https://github.com/hotosm/tasking-manager/pull/1739

eternaltyro:
That is correct @dakotabenjamin

xamanu:
Thanks, closing here then.

PR #1738, opened 2019-07-02
Prevent SQL Injection when executing custom SQL

xamanu:
Fixes: https://github.com/hotosm/tasking-manager/issues/1731


xamanu:
Rerolled this from #1732 on latest `develop` and changes. Tested it and seems to work well.

PR #1739, opened 2019-07-03
WIP: Move instance script out of template

eternaltyro:


infra/bootstrap.sh
```diff
@@ -0,0 +1,70 @@
...
+
+sudo apt-get -y install \        # Install Mapping related stuff
+    libgdal1-dev \               # Geospatial Data Abstraction Library (GDAL)
+    libgeos-3.5.0 libgeos-dev \  # Geometry engine
+    libproj9 libproj-dev \       # Cartographic projects & translations between CRS
+    libxml2 libxml2-dev \        # XML parsing and such
+    libjson-c-dev                # Construct JSON objects in C
+
+sudo apt-get -y install \
+    postgresql-9.5 \             # PostgreSQL server
```
xamanu: I think we are already running on postgresql 11 (also on live!?).

infra/ami.json
```diff
@@ -0,0 +1,49 @@
...
+      "Operating system: Ubuntu 16.04",
+      "Region: us-east-1",
+      "Instance Type: t2.micro",
+      "TODO: Move to Debian Stretch / Buster",
+      "TODO: Switch to NodeJS v10"
+  ],
+  "variables": {
+    "aws_access_key": "{{env `AWS_ACCESS_KEY_ID`}}",
+    "aws_secret_key": "{{env `AWS_SECRET_ACCESS_KEY`}}",
+    "instance_type": "t2.micro",
```
xamanu: Not sure whether this is relevant here, recently we have been using bigger machines.

infra/bootstrap.sh
```diff
@@ -0,0 +1,70 @@
...
+
+# Setup Node.JS v10.x
+curl -sL https://deb.nodesource.com/setup_10.x | sudo -E bash -
+sudo apt-get -y install nodejs
+
+# Install dependencies
+sudo npm install gulp -g
+npm install browser-sync --save # Probably add this to package.json?
+
+sudo git clone --recursive https://github.com/hotosm/tasking-manager.git /tasking-manager
```
xamanu: Most recently we had challenges with this being a fallback to latest `develop`. And it would be good to specify at this moment, and rely on the specific branch for the deployment here (e.g. `deployment/hot-tasking-manager`).

infra/ami.json
eternaltyro: This is simply the machine type it launches to configure in order to make an image out of. The type of instance at launch time is decided by us.

infra/bootstrap.sh
eternaltyro: Then we can change this. But Ubuntu 16.04 does not ship PostgreSQL 11. Ubuntu 18.04 ships PostgreSQL 10. Debian Buster will have PostgreSQL 11. Either way, not sure why we need the server on the instance if we use RDS?

infra/bootstrap.sh
eternaltyro: Yes, we can change this to deploy the appropriate branch.

infra/bootstrap.sh
xamanu: Right. If this is not necessary. Then we should probably omit.

eternaltyro:
Some weird failures:

```
==> amazon-ebs: E: Unable to locate package awscli
==> amazon-ebs: E: Package 'ruby' has no installation candidate
==> amazon-ebs: E: Package 'unzip' has no installation candidate
==> amazon-ebs: E: Package 'grub' has no installation candidate
```

<details>
```
    amazon-ebs: Package unzip is not available, but is referred to by another package.
    amazon-ebs: This may mean that the package is missing, has been obsoleted, or
==> amazon-ebs: E: Unable to locate package awscli
==> amazon-ebs: E: Package 'ruby' has no installation candidate
==> amazon-ebs: E: Package 'unzip' has no installation candidate
==> amazon-ebs: E: Package 'grub' has no installation candidate
    amazon-ebs: is only available from another source
==> amazon-ebs: E: Unable to locate package kpartx
    amazon-ebs:
    amazon-ebs: Package grub is not available, but is referred to by another package.
    amazon-ebs: This may mean that the package is missing, has been obsoleted, or
    amazon-ebs: is only available from another source
    amazon-ebs: However the following packages replace it:
    amazon-ebs:   grub2-common grub-pc
    amazon-ebs:
    amazon-ebs: Package ruby is not available, but is referred to by another package.
    amazon-ebs: This may mean that the package is missing, has been obsoleted, or
    amazon-ebs: is only available from another source
```
</details>

eternaltyro:
Similar report: https://askubuntu.com/questions/745610/apt-get-reports-no-installation-candidate-randomly-for-no-good-reason

infra/ami.json
```diff
@@ -0,0 +1,49 @@
...
+      "component": "tasking_manager",
+      "built_by": "packer",
+      "created_by": "{{user `created_by`}}",
+      "Name": "{{user `name`}}"
+    }
+  }],
+  "provisioners": [
+    {
+    "type": "shell",
+    "script": "boostrap.sh"
```
dakotabenjamin: typo here

dakotabenjamin: Typo, should be bootstrap.sh

dakotabenjamin: idk why github isn't showing latest changes- please ignore

eternaltyro:
> When I ran packer build infra/ami.json it failed to find the bootstrap script.

From the documentation: 

> This path can be absolute or relative. If it is relative, it is relative to the working directory when Packer is executed.

It makes sense for the CI to run packer from the root path. I'll change it.

infra/bootstrap.sh
eternaltyro: I'll leave it in for now until I figure out why it was put in in the first place.

eternaltyro:
@dakotabenjamin made the changes, please check.

Moved `infra` dir to inside `devops` and switched relative script path inside ami.json.

xamanu:
Creat! Can you please sqaush the commits into one? This will go to `develop` now, correct? And end up in a release within a couple of months. Or should this go live sooner?

eternaltyro:
@xamanu why squash the commits into one?

xamanu:
We do this to have a nicer commit history in the main branch with meaningful messages. In case it makes sense to have a couple that is also fine.

eternaltyro:
Understood, I think it should be possible to squash merge from the UI?

xamanu:
Sure, we can squash them from the UI. It gives you more space when doing yourself.

PR #1742, opened 2019-07-05
make api-docs url load even with trailing backslash

willemarcel:
api-docs was failing to load if we access with a trailing backslash

### DIFF
--- a/server/web/controller.py
+++ b/server/web/controller.py
@@ -32,7 +32,7 @@ def app(path):
     return send_from_directory(main.static_folder, 'app/' + path)
 
 
-@main.route('/api-docs')
+@main.route('/api-docs/')
 def api():
     """
     Route for API Docs welcome page


willemarcel:
I just noticed that all TM URLs don't work with a trailing backslash. If this is the desired pattern, I think it's better to discard this pull request.

We could think about enabling all URLs to work with trailing backslashes.

xamanu:
I don't think this is a design pattern and I would be totally up to support them. Or does anybody has a reason why we shouldn't?

eternaltyro:
This is a flask idiosyncrasy. Quoting flask documentation:

> An important detail to keep in mind is how Flask deals with trailing slashes. The idea is to keep each URL unique so the following rules apply:
>
>    1. If a rule ends with a slash and is requested without a slash by the user, the user is automatically redirected to the same page with a trailing slash attached.
>
>    2. If a rule does not end with a trailing slash and the user requests the page with a trailing slash, a 404 not found is raised.

I think we should make trailing slashes to be seamlessly redirected to path without trailing slashes. It is general convention to do so. This PR is valid.

xamanu:
Added an issue for all routes: #1797

Issue #1282, opened 2018-11-23
Split moving tasks to another location in the world


ethan-nelson:
The project https://tasks.hotosm.org/project/5537 is acting strangely: when you split a task, it disappears from the main grid in Timor-Leste and instead moves to the middle of the Indian Ocean. The tasks were originally around https://tasks.hotosm.org/project/5537?task=40 but when split move to around https://tasks.hotosm.org/project/5537?task=70

Looking at commit history for the most recent version push to master, this _seems_ to be a bug that was already present in the code as we did not touch the geometry calculations when ensuring task split history is preserved (https://github.com/hotosm/tasking-manager/blame/master/server/services/grid/split_service.py). No issues like this came up when testing on the staging instance, and projects that are at the top of the list on live do not display this problem when splitting either--so it does not appear to be widespread at this time. Nevertheless, still a problem.

Thanks to Andrew for finding this!

ethan-nelson:
https://tasks-stage.hotosm.org/project/5535 is an example of this occurring on staging

I've been looking at the raw tasks geojson from https://tasks-stage.hotosm.org/api/v1/project/5535/tasks?as_file=false and picked one of the tasks of the original zoom level and one at the next zoom level that was split. 

```
19 |  
-- | --
type | "Feature"
geometry |  
type | "MultiPolygon"
coordinates |  
0 |  
0 |  
0 |  
0 | -42.1874999924431
1 | 72.3957057007426
1 |  
0 | -41.8359374925061
1 | 72.3957057007426
2 |  
0 | -41.8359374925061
1 | 72.5017223456187
3 |  
0 | -42.1874999924431
1 | 72.5017223456187
4 |  
0 | -42.1874999924431
1 | 72.3957057007426
properties |  
taskId | 41
taskX | 392
taskY | 816
taskZoom | 10
taskIsSquare | true
taskStatus | "READY"
20 |  
type | "Feature"
geometry |  
type | "MultiPolygon"
coordinates |  
0 |  
0 |  
0 |  
0 | -68.6434492452437
1 | 42.1845617326509
1 |  
0 | -68.6434492452437
1 | 42.5957829683862
2 |  
0 | -68.0866664916311
1 | 42.5957829683862
3 |  
0 | -68.0866664916311
1 | 42.1845617326509
4 |  
0 | -68.6434492452437
1 | 42.1845617326509
properties |  
taskId | 39
taskX | 200
taskY | 407
taskZoom | 9
taskIsSquare | true
[...]

PR #1751, opened 2019-07-11
Country info in Project metadata

ramyaragupathy:
Closed the [earlier PR](https://github.com/hotosm/tasking-manager/pull/1611) due to rebase issues.

**Implementation Items:**
- [x] Setup migration to populate country info for existing projects
- [x] Frontend element in the metadata section to display existing country field
- [x] API call to fetch the country list
- [x] Upsert function to update country info
- [x] Geocoding API call to determine country with the initial geometry
- [x] Fix failing tests

image

cc @xamanu


ramyaragupathy:
Unit tests fails at `test_translation_source_file_contains_all_translatables - tests.server.unit.services.test_translation_service.TestTranslationService`

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/unittest/case.py", line 59, in testPartExecutor
    yield
  File "/usr/local/lib/python3.7/unittest/case.py", line 628, in run
    testMethod()
  File "/home/circleci/app/tests/server/unit/services/test_translation_service.py", line 40, in test_translation_source_file_contains_all_translatables
    self.assertEqual(count_new_strings, 0)
  File "/usr/local/lib/python3.7/unittest/case.py", line 852, in assertEqual
    assertion_func(first, second, msg=msg)
  File "/usr/local/lib/python3.7/unittest/case.py", line 845, in _baseAssertEqual
    raise self.failureException(msg)
AssertionError: 3 != 0
```

Haven't touched any of the translation file in this branch. Not sure what's causing this

ramyaragupathy:
One other issue:
- Country information was not populated for close to 1500 projects. I suspect this is related to ocean geometry from task splitting issue https://github.com/hotosm/tasking-manager/issues/1282

ramyaragupathy:
Per voice w/ @xamanu - unit test failure fixed with `python3 manage.py refresh_translatables` to export recent strings from front end files 🎉

server/models/postgis/project.py
```diff
@@ -135,6 +139,30 @@ def set_default_changeset_comment(self):
         self.changeset_comment = f'{default_comment}-{self.id} {self.changeset_comment}' if self.changeset_comment is not None else f'{default_comment}-{self.id}'
         self.save()
 
+    def set_country_info(self):
+        """ Sets the default country based on centroid"""
+        query = """SELECT ST_Y(ST_Transform(ST_Centroid(centroid), 4326)) || ', '
```
JorgeMartinezG: @ramyaragupathy I think we can use sqlalchemy here. Something like a static method similar to https://github.com/hotosm/tasking-manager/blob/develop/server/models/postgis/project.py#L632 :
```
lat, lng = db.session.query(func.ST_Y(Project.centroid), func.ST_X(Project.centroid)).filter(Project.id==project_id).one()
```

ramyaragupathy: Thanks for the flag @JorgeMartinezG. Replaced the centroid query and also the country fetch query

xamanu:
Started testing this today, but then as per conversation with @ramyaragupathy we should wait with the review still.

ramyaragupathy:
Looking to reduce the migration time (~2 hrs) by increasing the match probability using continent info

ramyaragupathy:
> Looking to reduce the migration time (~2 hrs) by increasing the match probability using continent info

This now takes about 40 minutes

ramyaragupathy:
Now the migration is done in 20 minutes 🎉

thadk:
I think it would be best to adapt the React interface with this after the `/client` code is removed and the APIv2 is in here. Looks okay though

ramyaragupathy:
Moving this to https://github.com/hotosm/tasking-manager/pull/1863

Issue #1628, opened 2019-05-28
Implement new project cards


willemarcel:
Improve the project cards on the contribute/explore page and appear as recommendations in other parts of the application.

Design reference: https://projects.invisionapp.com/share/VCQTFR6YSJR#/screens/353854508

thadk:
@xamanu @hexiaok  I noticed the final design does not show the Feature/Popular label above and to the left of the card in any of its examples. Is this still called for?

willemarcel:
@ramyaragupathy @JorgeMartinezG In the new frontend, the projects cards needs to show the `lastUpdated` date, but it's not on the `/project/search` endpoint yet. Could you add it?

thadk:
Thanks @JorgeMartinezG! Here are the fields I hope to connect to for this issue and https://github.com/hotosm/tasking-manager/issues/1639 . Any further help or pointers very welcome :

- [ ] Can you confirm that "lastUpdated" is the same as "last contribution by users"? 
- [ ] When the number of contributions is large, the design recommends using "number of contributions" instead of "last contribution by users" date/time. Is this available somewhere already?
- [ ] The other question I had was about how to query the search API for "featured" projects. (I think the idea of "popular projects" was removed from the design?)
- [ ] Is there a country-of-project field anywhere? The design calls for a relatively short filterable country/location name (Mozambique; Colombia; DRC) as both a query parameter and as data on the project. I'd be happy to help generate this field if there is nothing like it yet using a spatial join & natural earth data on some sort of data dump.
- [ ] The design calls for "total mappers" but I only see "active mappers" in the API – do we think this is okay?
- [ ] the "dueDate" field, rarely used, which is currently only on the project detail/summary endpoint, should maybe be in the project search as well.

willemarcel:
@thadk yes for the first topic, I confirmed it.
- About the country information, there is a PR https://github.com/hotosm/tasking-manager/pull/1751 
- About the campaign/mapping type, I think there is already a PR https://github.com/hotosm/tasking-manager/pull/1762
- Yes, the dueDate should be in the `search` endpoint.

JorgeMartinezG:
@thadk To include more information:

  **When the number of contributions is large, the design recommends using "number of contributions" instead of "last contribution by users" date/time. Is this available somewhere already? I only see "active mappers" in the API – do we think this is okay?**

There is also a PR which checks the number of contributions per day of a project https://github.com/hotosm/tasking-manager/pull/1768

[...]

PR #1764, opened 2019-07-18
 Feature/redesign ProjectCard

thadk:
Initial work for new React frontend around project cards for #1628. Added a featuredProjects component to home page, and will support local pagination using red left/right arrows.

(With black instead of white-based organization logos)
image

image

image


willemarcel:
Hi @thadk 

I pushed some improvements to https://github.com/hotosm/tasking-manager/commit/604271cd8b58789028d5bd65aa7ccc806dcb5e32 . Don't worry about  linting. I'm planning to add `prettier`.

Some details I found:

- [ ] The tooltip didn't show when the mouse is over the progress bar, but when it's almost on the bottom border of the card.
- [ ] We need a tooltip for validated percentage and another to mapped. For me only the validated is shown.
- [ ] The due date box wasn't exhibited here, I checked the API result and the project has a due date.
- [ ] On the image below, we can see `...` below the project title. Is it a replacement to some missing information?
- [ ] 2 tests are failing.

Screenshot from 2019-07-31 19-05-22

thadk:
> Hi @thadk
> 
> I pushed some improvements to [604271c](https://github.com/hotosm/tasking-manager/commit/604271cd8b58789028d5bd65aa7ccc806dcb5e32) . Don't worry about linting. I'm planning to add `prettier`.
> 
> Some details I found:
> 
>     * [ ]  The tooltip didn't show when the mouse is over the progress bar, but when it's almost on the bottom border of the card.
This was happening because of the new `shadow-hover` hovering background.

>     * [x]  We need a tooltip for validated percentage and another to mapped. For me only the validated is shown.
Not sure why only validation was appearing but I have moved the "Mapping %" hover to the full width div instead of the mapping div so it doesn't happen.

>     * [ ]  The due date box wasn't exhibited here, I checked the API result and the project has a due date.
I will modify my database and check this late today. Are you sure the due date is still in the future on your database? It never appears for past-due dates.

>     * [ ]  On the image below, we can see `...` below the project title. Is it a replacement to some missing information?
This was only happening with the test database items, not for the real items from the live server. Not sure what is special about them yet. 
I like the idea of truncating project titles in the CSS but maybe this is better to do in the React code? Have you seen any good libraries?

>     * [x]  2 tests are failing.
Thanks, renamed and fixed. Hope there is only 1 test for now on this branch.

willemarcel:
> I like the idea of truncating project titles in the CSS but maybe this is better to do in the React code? Have you seen any good libraries?

I have never used a library to do it. Don't care too much about it, we can communicate with the project managers to use a max title length or refine the implementation on the future.

thadk:
@willemarcel I think I have resolved each of your concerns!

willemarcel:
Merged with https://github.com/hotosm/tasking-manager/pull/1817

PR #1769, opened 2019-07-19
Add authentication views

willemarcel:


xamanu:
Can we clean these commits up and squash them into one or two? Thanks.

willemarcel:
@xamanu it was already squashed on the develop branch, but @fitoria forked it before I squashed it. I rebased the branch created by @fitoria with develop. Is it ok to squash again?

willemarcel:
@fitoria I was thinking that it would be a good improvement if the `redirect_to` parameter could receive not only the end of the url but the complete url (for example: `http://localhost:3000/login/`), that way we would avoid problems when people are accessing from localhost or even allow us to authenticate in a remote server (staging, for example). Could you try to make that change?

frontend/src/store/actions/auth.js
```diff
@@ -0,0 +1,49 @@
...
+  CLEAR_SESSION: 'CLEAR_SESSION'
+};
+
+export function clearUserDetails() {
+  return {
+    type: types.CLEAR_SESSION,
+  };
+}
+
+export const logout = () => dispatch => {
```
thadk: For Firefox, I wasn't able to replicate the problem Wille described where these come back after a refresh in same tab after logging out. It cleared them out completely.

willemarcel: Thanks, @thadk !

Issue #1590, opened 2019-05-22
Improve campaigns


xamanu:
Similarly as we are moving organizations from free tags to database objects (#1531) that can be referenced by project, that are grouped under one campaign. Campaigns should belong to none, one of several organizations and their admins will be able edit campaigns and add projects to them.

**Implementation steps:**

* [x] Implement database change for campaigns and changes to the Tasking Manager (filters for projects, etc.) #1606 
* [x] Add API views to associate and unassociate campaigns with users (in order to recommend projects based on their interest) #1607
* [x] Develop migration path for existing tags into campaigns. This fixes https://github.com/hotosm/tasking-manager/issues/943.
* [x] Allow editing of campaigns and associating it to organizations. If it is connected to an organization, admins of this organizations can edit, too.
* [x] Restrict the selection of campaigns to admins of associated organizations and campaigns without any affiliation to an organization.
* [x] Create campaign pages (https://github.com/hotosm/tasking-manager/issues/583)

xamanu:
This will be targeted for the redesign with the new framework #1541.

uwaiszaki:
@xamanu I have some doubts regarding the structure of campaigns part.
1) How projects and organisations should be connected to campaigns.
2) What should be the process to assign a campaign to a project? Whether it will be first assigned to the organization and then to the projects or should it be independently linked with the projects.

uwaiszaki:
![gsocstructure1](https://user-images.githubusercontent.com/32809190/59446604-720b8a00-8e1f-11e9-862e-f932a2e84b21.png)
After discussion with @xamanu, we are going to use this structure for teams, organisations, and campaigns.
1) Structure for teams is based on the discussion in #1531.
2) user-interest-campaigns table will be used to show the recommended campaigns and projects to the users based on their previous searches. 
3) For many to many relations between campaigns and projects/organisations, campaign_projects and campaign_organisations will be used.
4) The process to assign a project to an organisation will be done by first assigning organisation to it and then assigning projects under that organisation.

xamanu:
This is our proposal. @hotosm/tech and @zlavergne please let us know what you think. Thanks.

verdy-p:
in your schema, "campaign_projects" references:
- (by foreign key organisation_id) "organisations".id, and
- (by foreign key project_id) and "projects".id
The projects table then references (by by foreign key organisation_id)
"organisations".id again

So it is not normalized, there's two possible organisations from the
[...]

PR #1770, opened 2019-07-20
[WIP] Team & Orgs Backend

### LINKED ISSUE #1531: Teams and Orgs backend
We are working on implementing a simple model for teams and organizations to satisfy our internal requirements. It's a simple design that we hope can be the base of future teams/orgs infrastructure. 

Our current design creates the following tables:

#### organizations
list of organizations

| Column | Type|
|--|--|
| id|int|
|name|varchar|
|visibility|ENUM|

#### org_admins
list of users who can make changes to organizations

|Column|Type
|-|-|
|user_id (FK) |int|
|org_id(FK)|int|

#### teams
list of teams and the org they belong to  (team doesn't have to belong to an org)

|Column|Type
|-|-|
|id |int|
|name|varchar|
|org_id (FK)|int|

#### team_members
list of users in a team

|Column|Type
|-|-|
|team_id (FK) |int|
|user_id (FK)|int|
|role*|ENUM|

\* Roles: `manager` or `editor`

#### projects
Change the current organization implementation

|Column|Type|
|-|-|
|~organization_tags~||
|organization_id (FK)| int|

This is a simple implementation that satisfies our internal requirements. With that, we don't plan on building out all the features of teams but rather create a good base on which the rest of the infrastructure can be built. 

I would appreciate any feedback, especially if anyone sees any blockers for iteration.

zlavergne:
This is the WIP backend work for teams and orgs so the work for campaigns can be finished.

I'm not expecting changes but I would prefer if it wasn't merged until I give the go-ahead just in case 🤓 

**Note:** This PR requires a db migration

This is the implementation for #1531 

Closes #1531


xamanu:
Thanks a lot for this PR. This helps to advance also on the #1590 issue that is related.

xamanu:
This is included and has continued work on top in #1858.
Closing here.

PR #1774, opened 2019-07-20
Handle random task choice for existing projects #1772

### LINKED ISSUE #1772: Project save fails with new release
Capture

Project save function fails with `{'enforceRandomTaskSelection': ConversionError(\"This field is required.\")}` error message.

**How to reproduce?**
- Choose any of the existing project created prior to this release
- Click edit project
- Now just try saving, you should get the error message

**What's the cause?**
- There's a new field `enforce_random_task_selection` in the projects table
- By default during migration all these values are populated to null on DB 
- Project DTO has the field requirement set to True, which means the application expects a non-null value during data transfer.
- However, all the existing projects have this field value set to null. So while saving this  throws an error

**Solution**
- Add a server default value in migration file
- And set the project DTO requirment to false

ramyaragupathy:
Fixes #1772


ramyaragupathy:
> but still can't save the project.

Can you please elaborate?

xamanu:
I get the same UI error as before. But no error in the log anymore.

ramyaragupathy:
If you don't see console errors, then the frontend shouldn't throw error as well

PR #1777, opened 2019-07-22
Update API Info text in stats section

ramyaragupathy:
image


xamanu:
Can we use this as a placeholder to replace quite a few description texts, within #1780?

ramyaragupathy:
Agreed

PR #1785, opened 2019-07-24
Add more autoscaling options (#1701)

dakotabenjamin:
This fix adds more options for autoscaling based on the production/staging environment and a new cloudformation parameter AutoscalingType. Set to `1-4` for smaller production sites and `3-12` for larger (i.e. the main one) sites. 

CircleCI has been updated with the parameters for TeachOSM, Assisted, and Production.


xamanu:
@dakotabenjamin what is the status of this?

PR #1788, opened 2019-07-25
update swagger ui

willemarcel:


xamanu:
The first error I reported is a local issue, not related to this PR.

PR #1790, opened 2019-07-25
Adds ml-enabler integration

fitoria:
This adds ml-enabler integration to TM. (continued from #1787 )

To test:

Update requirements

run looking-glass locally or update environment variables to point to a looking glass instance (see config.py)

Create a project on step 3 you'll see the "ML Layer" button, click it to generate a prediction, this will take a while;
Screen Shot 2019-07-24 at 11 06 38 PM

After the prediction finishes it should look like this:

Screen Shot 2019-07-24 at 11 09 19 PM

After that the project can be created and task annotations will be created too so project detail will have the "ML Layer" button on too.


server/services/ml_enabler_service.py
```diff
@@ -0,0 +1,104 @@
...
+class MLEnablerService:
+
+    @staticmethod
+    def get_prediction_from_bbox(model_name:str, bbox: str, zoom: int=18):
+        """Sends requests to create a new prediction for a bounding box from a model id"""
+        api_url = current_app.config['ML_ENABLER_URL']
+        model_id = get_model_id(api_url, model_name)
+        url = f'{api_url}/model/{model_id}/tiles' 
+        params = dict(bbox=bbox, zoom=zoom)
+        response = requests.get(url, params=params)
```
JorgeMartinezG: We should include after getting the response, a raise error depending on status. It is solved using _response.raise_for_status()_. This would raise an error depending on the status code received as response. Then, we catch that error with **requests.exceptions.RequestException** in _ml_enabler_apis.py_

server/services/ml_enabler_service.py
```diff
@@ -0,0 +1,104 @@
...
+            main_response = {'status': 'error', 'error': data.get('error')}
+
+        return main_response 
+
+
+    @staticmethod
+    def get_all_models():
+        """ Get all models on ml-enabler"""
+        url = f'{current_app.config["ML_ENABLER_URL"]}/model/all' 
+        return requests.get(url).json()
```
JorgeMartinezG: Same here

fitoria: updated it with your suggestions

geohacker:
Thank you for working in the integration @fitoria. This is looking really cool, especially the task splitting. Few thoughts as I go through the PR. 

* ml-enabler is built with the assumption that TM will be one of the consumers
* TM shouldn't have to integrate the models (like looking-glass). This is the core idea of ml-enabler, that it becomes the central registry of models and TM doesn't have to worry about it.
* In the current integration approach, we would have to integrate every new model into the TM codebase. This would get very quickly difficult to maintain. 
* [ml-enabler-cli](https://github.com/hotosm/ml-enabler-cli) is the glue between the models and API at the moment. This part would eventually get integrated to the ml-enabler API when we build container management.
* Asking the model for predictions shouldn't be part of the TM request/response cycle. Models can take unpredictable amount of time which means the current approach would result in problems.

I would suggest that TM integrate the ml-enabler API as a consumer, and focus on fetching the results as necessary. Something like:
1. TM can list all models registered on ml-enabler
2. TM can also filter the list based on the AOI and if the model has a prediction for that AOI
3. TM can fetch the predictions and store them in the annotations API
4. And use the information like rest of the PR.

I'd be happy to talk through this a bit more if necessary.

cc @xamanu @batpad @wronk

xamanu:
The ml-enabler integration should ideally also being able to use both available models from the ml-enabler (looking glass and MS building APIs). Can we add a good way to select the available ones for each region?

fitoria:
Removes backend prediction creation call
Added a selector to fetch the prediction from different sources on ml-enabler:

Screen Shot 2019-08-01 at 4 28 05 PM

Screen Shot 2019-08-01 at 4 28 09 PM

xamanu:
Thanks, this has been deployed on tasks-assisted. It's backend is ready to go into the main code base in PR #1919. The frontend has changed entirely and we need to look into this again with react now.

Issue #886, opened 2017-10-26
Tasking Manager does not load


OttoROSM:
When browsing to: _http://tasks.hotosm.org/_ the only thing that loads is the HOT Tasking Manager logo.
This happens on my laptop using several browsers:
- Chrome
- FireFox
- Internet Explorer
- Edge

I'm running the latest Java and Windows 10 fall update.
Via Chrome om IOS I have no issue in accessing the tasking manager.

OttoROSM:
I found out what is causing it. I am using Heimdal pro and this site blocked traffic to cdn.jsdelivr.net
After I switched Heimdal off, reloaded HOT and turned it on again, the site loads fine again.

ethan-nelson:
Do we have any interest in hosting the JS resources ourselves?

pgiraud:
@ethan-nelson I was also suprised recently to figure out that we are not hosting the resources ourselves. Though I understand the interest to use CDN's it prevented me from working on some development while I was offline.

verdy-p:
The problem is not the use of any CDN, but the current choice of CDN used, which is used by too many rogue spammers (most often for very intrusive ads fill of privacy issues in their scripts).

If you have nothing better, consider hosting the script elsewhere I even think that it would be fine to host these scripts in a Wikimedia user page or Wikimedia Labs host: at least there, the hosting envioronment has strong poilicies about respecting user privacy and security, and there's little chance it will be blocked. However on an Wikipedia user page, there's the risk of being blocked in some countries that don't like contents hosted on Wikipedia (notably being blocked in China by their national infrastructure based on mandatory "transparent proxies" applied by local ISPs).

There's much less risks being blocked by third parties (notably antispam/animalwares and other security tools) if they are hosted the Wikimedia Labs farm (but it requires an active project to be maintained there and some presence to reply to incoming issues, otherwise the Wikimedia Labs-hosted project may be closed administratively: make sure you correctly describe the project, and have suitable licences and you keep it open, there's no reason Wikimedia will block scripts used by OSM task managers when we have good cooperation with them). Wikimedia is very active at fighting abuses and solving any issue that could happen if one hosted project is harvested by malicious third parties, spamwares, spywares, virus hosting, and so on: it can clean up these projects rapidely and make them live again very fast, thanks to its very active local community of maintainers and admins.

[...]

PR #1795, opened 2019-07-26
Switch CDN from jsdelivr to Cloudfront

### LINKED ISSUE #912: Host javascript resources ourselves
Feature request from #886 discussion: host javascript resources ourselves instead of CDN.

eternaltyro:
Experimentally switch to Cloudfront the resources hosted in jsdelivr. Fixes #912  and addresses #886

### DIFF
--- a/client/index.html
+++ b/client/index.html
@@ -33,11 +33,11 @@
     <link href="node_modules/openlayers/dist/ol-debug.css" rel="stylesheet">
     <!-- /build -->
 
-    <!-- build:css https://cdn.jsdelivr.net/openlayers.geocoder/latest/ol3-geocoder.min.css -->
+    <!-- build:css https://cdn.hotosm.org/js/ol3-geocoder.js -->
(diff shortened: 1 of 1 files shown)


xamanu:
This is targeting the AngularJS code, which will be removed from `develop` anyways. Can you please file this PR against `master`, then we will include it into the next hotfix release. Thank you!

eternaltyro:
Thanks @xamanu @ramyaragupathy .

I raised #1798 in place of this. I simply cherry-picked the jsdelivr changes. If that looks okay, this can be closed.

xamanu:
Doubleposting my comment from #912:

> The downside of just using HOT's CDN is that all instances will use these resources. Ideally a TM should be able to live on it's own and without connections to anywhere (thinking about a use case inside of POSM, ...).
> 
> Probably we want to make this configurable for big instances to use any CDN they want to chose. And fallback to a local serving of the libraries if not configured otherwise.

PR #1800, opened 2019-07-29
header improvements

willemarcel:
Add border-bottom to active header nav link and load OSM user profile picture and store it on the global state


frontend/src/components/header/index.js
```diff
@@ -21,17 +22,46 @@ const menuItems = [
...
   renderMenuItems() {
     return(
       <div className="v-mid">
         {this.menuItems.map((item, n) =>
-          <Link to={item.link} key={n} className={ this.linkCombo }>
-            <FormattedMessage {...item.label} />
-          </Link>
+          <ActiveNavLink to={item.link} key={n} isActive={this.isActive}>
+            <FormattedMessage {...item.label}/>
+          </ActiveNavLink>
```
thadk: See above if you agree:
```suggestion
          </TopNavLink>
```

server/services/users/authentication_service.py
```diff
@@ -68,6 +68,11 @@ def login_user(osm_user_details, redirect_to, user_element='user') -> str:
 
         osm_id = int(osm_user.attrib['id'])
         username = osm_user.attrib['display_name']
+        try:
+            # get gravatar profile picture file name
+            user_picture = osm_user.find('img').attrib['href'].split('/avatar/')[1].split('?s=')[0]
```
thadk: **You can mostly ignore the below, I realized that I just need to redo my Docker instance to enable the new API:**

This didn't seem to work for my user: 
This code: image
OSM:
image
3.4 TM:
image

I did see one entry in Network tab that looks like it found my gravatar image but I was not able to reproduce it.

Here's the element on OSM.org:
``

Using the dispatcher in Redux dev tools:
```
{
type: 'SET_PICTURE',
userPicture:'fcf1c2b12340d8be939430bcd011e15d.jpg?s=100&amp;d=https%3A%2F%2Fwww.openstreetmap.org%2Fassets%2Favatar_large-afe7442b856c223cca92b1a16d96a3266ec0c86cac8031269e90ef93562adb72.png'
}
``` 
resulted in:
image

and here's the tag for what I'm seeing with this code:
https://gist.github.com/thadk/79a9b26d77285cc2b2ab6a9dcd2baeaa

frontend/src/components/header/index.js
```diff
@@ -21,17 +22,46 @@ const menuItems = [
   {label: messages.help, link: "help"}
 ];
 
+
+const ActiveNavLink = (props) => {
```
thadk:
```suggestion
const TopNavLink = (props) => {
```
if you agree, see below

frontend/src/components/button.js
```diff
@@ -14,3 +14,17 @@ export function Button({ onClick, children, className }: Object) {
     </button>
   );
 }
+
+
+export function CustomButton({ onClick, children, className }: Object) {
+  return (
+    <button
+      onClick={onClick}
+      aria-pressed="false"
```
thadk: great to see aria – should we try to do accessibility broadly?

frontend/src/components/header/index.js
```diff
@@ -7,6 +7,7 @@ import { FormattedMessage } from 'react-intl';
 import messages from './messages';
 import { ORG_URL, ORG_NAME, API_URL } from '../../config';
 import logo from '../../assets/img/main-logo.svg';
+import profilePic from '../../assets/img/user.jpg';
```
thadk: Ah hah, the default.

Issue #1805, opened 2019-07-30
Switch to new frontend


xamanu:
In order to focus on the development of the redesign and transition to the new frontend #1757 the code base should build to new frontend, the ci run it and the documentation mention it.

PR #1807, opened 2019-07-30
Update and cleanup infrastructure components

xamanu:
* Moved `devops` to a more generic `scripts` directory.
* Moved `cloudformation` and `infra` (now `packer`) under `scripts/aws`
* Updated some old nodejs versions.
* Updated Docker to rely on postgresql 11 and Debian stable (buster)
* Removed `package.json` on root directory
* Removed install script for oldstable Debian Stretch.


xamanu:
@dakotabenjamin and @eternaltyro feel free to check and comment your thoughts. I still need to do some testing on the Docker side of things before asking for a formal review. Thanks.

xamanu:
Tested the Docker stuff today. Works fine. Removing "Work in Progress".

scripts/aws/cloudformation/tasking-manager.template.js
```diff
@@ -169,9 +169,9 @@ const Resources = {
         'sudo apt-get -y install python3.6-dev',
         'sudo apt-get -y install python3.6-venv',
         'sudo apt-get -y install curl',
-        'curl -sL https://deb.nodesource.com/setup_6.x > install-node6.sh',
-        'sudo chmod +x install-node6.sh',
-        'sudo ./install-node6.sh',
+        'curl -sL https://deb.nodesource.com/setup_10.x > install-node10.sh',
```
eternaltyro: can we do `curl -o install-node10.sh -sL https://deb.nodesource.com/setup_10.x` . Better to use this than shell redirection.

Alternatively, to make this idempotent, we can use a temporary file, like so:
```
INSTALL=$(mktemp)
curl -o ${INSTALL} -sL https://deb.nodesource.com/setup_10.x
chmod a+x ${INSTALL}
sudo ${INSTALL}
```

Depending on the install situation, this may or may not help.

scripts/docker/tasking-manager/Dockerfile
```diff
@@ -24,7 +24,7 @@ RUN pip install --no-cache-dir -r requirements.txt
 ## CLIENT CONFIGURATION
 
 # Install nodejs.
-RUN curl -sL https://deb.nodesource.com/setup_9.x | bash - \
+RUN curl -sL https://deb.nodesource.com/setup_10.x | bash - \
```
eternaltyro: Buster has nodejs in its package list. We may not need this if we do `apt install nodejs` 

https://packages.debian.org/buster/nodejs

xamanu: Oh, right. Very good. Removed this entirely on the buster images. Thanks.

scripts/aws/cloudformation/tasking-manager.template.js
xamanu: Cool. Changed as suggested.

xamanu:
By the way, now the (frontend) tests are failing, and i'm a bit lazy to look into them, because we are changing the frontend anyway in this PR #1805, in which this commit here has been included. With the new frontend tests run through successfully.

xamanu:
Thanks for your review, @eternaltyro. Will close here and move on to the PR #1812 for @dakotabenjamin to check on the deployment for the new frontend.

PR #1808, opened 2019-07-31
Fix area query

ramyaragupathy:
Updated query for
- [Home page stats](https://tasks.hotosm.org/api-docs/swagger-ui/index.html?url=https://tasks.hotosm.org/api/docs#!/stats/get_api_v1_stats_summary)
- [Proejct stats](https://tasks.hotosm.org/api-docs/swagger-ui/index.html?url=https://tasks.hotosm.org/api/docs#!/stats/get_api_v1_stats_project_project_id)

All areas are returned in `sq.km`

### DIFF
--- a/server/models/postgis/project.py
+++ b/server/models/postgis/project.py
@@ -468,17 +468,10 @@ class Project(db.Model):
         """ Create Project Summary model for postgis project object"""
         project_stats = ProjectStatsDTO()
         project_stats.project_id = self.id
-        polygon = to_shape(self.geometry)
-        polygon_aea = transform(
-            partial(
-                pyproj.transform,
-                pyproj.Proj(init="EPSG:4326"),
-                pyproj.Proj(proj="aea", lat1=polygon.bounds[1], lat2=polygon.bounds[3]),
-            ),
-            polygon,
-        )
-        area = polygon_aea.area / 1000000
-        project_stats.area = area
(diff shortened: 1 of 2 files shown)


server/models/postgis/project.py
```diff
@@ -468,17 +468,10 @@ def get_project_stats(self) -> ProjectStatsDTO:
...
-            partial(
-                pyproj.transform,
-                pyproj.Proj(init="EPSG:4326"),
-                pyproj.Proj(proj="aea", lat1=polygon.bounds[1], lat2=polygon.bounds[3]),
-            ),
-            polygon,
-        )
-        area = polygon_aea.area / 1000000
-        project_stats.area = area
+        project_area_sql = "select ST_Area(geometry, true)/1000000 as area from public.projects where id = :id"
```
JorgeMartinezG: @ramyaragupathy I think we would need to cast the geom column as geography type in order to get the area in squared meters. Please check the third example from https://postgis.net/docs/ST_Area.html

ramyaragupathy: @JorgeMartinezG - I tried the examples with the power value mentioned in the documentation. The one in the PR now `ST_Area(geometry, true)/1000000` seems to be matching with the current turf value returned

xamanu: To me it seems accurate. The example @JorgeMartinezG was referring to seems to be in a different projection (EPSG:2249) in our case this query should return m2 or km2 after the division.

PR #1819, opened 2019-08-06
Featured projects fix

willemarcel:


thadk:
oops just saw this. It looks good. I also added better loading cards since the wait time on staging at least can be quite long.

PR #1812, opened 2019-07-31
Switch to react frontend

xamanu:
This PR includes #1807  and fixes #1805.

* Removed `client` directory including the old frontend based on AngularJS.
* Adjusted the server to load on the new frontend instead.
* Updated the Docker setup to use the new frontend
* Updated documentation to reflect the new frontend
* Updated deployment scripts to build and serve the new frontend (couldn't test it myself)

PR #1820, opened 2019-08-07
Improve frontend loading

xamanu:
This PR attempts to fix the issues risen in #1812


thadk:
I found that the documentation recommends using their node  package serve
It also gives a workaround for this issue we are having (I think) https://facebook.github.io/create-react-app/docs/deployment#customizing-environment-variables-for-arbitrary-build-environments

Here is my attempt so far without yet adding that `env-cmd` advice: https://github.com/thadk/tasking-manager/commit/4900582394f1ae5556a76d3964da362af1cafe80

Here are two more older solutions that seem to replicate the netlify feature somehow:
https://github.com/inloop/cra-docker (simple: use jo to write certain env contents to a JSON file)
Maybe too complicated: https://www.freecodecamp.org/news/how-to-implement-runtime-environment-variables-with-create-react-app-docker-and-nginx-7f9d42a91d70/ (edited)

xamanu:
Thanks @thadk and @willemarcel for reviews and help!
Now we are sticking to the `REACT_APP_*` rule. `env-cmd` made quite a lot of problems. Piping now the environment variables through a bash command for builds and starts.

frontend/src/config/index.js
```diff
@@ -1,39 +1,23 @@
...
-
-
-// DEFAULT locale
-export const DEFAULT_LOCALE = process.env.TM_DEFAULT_LOCALE || 'en';
-
-
-// PROJECT CARD CONFIG
-export const PROJECTCARD_CONTRIBUTION_SHOWN_THRESHOLD = process.env.PROJECTCARD_CONTRIBUTION_SHOWN_THRESHOLD || 5;
\ No newline at end of file
+var dotenv = require('dotenv')
```
willemarcel: let's use ES6 import http://es6-features.org/#ValueExportImport

thadk: I think this will work:
```javascript
import dotenv from 'dotenv'
import dotenvExpand from 'dotenv-expand';
```
https://github.com/motdotla/dotenv-expand/issues/19

thadk: Oh, I see dotenv not even used explicitly in the final version I think.

willemarcel:
It worked well when I served the frontend with npm/yarn.

When I built the frontend and run only the flask server, all requests to the API were redirected to the frontend.

xamanu:
Wasn't able to reproduce the problems you are facing @willemarcel. Let's connect today and make sure it works.

xamanu:
Checked it today with @willemarcel and after some rounds of debugging it is open for review again. Thanks.

PR #1822, opened 2019-08-12
API version 2 - Step one: Identifiers & Methods

ramyaragupathy:
* API updates for:
     - [x] systems
     - [x] projects
     - [x] users
     - [x] tasks
     - [x] annotation
     - [x] organisations
* Init file updates
* JSON serialized names review
* Error messages review
* Trailing backslash
* HTTP verbs


xamanu:
I think this is getting ready for review, @ramyaragupathy, @willemarcel and @JorgeMartinezG. Please have a look. I was able to successfully test it locally. But the changes in the endpoints require me to force empty cache reload on every page. Please make sure you are checking without any caching happening.

willemarcel:
@ramyaragupathy Let's add the trailing slash on all endpoints?
Why do we need the `preferred_locale` and why it's mandatory on project search? In my opinion, the API doesn't need to deal with internationalization.

ramyaragupathy:
I've updated the error messages and reviewed serialized names. For some reason server requirements installation fails at `psycopg` . Please let me know if this gets through fine for others.

willemarcel:
> Why do we need the preferred_locale and why it's mandatory on project search? In my opinion, the API doesn't need to deal with internationalization.

Probably to return the project descriptions and other informations only in one language. So it's ok to keep it.

willemarcel:
I think GET Method on `/api/v2/users/` should be restricted to admin or project managers users.

ramyaragupathy:
> I've updated the error messages and reviewed serialized names. For some reason server requirements installation fails at `psycopg` . Please let me know if this gets through fine for others.

Fixed through `sudo installer -pkg /Library/Developer/CommandLineTools/Packages/macOS_SDK_headers_for_macOS_10.14.pkg -target /` - [OS specific issue](https://github.com/python-pillow/Pillow/issues/3438#issuecomment-435169249)

xamanu:
> Probably to return the project descriptions and other informations only in one language. So it's ok to keep it.

Yes, in this step, let's not improve on the logic. It is already becoming quite a beast.

> I think GET Method on /api/v2/users/ should be restricted to admin or project managers users.

Yes, I agree, but let's do this in a next step. As long as it wasn't restricted before.

ramyaragupathy:
Per chat addressing the following through another PR:

* GDPR review
* Data units review
* Profiling each endpoint
* test case coverage
* New functionalities
    * [ ] Project Functions - nogeometries,  no tasks
    *  [ ] task functions - list comments for a specific task
*  Head operation support in swagger

cc @xamanu @JorgeMartinezG @fitoria @willemarcel

PR #1832, opened 2019-08-15
add location prop to header and router

willemarcel:


frontend/src/views/settings.js
```diff
@@ -0,0 +1,32 @@
...
+
+
+export function Settings() {
+  const userIsloggedIn = useSelector(state => state.auth.get('token'));
+  const userDetails = useSelector(state => state.auth.get('userDetails'));
+
+  if (userIsloggedIn) {
+    return (
+      <div className="pull-center">
+        <UserTopBar />
```
thadk: If it is possible the user sees a page like this immediately after logging in, I think maybe we should wrap things like these in a loading component like `<React-placeholder ready={redux-says-it-is-loaded}>` so it doesn't flash from "Mapper" to your user name and then flash again to add the changeset count. Maybe also the input fields in personal information: on slow connections they could lose their work users start typing.
settings-animation
When you refresh, sometimes the user image even shows as a strange shape while it loads.

(also I clearly need to map with smaller changesets so I can get to a higher level 😊 than beginner )

frontend/src/views/welcome.js
```diff
@@ -1,30 +1,56 @@
...
-        <div className="cf pa4 bg-tan">
-          <div className="fl w-100 w-60-l pb3 pr3-l">
-            <UserInterests />
-            <HelpCard />
-          </div>
-          <div className="fl w-100 w-40-l pb3 pl3-l">
-            <UserInformationForm />
-          </div>
-        </div>
+        {completeness <= 0.5 ? <IncompleteProfile />
```
thadk: See wrapping with `<react-placeholder>` comment above.

frontend/src/views/settings.js
willemarcel: I added the placeholders. On the form, I decided to keep it disabled until the data is loading.

willemarcel: I solved the image problem by setting the width.

willemarcel:
@thadk I also added some new components to Settings view, but it's not fully working because I would like to use the switch toogle you added on your PR and I would like to wait for the update user details endpoint to support PATCH requests. I think we could merge after your review and then I come back working on this.

I made a change in the auth actions to execute logout if the request to the user-detail api endpoint fails.

frontend/src/App.js
```diff
@@ -9,23 +9,31 @@ import { AboutPage } from './views/about';
...
-          <Home path="/" />
-          <AboutPage path="about" />
-          <Authorized path="authorized" />
-          <Login path="login" />
-          <Welcome path="welcome" />
-        </Router>
-      </div>
-      <Footer />
-    </div>
+    <Location>
```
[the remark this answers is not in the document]
thadk: We can go with it like this if you want. Just so I was clear, the alternative i was suggesting was having 2 routers: 

The first router was just for the header. The second one is like what we have now: for the rest of the content. 
Then you should be able to use the automatic prop that contains the location in every <Route/> and you won't need to wrap the whole thing just for the Header class to get that <Route/> prop.

frontend/src/views/welcome.js
thadk: This one was jumping from 0->90% too as the data loaded. Maybe you fixed it somehow, when I get a chance to try this tomorrow I will check it.

PR #1837, opened 2019-08-20
Featured projects

### LINKED ISSUE #1836: Featured projects
Enable a project to be featured by Project Managers. The following tasks are:
- Update database model to set a project as featured.
- Create API endpoints to list, set and unset projects as featured. Returning response must have the same structure as project search Endpoint
- Create services.
- Create tests for these services.

JorgeMartinezG:
This PR closes #1836 . The following endpoints were included:

`GET /api/v1/project/feature`To get list of featured projects.
`POST /api/v1/project/feature/<project_id>` which sets a project as featured
`DELETE /api/v1/project/feature/<project_id>` which unsets a project as featured

**How to test**:
1. Login to TM and update your credentials as PM or admin.
2. Create two or more projects.
3. Set a project as featured.
4. List all featured projects.


thadk:
Hmm, is this expected behavior? I think the GET endpoint should be open to all.
image

JorgeMartinezG:
You are right! We should keep same consistency with project `/search` endpoint. I've just added a commit

thadk:
The GET query works well now and works as needed! Thanks @JorgeMartinezG

thadk:
@JorgeMartinezG Can you maybe add an index to this if you haven't so it's speed is optimized too please since it shows on the front page?

willemarcel:
Thanks @JorgeMartinezG !
I tested it and it worked well. I just rebased with the current develop and added a commit with a migration update.
I would like to change the urls to match the API v2 plans, I suggest:
* `/projects/queries/featured` for the list view
* PUT to `/projects/<id>/actions/feature` to feature
* PUT to `/projects/<id>/actions/reset-feature` to remove it from your featured projects

While we don't finish the API v2, I think we could continue using the singular form `/project/`, but I think the final part of the URL can already follow the new pattern. Before doing the change, wait for the opinion of @xamanu and @ramyaragupathy .

xamanu:
I agree, @willemarcel, this will be helpful.

xamanu:
Just rerolled on latest develop and followed the suggestions of @willemarcel and further prepared this also in the structure for the new API version 2.

xamanu:
Rerolled on APIv2 and added the following endpoints:

* GET: `/api/v2/projects/queries/featured`
* POST: `/api/v2/projects/<int:project_id>/actions/feature`
* POST: `/api/v2/projects/<int:project_id>/actions/remove-feature`

server/api/projects/actions.py
```diff
@@ -126,3 +127,103 @@ def post(self, project_id):
             error_msg = f"Send message all - unhandled error: {str(e)}"
             current_app.logger.critical(error_msg)
             return {"Error": "Unable to send messages to mappers"}, 500
+
+
+class ProjectsActionsFeatureAPI(Resource):
+    @tm.pm_only(True)
+    @token_auth.login_required
+    def post(self, project_id):
```
ramyaragupathy: Based on AP redesign discussion - this should be a PATCH request

server/api/projects/actions.py
```diff
@@ -126,3 +127,103 @@ def post(self, project_id):
...
+        try:
+            ProjectService.set_project_as_featured(project_id)
+            return {"Success": True}, 200
+        except NotFound:
+            return {"Error": "Project Not Found"}, 404
+        except ValueError as e:
+            error_msg = f"FeaturedProjects POST: {str(e)}"
+            return {"error": error_msg}, 400
+        except Exception as e:
+            error_msg = f"FeaturedProjects GET - unhandled error: {str(e)}"
```
ramyaragupathy: Let's not show too much information in our error messages @JorgeMartinezG - what do you think? Just using `str(e)` reveals too much about the backend

server/api/projects/actions.py
```diff
@@ -126,3 +127,103 @@ def post(self, project_id):
...
+            current_app.logger.critical(error_msg)
+            return {"error": error_msg}, 500
+
+
+class ProjectsActionsUnFeatureAPI(Resource):
+    @tm.pm_only(True)
+    @token_auth.login_required
+    def post(self, project_id):
```
ramyaragupathy: Feel this must be a PATCH to projects

server/models/postgis/project.py
```diff
@@ -90,6 +90,9 @@ class Project(db.Model):
     )  # Force users to edit at random to avoid mapping "easy" tasks
     allow_non_beginners = db.Column(db.Boolean, default=False)
     private = db.Column(db.Boolean, default=False)  # Only allowed users can validate
+    featured = db.Column(
+        db.Boolean, default=False
+    )  # Only PMs can set a project as featured
```
ramyaragupathy: Current changes lets anyone update a project's `featured` state. Should we include a role check in `set_as_featured` and `unset_as_featured` @JorgeMartinezG

server/api/projects/actions.py
xamanu: Have we agreed on a consistent way of calling `actions` endpoints? They are different than the REST ones. I would propose to use `POST` for all of them. We also seem to use `PUT` on some others. A well defined way of selecting it, would be good.

server/api/projects/actions.py
xamanu: Same convo as above with the POST on an `actions` endpoint.

server/models/postgis/project.py
JorgeMartinezG: Could you please elaborate more on your comment?

PR #1684, opened 2019-06-07
Feature/user tasks api view

fitoria:
fixes #1597 

How to test:

Authenticated ajax request to /api/v1/user/tasks

Filters as get params, available filters: status, project_id, min_action_date, max_action_date

Sample request url:  ```/api/v1/user/tasks?status=READY&project_id=2&min_action_date=2019-04-9&max_action_date=2019-04-10```

Sample response:

```
{
    "tasks": [
        {
            "taskId": 9,
            "projectId": 1,
            "taskStatus": "READY",
            "taskHistory": [],
            "taskAnnotation": null,
            "isUndoable": false,
            "autoUnlockSeconds": 7200
        },
        {
            "taskId": 16,
            "projectId": 1,
            "taskStatus": "READY",
            "taskHistory": [],
            "taskAnnotation": null,
            "isUndoable": false,
            "autoUnlockSeconds": 7200
        }
    ]
}
```

Issue #1599, opened 2019-05-24
Create “My tasks” frontend page


willemarcel:
Implement frontend view that show the tasks a user has touched as a mapper or as a validator.

Reference: https://projects.invisionapp.com/share/VCQTFR6YSJR#/screens/359643157 

Depends on https://github.com/hotosm/tasking-manager/issues/1597

thadk:
We might want pagination on this API similar to messages, projects, etc.

thadk:
Remember hover doesn't work on iPads and phones, but this only appears on hover: <img width="976" alt="image" src="https://user-images.githubusercontent.com/283343/71355709-e0a76e00-254d-11ea-8bbf-edd168f204f4.png">

PR #1848, opened 2019-08-29
Endpoint for my tasks

### LINKED ISSUE #1597: Create user tasks API view
Implement API view that lists the tasks of all projects mapped or validated by a user. Allow to apply filters by date, project id, mapping/validating, plus ordering options.

fitoria:
fixes #1597 was PR #1684

How to test:

Authenticated ajax request to /api/v1/user/tasks

Filters as get params, available filters: status, project_id, min_action_date, max_action_date

Sample request url: /api/v1/user/tasks?status=READY&project_id=2&min_action_date=2019-04-9&max_action_date=2019-04-10

Sample response:
```
{
    "tasks": [
        {
            "taskId": 9,
            "projectId": 1,
            "taskStatus": "READY",
            "taskHistory": [],
            "taskAnnotation": null,
            "isUndoable": false,
            "autoUnlockSeconds": 7200
        },
        {
            "taskId": 16,
            "projectId": 1,
            "taskStatus": "READY",
            "taskHistory": [],
            "taskAnnotation": null,
            "isUndoable": false,
            "autoUnlockSeconds": 7200
        }
    ]
}
```

### DIFF
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -644,7 +644,7 @@ def add_api_endpoints(app):
 
     # Users REST endpoint
     api.add_resource(UsersAllAPI, "/api/v2/users/")
-    api.add_resource(UsersRestAPI, "/api/v2/users/<int:userid>/")
+    api.add_resource(UsersRestAPI, "/api/v2/users/<int:user_id>/")
     api.add_resource(
         UsersQueriesUsernameFilterAPI, "/api/v2/users/queries/filter/<string:username>/"
     )
@@ -669,7 +669,7 @@ def add_api_endpoints(app):
         "/api/v2/users/<string:username>/actions/set-expert-mode/<string:is_expert>/",
     )
 
-    api.add_resource(UsersTasksAPI, "/api/v2/users/<string:username>/tasks/")
+    api.add_resource(UsersTasksAPI, "/api/v2/users/<int:user_id>/tasks/")
     api.add_resource(
         UsersActionsVerifyEmailAPI, "/api/v2/users/me/actions/verify-email/"
     )
--- a/server/__init__.py
+++ b/server/__init__.py
@@ -242,6 +242,7 @@ def add_api_endpoints(app):
         UserRecommendedProjectsAPI,
         UserInterestsAPI,
     )
+    from server.api.users.tasks import UsersTasksAPI
     from server.api.users.actions import (
         UsersActionsSetUsersAPI,
         UsersActionsSetLevelAPI,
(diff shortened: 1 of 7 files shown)


willemarcel:
Thank you very much, @fitoria! Sorry for the delay on reviewing it.

- Instead of min_action_date and max_action_date, we could use `start_date` and `end_date`.
- It would be good if we could sort by action_date
- We could also include the `action_date` field on the response

server/__init__.py
```diff
@@ -476,6 +477,7 @@ def add_api_endpoints(app):
     api.add_resource(
         UsersQueriesUsernameAPI, "/api/v2/users/queries/<string:username>/"
     )
+    api.add_resource(UserTasksAPI, "/api/v2/user/tasks/")
```
xamanu: Should probably be `/api/v2/users/me/tasks/`.

server/api/users/resources.py
```diff
@@ -193,3 +194,90 @@ def get(self, username):
             error_msg = f"User GET - unhandled error: {str(e)}"
             current_app.logger.critical(error_msg)
             return {"Error": "Unable to fetch matching users"}, 500
+
+
+class UserTasksAPI(Resource):
```
xamanu: Should be `UsersTasksAPI` and be in a `tasks.py`

server/__init__.py
```diff
@@ -215,6 +215,7 @@ def add_api_endpoints(app):
         UsersAllAPI,
         UsersQueriesUsernameAPI,
         UsersQueriesUsernameFilterAPI,
+        UserTasksAPI,
```
xamanu: Should be `UsersTasksAPI` and live in `server/api/users/tasks.py`

xamanu:
This is the base for #1599

xamanu:
As per request of @willemarcel, please extend to filter on `taskStatus` (e.g. = `locked`).

willemarcel:
Thanks, @fitoria ! 

"taskHistory", "isUndoable", "autoUnlockSeconds" are not necessary, but it's not a big problem to have them. Although `taskHistory` always returned an empty array for me.

Do you think it would be complex if we add the user of the last task action? If it is complex or it requires too much from the database, I prefer to not having them.

xamanu:
There are also some conflicts, please rebase on latest `develop`. Thanks.

willemarcel:
I just saw that this PR is also related to https://github.com/hotosm/tasking-manager/issues/1597, so maybe we could change the url from `/users/me/tasks` to `users/:username/tasks`, so it will handle the two needs that we have

xamanu:
Shouldn't we use the `user_id` in the good old Restful way?

willemarcel:
Thank you, @fitoria! It seems good to me.

willemarcel:
I changed the URL param to user_id instead of username, @xamanu

PR #1852, opened 2019-09-02
Fix api path loading on localhost

xamanu:
It was reported that with the new frontend the backend api call weren't always successful. While I had a difficult time reproducing it, I was able to do so in a Chromium. API calls aren't successful, but when force uncached reload with CTRL+SHIFTR+R (or SHIFT+F5) it worked.

This if-clause for localhosts is causing the problem. Taking it out here. But not completely sure whether we will need this or not.


willemarcel:
It worked well in Chrome. In Firefox I only was able to login when I went to a private tab. I tried to clear history, cookies, cache and localStorage of all ways possible... I have the hope it will work better on non local environments.

Issue #1849, opened 2019-08-30
Project Search API is slow for large resultsets


thadk:
When using the new front-end, the `project/search` API is fairly slow on the staging server, taking up to 15 seconds to return the default results (~2500 projects) for no specific filters.  If multiple people are using the API it seems to get even slower and can time out and show bad gateway. If you filter to a smaller set, such as the Intermediate (~200)  or Advanced (~30 projects), the API returns much more quickly.

It could be related to the query JOIN of all projects with their task history to get the number of distinct`total_contributor`'s to each project. This is hard to do efficiently in one query because we cannot look at *only* the one requested page of results: The map needs to show all the projects returned for browsing.  @JorgeMartinezG pointed to this line: https://github.com/hotosm/tasking-manager/blob/develop/server/services/project_search_service.py#L70

If possible in the API, maybe we can do two queries in one API endpoint: 

* one for `mapResults` that only gets the ID and centroid with no `LIMIT` covering every project, 
* and another query for `results` that does the join with the task history but is only covering the requested page of results. 

This change would allow the front-end to use the existing implementation. As a relatively straightforward optimization, it could ask the API to turn off the first `mapResults` SQL query and output if the front-end already has downloaded the `mapResults[]` GeoJSON for those search parameters on future `project/search` API requests for later pages. Doing this in one API also saves bandwidth without adding new requests for the client to make.

Overall, maybe we can research where the bottleneck is on this API for large results.

JorgeMartinezG:
@xamanu this issue should be closed since it is included within #1854

PR #1854, opened 2019-09-02
Perform taskhistory join only within paginated results query

JorgeMartinezG:
This PR optimizes the project search endpoint stated within issue #1849 . The taskhistory join table is executed only for the paginated results query.

### DIFF
--- a/server/services/project_search_service.py
+++ b/server/services/project_search_service.py
@@ -56,33 +56,28 @@ class BBoxTooBigError(Exception):
 class ProjectSearchService:
     @staticmethod
     def create_search_query():
-        query = (
-            db.session.query(
-                Project.id.label("id"),
-                Project.mapper_level,
-                Project.priority,
-                Project.default_locale,
-                Project.centroid.ST_AsGeoJSON().label("centroid"),
-                Project.organisation_tag,
-                Project.campaign_tag,
-                Project.tasks_bad_imagery,
-                Project.tasks_mapped,
-                Project.tasks_validated,
-                Project.status,
-                Project.total_tasks,
-                Project.last_updated,
-                Project.due_date,
-                Project.country,
-                func.count(distinct(TaskHistory.user_id)).label("total_contributors"),
-            )
-            .outerjoin(TaskHistory, TaskHistory.project_id == Project.id)
-            .filter(Project.private is not True)
-        )
+        query = db.session.query(
+            Project.id.label("id"),
+            Project.mapper_level,
+            Project.priority,
+            Project.default_locale,
+            Project.centroid.ST_AsGeoJSON().label("centroid"),
+            Project.organisation_tag,
+            Project.campaign_tag,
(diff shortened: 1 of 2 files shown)


thadk:
I wasn't sure how to cleanly rebase this with the order-by lines, so you'll have to do that at some point @JorgeMartinezG

thadk:
Also it would be helpful if you could mention the before/after time of the endpoint on a big database. So far with my testing the time is still 38seconds which makes it seem the issue is still here with this set up.

JorgeMartinezG:
@thadk Could you please check again? I found the bottleneck. What is happening is that the join query occurs for all projects and then, the SQLAlchemy paginate filters, so no optimization was running. The solution is to remove the query, and using only the paginated results, create a database query to get the total contributors.

The response time was reduced within my local computer from 12 secs to 300 ms.

thadk:
This seems quite a bit better! Thanks. My first default query still took 21,000ms though for some reason: maybe it needed to read it from disk? Next, the mapperLevel=beginner query took only 1,250ms, and mapperLevel=1714ms. After the first time mentioned above, sorting the default query takes only about 650ms. Using the full-text search takes about 400ms.

thadk:
Would it save any more time if you `SELECT`ed fewer columns in the initial map query and then joined in the others for the paginated result?

thadk:
@JorgeMartinezG can you please also check if the TaskHistory/`totalContributers` calculation of the `featured` API could be improved in the same way. It might also be slower than needed. We only want the history for the few matching featured projects.

thadk:
aw, sorry @JorgeM looks like we had both done the `country` migration, but it hasn't hit staging yet, so this is crashing: https://github.com/hotosm/tasking-manager/pull/1854
```json

{"error": "FeaturedProjects GET - unhandled error: (psycopg2.errors.UndefinedColumn) column 
projects.country does not exist\nLINE 1: ..._updated, projects.due_date AS projects_due_date, projects.c...\n                                                             ^\n\n[SQL: SELECT projects.id AS id, 
projects.mapper_level AS projects_mapper_level, projects.priority AS projects_priority, 
projects.default_locale AS projects_default_locale, ST_AsGeoJSON(projects.centroid) AS centroid, 
projects.organisation_tag AS projects_organisation_tag, projects.campaign_tag AS 
projects_campaign_tag, projects.tasks_bad_imagery AS projects_tasks_bad_imagery, 
projects.tasks_mapped AS projects_tasks_mapped, projects.tasks_validated AS 
projects_tasks_validated, projects.status AS projects_status, projects.total_tasks AS 
projects_total_tasks, projects.last_updated AS projects_last_updated, projects.due_date AS 
projects_due_date, projects.country AS projects_country \nFROM projects \nWHERE 
projects.featured = true GROUP BY projects.id]\n(Background on this error at: http://sqlalche.me/e/f405)"}
```

PR #1857, opened 2019-09-04
PATCH updates for User Profile

willemarcel:
PATCH updates for user profile should do just partial updates of select fields, instead of the entire row. This PR implements the same


server/models/postgis/user.py
```diff
@@ -88,28 +88,10 @@ def update_username(self, username: str):
...
-        self.country = user_dto.country
-        self.name = user_dto.name
-        self.validation_message = user_dto.validation_message
-        self.default_editor = user_dto.default_editor
-        self.mentions_notifications = user_dto.mentions_notifications
-        self.comments_notifications = user_dto.comments_notifications
-        self.projects_notifications = user_dto.projects_notifications
-        self.expert_mode = user_dto.expert_mode
+        for attr, value in user_dto.items():
+            print(attr, value)
```
willemarcel: let's remove that print command

PR #1858, opened 2019-09-05
Backend for teams, organisations and campaigns

xamanu:
Backend for https://github.com/hotosm/tasking-manager/issues/1531

* Introduce teams
* Extended organisations and campaigns from tags to full objects
* Migration paths for organisations and campaigns


migrations/versions/0eee8c1abd3a_.py
```diff
@@ -55,8 +55,9 @@ def upgrade():
             if is_match:
                 project_continent = continent["properties"]["CONTINENT"]
                 with open(
-                    "migrations/" + project_continent + ".json", "r", encoding="utf-8"
-                    "scripts/world/" + project_continent + ".json"
+                    "migrations/" + project_continent + ".json",
```
xamanu: I think this is also in `scripts/world/`

xamanu:
Just checked on it again and saw one little type, I think. There is also a tiny conflict. @fitoria can you reroll this please?

On the side of revision, I feel like I did too much on the code, and I'd prefer @ramyaragupathy to review it.

xamanu:
Restored commit history. Ready to be merged in.

Issue #1111, opened 2018-05-23
Create templates for project instructions


geohacker:
It'd be great if we can support templates for project instructions that people can edit/fill in when they create new projects - this will make sure all the necessary instructions are captured, and people wouldn't have to copy/paste or try to come up with new text all the time.

cc @bhousel @geomantic @mcawley

bgirardot:
Hi,

We have something like this now, the text at least has been worked out and translated.

It is more complicated that it seems from a UI perspective, because the point is to have a good, well written set of instructions translated, but you want to make any changes, you have to manage it across all the languages.

We were thinking something like a checkbox for each instruction, automatically grouped into the headings, just check which instructions you would like included. That is most of the editing, "this one does not apply to me, etc etc".

These are in markdown basically, you can copy paste them to projects now if you 'edit source' to get the markdown for the Tasking Manager (already partially translated):

https://wiki.openstreetmap.org/wiki/OSM_Tasking_Manager/Default_roads,_buildings,_waterways_instructions

Original version here (pure markdown for copy/paste):
https://wiki.openstreetmap.org/wiki/User:Bgirardot/Tasking_Manager_Default_Roads,_Buildings,_Waterways_Instructions

xamanu:
I would like to add to this the idea of maintaining instructions on the OSM wiki, within data items, that are machine readable and can be pulled in by the Tasking Manager. This could potentially update the instructions on older projects, too.

russdeffner:
Great to see this conversation started again. I wonder if the wiki is not the ideal place for the instructions. We need them to be very basic and applicable globally, while allowing for pm's to add the local schema. I could see something like 'standard highway instructions' to be very enticing to edit a whole bunch on the wiki and if they're automatically ported/updated into projects that may not be a good idea. I think we want some kind of 'versioning control' on default instructions so we can delicately craft them to apply to all projects of that type and make sure changes don't negatively affect existing projects.

giblet22:
@russdeffner is there already a wiki page set up for this?

russdeffner:
[...]

PR #1859, opened 2019-09-05
Create autocomplete project search

### LINKED ISSUE #1850: Project Search Full-Text Partial Word Auto-complete
The new TM 4.0 design calls for an auto-complete project search field on the "Explore Projects" page. After research, this is commonly implemented using an Elastic Search instance which is not part of TM. A few people have tried to implement it using only PostgreSQL but there are manual steps to increase its usefulness.

image

The PostgreSQL GIN index, [which is already set up for the project search](https://github.com/hotosm/tasking-manager/blob/d07e1b6ba1b5c7ee5e2e9a5b1beb8ad1244e475b/scripts/database/migration-from-tm2-postgres.sql#L91-L99), likely needs a new API created to enable the autocomplete in the design spec as it is shown. But there are some choices.

There are two things the autocomplete API could return using PostgreSQL:
1. **query suggestions**: suggested words like "Malaria" or "Mozambique" as show in the design mockup. Query suggestions are made up of ngrams. Ngrams are are small sets of words together. 
2. **actual projects results**: actual project titles matching some of the words or partial word typed so far

There is a third option. It is mostly client-side and technically simpler. We could offer autocomplete for all titles of projects in the active search. We could add the project titles to the `project/search` API for all pages of the results, along with the `id` and map location. For 3000 projects, this should be easy to handle on the client side. We could still allow the user to ignore the autocomplete and use the full-text search too.

The fourth option is to leave the text search box as-is and/or to just make the existing `project/search` search support partial word prefix matching. It could have no new autocomplete popup since it already live-updates shown project cards on typing. It would be a bit slower.

Currently when the `project/search` is given a text query, the API only responds with projects that contain the matching **whole** words from the index. Instead, this new API should respond given incomplete words with ngrams or projects the GIN full text search index.

“prefix matching” using a syntax such as `to_tsquery('humanit:*')` as below for the `test & ar...` search. It can allow us to do an example search of "`ssa ar`" to find the results with "SSA" or "arbitrary" in my test database.
(note: the main `project/search` currently replaces spaces with `|` operator):
`SELECT * FROM project_info WHERE text_searchable @@ (to_tsquery('ssa | ar:*')) = true;`
image

For 1) my research indicates we would need to build an ngram table from the document text which we would return the query suggestions from, based on frequency of the ngrams. We might update it on inserts. I haven't seen any examples of re-using the `tsvector` data we have in `project_info.text_searchable`. Here is a in-depth guide with pseudocode: https://towardsdatascience.com/implementing-auto-complete-with-postgres-and-python-e03d34824079 

For 2) I think we can mostly use the code we have and just simplify the query from `project/search` to make it faster with a new endpoint and add the prefix matching feature above. I don't think it would be as useful as query suggestions. This elastic search partial reimplementation might help if the simple prefix matching doesn't work: https://stackoverflow.com/questions/56894979/edge-ngram-search-in-postgresql

I think the main thing is that we not over-engineer this. Based on the research, it could be a time sink to implement this. With only about 3,000 projects, there is not really much text to search. A basic solution probably solves it.

Other resources
* [very simple postgresql text search](https://stackoverflow.com/questions/48949638/postgresql-fastest-index-for-autocomplete)
* https://www.postgresql.org/docs/11/textsearch-controls.html
* https://www.postgresql.org/docs/9.6/pgtrgm.html
* https://github.com/wagtail/wagtail/issues/3709 and https://gist.github.com/BertrandBordage/4aee04f6931e4fa8c2c351eb7eeb8e31
* https://czep.net/17/full-text-search.html
* `select to_tsvector('simple', 'Yorick Peterse') @@ to_tsquery('simple', 'yor:*') --true`: the  "simple" turns off English specific features but you cannot mix dictionaries.
* https://postgrespro.com/list mailing list search query suggestions do use PostgreSQL`pg_trgm` according to [this document linked by CrunchyData/Paul Ramsey at FOSS4G-NA](https://www.postgresql.eu/events/pgconfeu2018/sessions/session/2116/slides/137/pgconf.eu-2018-fts.pdf) though no source obvious.

JorgeMartinezG:
This PR closes #1850 . The endpoint returns the `project_id` and `name` matched given a keyword

**How to test**:
XGET _/api/v2/projects/queries/auto-complete/?keyword=colombia_


thadk:
* I think w should add a LIMIT of 20. The query above returns 63 results on the keyword "colombia" but takes 26,060 milliseconds for the June 29 database. With a limit of 20 this drops a lot (as long as it can find 20 results)
* Multi word searches don't seem to return any results. I think we should use an `OR` search term between the words just like the search box. Alternatively, you could return the project ID of the matching project in a JSON object with the project name so I can set the search to e.g. `#1111` when the project title for `#1111` is clicked from the choices.
* Partial word searches don't seem to return results yet. See my comment in the issue about `:*`and we may have to figure out how translate this into sql-alchemy.

thadk:
Thanks @JorgeMartinezG – mostly looks good. I put my code front-end implementation in this branch too. I just noticed that a search for `ts` on the autocomplete returns these strange results:

image

Also some of the projects it is returning do not necessarily exist when I type the `#xxxx` in the search box.

willemarcel:
@thadk @JorgeMartinezG I don't know if something is missing on my environment, but it's not working well. These are the projects I have on my DB:

Screenshot_2019-10-08 HOT Tasking Manager

Some searches works, for example when I type "re"

Screenshot_2019-10-08 HOT Tasking Manager(1)

But if I type "review", it doesn't return anything (both in the autocomplete or in the project search)

Screenshot_2019-10-08 HOT Tasking Manager(3)

xamanu:
@JorgeMartinezG Can you please share the status on this issue? Thank you.

thadk:
@JorgeMartinezG can you merge the backend this week so I can make sure there isn't anything weird with the new version of the searchbox before I finish?

Wille might need to update his search index if it seems okay to you.

xamanu:
@JorgeMartinezG What is needed to get this in?

ramyaragupathy:
@JorgeMartinezG @willemarcel  - should we recreate this fix in a separate branch?

ramyaragupathy:
stale

PR #1862, opened 2019-09-10
Add views to select pages to map and validate

willemarcel:


frontend/src/components/taskSelection/mappingTypes.js
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+
+import { RoadIcon, HomeIcon, WavesIcon, TaskIcon } from '../svgIcons';
+
+export function MappingTypes({ types=[], colorClass }: Object) {
```
thadk: Good to see this as a separate component.

frontend/src/components/taskSelection/map.js
```diff
@@ -0,0 +1,166 @@
...
+            'BADIMAGERY', '#e04141',
+            'rgba(0,0,0,0)'
+          ],
+        }
+      }, 'tasks-icon');
+      map.addLayer({
+        id: 'selected-tasks-border',
+        type: 'line',
+        source: 'tasks',
+        paint: {
```
thadk: I think the task borders should be smaller at wider zoom levels so it doesn't turn all white. Sometimes it is easy to design/test those things on a throw-away map in Mapbox Studio and then download the final zip file of the MapboxGL styles from the main page to copy/paste into the Javascript. (edited)

thadk:
@willemarcel Haven't found the next PR for this work yet but this project's imagery seems to cause some problems yet: http://localhost:3000/projects/5950/map

PR #1865, opened 2019-09-11
Include mappingTypes in project summary

ramyaragupathy:
Updated ProjectSummary DTO and related function to return mappingTypes as a String Array.


willemarcel:
Thanks again, @ramyaragupathy !

PR #1681, opened 2019-06-07
API endpoint to list project contributions by day

JorgeMartinezG:
This Pull request closes #1615 .

- When creating a request to `/api/v1/stats/project/<int:project_id>/contributions/day`, the response returns the contributions of mapped and validated actions by day of a given project.

**Note.** To test this pull request, I recommend to use a database dump.

PR #1868, opened 2019-09-12
Add the Project Detail page to TM 4

### LINKED ISSUE #1630: Redesign the project page
Redesign the project page according to the design reference: https://projects.invisionapp.com/share/VCQTFR6YSJR#/screens/355947347

thadk:
This PR closes #1630 once complete. 

Screenshot_2019-10-15 HOT Tasking Manager(3)


server/services/project_service.py
```diff
@@ -77,16 +77,24 @@ def get_contribs_by_day(project_id: int) -> ProjectContribsDTO:
 
         contribs_dto = ProjectContribsDTO()
         dates = list(set(r[1] for r in stats))
-        dates.sort(reverse=True)
+        dates.sort(reverse=False)  # Why was this reversed?
```
thadk: @JorgeMartinezG Not sure why this was reversed, hopefully nothing unexpected broke when I un-reversed it for the cumulative counts.
https://github.com/hotosm/tasking-manager/blame/5c151ba8cd60aac72a194fb4e67b2746cdc49d34/server/services/project_service.py#L80

frontend/src/components/taskSelection/taskList.js
```diff
@@ -1,20 +1,21 @@
...
+import React from 'react';
 import { FormattedMessage, FormattedRelative } from 'react-intl';
 
 import messages from './messages';
 import { useFetch } from '../../hooks/UseFetch';
 import { compareTaskId, compareHistoryId } from '../../utils/sorting';
 
 function getLastTaskActivity(task, activities) {
-  return activities.filter(
+  return activities && activities.filter(
```
thadk: This line was crashing without the truth check

thadk:
Noticed some issues to keep in mind with the shared components for Select Task and Project Detail pages:
- The markdown parser is more strict than the old one so `###Header` doesn't work but `### Header` does. Same with e.g. `![Oil Palm Outlined] (https://i.postimg.cc/BbB53Cj8/outlined-Orchard.png)` – it doesn't work with this Marked parser.
- and the imagery component can overflow with long Mapbox URLs

For example, see: https://user-images.githubusercontent.com/283343/65552208-dabfd380-def1-11e9-9f21-eacb628bebe6.png

server/services/project_service.py
thadk: If you can suggest a way to get the total number of tasks after the splits from the backend, I think that would help the visualization a lot! Right now 300% is hard to understand.

willemarcel:
Nice work, @thadk 

I pushed some changes fixing details on the style.

Would you be able to make the chart look more similar to what was planned? https://drive.google.com/open?id=1cyjNIqNxg8IBbeHb9oNLd9NUM9FHpLSX

About the anchors link, I was thinking about using that code: https://github.com/dominikbulaj/react-scroll-into-view/blob/master/src/index.js (without installing to avoid injecting more dependencies)

Do you think we could start working on the `Question & Comments` section?

thadk:
@willemarcel Thanks, those look good!

For the chart, the tooltips on the area chart is much less useful. We may be able to draw both on top of each-other but I'm not sure. 
The color of the tickLines can be customized easily. We really need to fix the percent problem on the backend to fix that axis, then we can remove the axis label. Do you like the 
image vs 
image
 – I thought this way I did was an improvement, though it will need to be internationalized and load the colors from the theme with classes.

Good idea on `react-scroll-into-view` – I saw that you used it before.

I don't think I will be able to implement the comments section in this job as there are a lot of other areas to finalize and we haven't talked about the comments API yet. I'd be happy to give suggestions on the API though. It seems important to keep comments updated so people can handle them if they are important. I found this [example codebase](https://github.com/PierreRochard/realtime-flask-experiment) with [2 or 3 "Realtime Web" blog posts](https://rochard.org/) showing how to do Flask-SocketIO+PostgreSQL Listen/Notify instead of adding a new `message queue` service if you don't really want to. This could also be part of, or help with notifications site-wide.

willemarcel:
@thadk  I prefer the `* done * validated`. For me, it's ok if we have just a line chart (without the background colour below the lines). Just set the chart colours to be more similar to the mockups.

server/services/project_service.py
JorgeMartinezG: @thadk the reversed=True is for listing the most recent dates first. cc @willemarcel

willemarcel:
> The backend side is OK. Just a quick comment with the reversed sorting of dates. If @willemarcel agrees not to list most recent dates first in the frontend side, this PR may be merged

the first results on the api should be the most recent ones

server/services/project_service.py
thadk: @JorgeMartinezG Thanks! Do you see how in the screenshot visualization "Contributions Timeline" the percent's are above 100% – do you know a way to get the number of tasks after splitting on the backend?

thadk:
Thanks all – I can reverse it back in Python.

server/services/project_service.py
JorgeMartinezG: you mean the total_tasks column is not giving the correct value? One quick way is to filter all tasks given project id, like:
```
Task.query.with_entities(func.count(Task.project_id).label('total'))
.filter(Task.project_id==project_id)
.group_by(Task.project_id)
```

Have not tested locally, but it should work

thadk: Just saw this: I will try that @JorgeMartinezG

thadk: @JorgeMartinezG No, that is returning the same number as `project.total_tasks`. At least for https://xenodochial-ramanujan-f0fd23.netlify.com/projects/6175  it should be greater than `1150`.

thadk: Oh, I counted the number of tasks in the geojson and it was also 1150, even though the highest ID is 1168. I wonder if the backend code is getting the cumulative number of task editing sessions opened instead of cumulative number of tasks mapped or validated. Can someone take this backend back over, please?

thadk:
I think this is ready for review and merging. 

If we review the way this old PR is counting "contributions" on the `projects/{project_id}/contributions/queries/day/` endpoint so that it a task mapped twice across time, I think the percent in my visualization will no longer exceed 100%. https://github.com/hotosm/tasking-manager/pull/1681

willemarcel:
Thanks for the work, @thadk !

I pushed some improvements to the project detail page.

I also disabled the zoom to project boundaries on task selection because when the project boundaries is huge, zooming out would make the user lose context of what they selected.

The task layer is failing to load most of the time (it doesn't raise an error anymore, but the layer is visible only when I select a task on the list). On those cases, the unselect doesn't update the map (even when clicking only on the list).

PR #1871, opened 2019-09-17
Capture latest status of tasks

ramyaragupathy:
Per chat w/ @willemarcel - introducing `api/v2/projects/{project_id}/activities/latest/` to list down all tasks for a projects with:
* task ID
* current status
* last activity date and
* last mapper name who contributed to the task


willemarcel:
@ramyaragupathy I fixed the error that I discovered. Could you review when possible?

PR #1884, opened 2019-09-30
Store user profile url within database

### LINKED ISSUE #1879: Save user's profile picture URL when user logs in
When the user register or logs in on OSM, store on the user's database entry the URL of his profile picture on OSM, that way we can exhibit it on some new frontend views.

Also update the `/api/v2/users/{userid}/` and `/api/v2/users/queries/{username}/` endpoints to return that information.

JorgeMartinezG:
This PRs stores OSM user profile picture url within TM database. Also, a test was changed in order to store also this url.

image

This PR closes #1879


JorgeMartinezG:
@ramyaragupathy @willemarcel I have just updated the branch with the update profile feature. Right now you cannot see the profile picture within the frontend because the picture url was removed from response url. However, you can check it from API calls:

```
/api/v2/users/{userid}/
/api/v2/users/queries/{username}/
```

migrations/versions/996866bb604e_.py
```diff
@@ -0,0 +1,42 @@
+"""empty message
+
+Revision ID: 996866bb604e
+Revises: 7435b0a865e6
```
xamanu: I think we advanced already in `develop` to a new migration

JorgeMartinezG:
@xamanu Rebased branch with latest develop and created new migration on top of it

Issue #1579, opened 2019-05-15
ML based suggestions for tasks


xamanu:
Based on the Tasks Annotation backend (#1577, from #1349 and related to #1264) the ML derived information from *looking glass* shall be integrated in order to facilitate the best tasks to the right mapper and validator.

The new [design introduces lists of tasks for the project page](https://projects.invisionapp.com/share/VCQTFR6YSJR#/screens/358973356), these should be restricted to the ones the user actually can do actions on (based on permissions, or status) and sorted by the ones that are best suitable. In case the *looking glass* ML source has been activated for a project, use this data to present tasks with low estimated complexity to unexprienced mappers, and estimated tasks with high complexity to advanced mappers. Likewise for validators.

fitoria:
WIP here: https://github.com/hotosm/tasking-manager/tree/feature/task-difficulty-sort

PR #1885, opened 2019-10-02
Feature:task difficulty sort

fitoria:
Fixes for #1579

### DIFF
--- a/server/api/tasks/resources.py
+++ b/server/api/tasks/resources.py
@@ -5,14 +5,16 @@ from flask import send_file, Response
 from flask_restful import Resource, current_app, request
 from schematics.exceptions import DataError
 
+from server.services.mapping_service import MappingService, NotFound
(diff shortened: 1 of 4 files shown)


server/api/tasks/resources.py
```diff
@@ -381,10 +396,41 @@ def get(self, project_id):
...
-            return locked_tasks.to_primitive(), 200
+
+            if status:
+                try:
+                    status = TaskStatus[status.upper()].value
+                except KeyError:
+                    status = None
+
+            if order_by not in [
+                "building_area_diff",
```
xamanu: Can we use some more generic term here? I think for now these are buildings, but soon there will be roads or others. So what about using `effort_prediction_diff` or something like that?

xamanu:
Can you also please squash the commits? Or do you want me to do it while merging?

Issue #1601, opened 2019-05-24
Filterable tables - server api


willemarcel:
There are a lot of new filterable tables for tasks, projects and notifications. Probably it makes sense to think about a generic way on how to produce filterable (and searchable?) tables.

This issue is about the implementation on how tables are generated on the TM server side and the way they are exposed to the frontend client through the internal API.

Coordinate with @nrotstan around their work and how this fits together.

References:
https://projects.invisionapp.com/share/VCQTFR6YSJR#/screens/358164913
https://projects.invisionapp.com/share/VCQTFR6YSJR#/screens/357087237
https://projects.invisionapp.com/share/VCQTFR6YSJR#/screens/358973357
https://projects.invisionapp.com/share/VCQTFR6YSJR#/screens/357652194
https://projects.invisionapp.com/share/VCQTFR6YSJR#/screens/357651168

xamanu:
@willemarcel @fitoria what is the status? Can this be considered as done?

willemarcel:
It didn't go as we planned. We are using the common endpoints for each listing item view.

PR #1886, opened 2019-10-02
Adds sorting for projects and projects by owner

fitoria:
Fixes some views for #1601

### DIFF
--- a/frontend/src/App.js
+++ b/frontend/src/App.js
@@ -10,7 +10,7 @@ import { ProjectCreate } from './components/projectCreate/create';
 import { ORG_NAME } from './config';
 import { Home } from './views/home';
 import { AboutPage } from './views/about';
-import { ProjectsPage, ProjectsPageIndex, MoreFilters, ProjectDetailPage } from './views/project';
+import { MyProjectsPage, ProjectsPage, ProjectsPageIndex, MoreFilters, ProjectDetailPage } from './views/project';
 import { Authorized } from './views/authorized';
 import { Login } from './views/login';
 import { Welcome } from './views/welcome';
@@ -45,6 +45,7 @@ function App() {
             <Welcome path="welcome" />
             <Settings path="settings" />
             <EmailVerification path="verify-email" />
+            <MyProjectsPage path="projects/" />
             <SelectTask path="projects/:id/map" />
             <ProjectDetailPage path="projects/:id" />
             <Redirect from="project/:id" to="projects/:id" noThrow />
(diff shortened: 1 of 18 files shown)


xamanu:
Please reroll, @fitoria. Thanks.

xamanu:
@thadk can you please check, whether this has all you need for the "my projects" page?

thadk:
@xamanu @willemarcel Just to be clear, does the main view of My Project page only show projects that either You Favorited, You Contributed to, or You Created?
image

It seems like I have an implementation to get the set of projects favorited (/api/v2/projects/${projectId}/favorite/) and now this gives me a way to get the set of projects created (/api/v2/projects/?createdByMe=true) but I still need a way to get the set of projects that I contributed to (can I use the task API?). Then I think I need to "union" the three sets together on the front end, unless we can/want to do all three in this PR.

willemarcel:
@thadk I think the all options should show everything, but if it's too heavy to load the data from the 3 requests, we could remove the `ALL` option.

You can get the projects a user contributed to (the main function of that view) on `/api/v2/projects/queries/{username}/touched/` (it seems it's failing)

`/api/v2/projects/queries/myself/owner/` returns the projects the current user created.

thadk:
It seems like we might want to search/sort the projects "touched" on the backend using the `/project/` api, as that could be many (hundreds?) projects for some users. Also the front-end filter would not include the full-text search data.

thadk:
@fitoria For the "all" and "contributed" buttons shown above, we decided in the daily call that we need ?contributedToByMe=true as an OR query in addition to ?createdByMe=true (so we can request both at the same time and see a list of both types of projects.) We need to be a little careful with performance because this API is used for everyone using the site so please only use new JOINs if these two special parameters are requested, or re-use JOINs where it is possible. The `/api/v2/projects/queries/{username}/touched/` API might give you a hint as to how to do it efficiently.

Wille said that the `/api/v2/users/queries/favorites/` endpoint should already include the needed fields for project cards, so no action is needed yet on that. I will continue working on the frontend.

thadk:
I think this is ready for review to merge again. We can do the backend for the other tabs in another PR. This will be important for testers to navigate to their own maps.

thadk:
Creates the link for editing project you've created behind "Project Page", to this view https://github.com/hotosm/tasking-manager/pull/2034
Currently it shows this link to all projects, but if we sent information about whether you have management capabilities for that project on your account, it could change/hide the Edit Project link.

thadk:
Also this does not show Draft projects yet, maybe that should be added as another filter in the next followup for this PR.
`CreatedByMe` (done), `CreatedByMeArchived`, `CreatedByMeDraft` (?), `FavoritedByMe`, and `ContributedToByMe`

PR #1896, opened 2019-10-04
Update serialised names for Summary

ramyaragupathy:
Per chat w/ @willemarcel - updated serialized name for:
* enforceMapperLevel
* enforceValidatorRole

Added the following new fields:
* allowNonBeginners
* private


willemarcel:
@ramyaragupathy I added an additional field that was missing and we didn't notice

PR #1899, opened 2019-10-04
Include gender field in user profile

### LINKED ISSUE #1372: Add gender to user profile
Let's make sure we add a gender field as part of user profile redesign @xamanu @smit1678 

Gender is a key metric we're working to track (and improve!) across all programs and until now has been very difficult to do so for our remote mapping contributor community.

My preference would be to make it required with a 'prefer not to say' option so that users don't have to share the info if they choose not to.

JorgeMartinezG:
This pull request closes #1372 and changes were done within both client and server.

The following gender options were included:
- Male.
- Female.
- Prefer to self-describe
- Prefer not to say.

**How to test**

0. Execute database migrations.
1. start your client port 3000
2. Login with your osm credentials.
3. Go to your user settings.
4. Select the desired gender option.
5. Press save

**Note.** Every time some user select the third option, a text field will appear.
image


JorgeMartinezG:
@thadk @willemarcel In the frontend side, I had to change the partial completeness test, since two new fields are included within the PROFILE_RELEVANT_FIELDS, the calculation differs slightly. Let me know if there are suggestions in order to keep previous value

https://github.com/hotosm/tasking-manager/pull/1899/files#diff-f424ab7f85948cd18e09d7b434b10d42R125

willemarcel:
Thanks, @JorgeMartinezG ! I runned the migrations, but it's failing to load the user details endpoint:

```
127.0.0.1 - - [07/Oct/2019 09:36:27] "OPTIONS /api/v2/users/queries/wille/ HTTP/1.1" 200 -
[2019-10-07 09:36:27,208] CRITICAL in resources: User GET - unhandled error: None is not a valid UserGender
```

Another suggestion: rename `self_description` field to `self_description_gender`.

JorgeMartinezG:
@willemarcel I added the fix with rebased branch and modified the column name

xamanu:
@JorgeMartinezG Can we sort the answers by alphabetical order?

JorgeMartinezG:
@xamanu Done!
image

PR #1900, opened 2019-10-06
define actions to users based on project and task status

willemarcel:
Functions to verify if a user can map and/or validate a project and task and then show on the interface the action that the user can execute


xamanu:
Please squash the commits before merging.

PR #1905, opened 2019-10-08
Include allowedUsers list for private projects

ramyaragupathy:
Per chat w/ @willemarcel - added a new field to summary endpoint that lists allowed users for private project.

```
"allowedUsers": [
    "Abdul-Basit_eHA",
    "kazeem",
    "ugwuzor chinwe",
    "seyeabrahams",
    "Junaid Ibrahim",
    "tobiwonderful",
    "Hadeezah Musa",
    "kerry_o"
  ]
```
@willemarcel - what do you think will be useful here - user id or username? At present this returns username, I feel it should return IDs though

### DIFF
--- a/server/models/dtos/project_dto.py
+++ b/server/models/dtos/project_dto.py
@@ -388,6 +388,7 @@ class ProjectSummary(Model):
     )
     allow_non_beginners = BooleanType(serialized_name="allowNonBeginners")
     private = BooleanType(serialized_name="private")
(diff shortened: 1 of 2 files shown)


willemarcel:
Thanks, @ramyaragupathy ! The ideal would be to use uid, but it would be complex to make that change, as we would need to migrate the data and change that endpoint: `/api/v1/user/search/filter/<username`

Issue #1907, opened 2019-10-09
Make private project only accessible by allowed users and ADMIN


willemarcel:
By chat with @xamanu, we thought about making everything behind `/api/v2/projects/<id>/` only accessible to the `allowedUsernames` and to the `ADMIN` role users.

Return 401 status code error if request is not authenticated and 403 if the user has not permission to access the endpoint.

ramyaragupathy:
Fixed through #2492

PR #1911, opened 2019-10-09
Restrict access based on team and org permissions

ramyaragupathy:
Per #1907  - added a backend check for allowed user list on private projects.

A private project can be accessed by:
* admins
* project author
* list of allowed users


willemarcel:
The authorization on `/api/v2/projects/{project_id}/` is inconstant. For example, I tried to access that a private project without authentication and it returned 403 code, then I logged in with an allowed user and it worked. then I made some request without the authentication header or with other user token and it worked. I made a request using curl without authentication and it worked.

Some other points that we need to work on:

- [ ] `/api/v2/projects/{project_id}/actions/feature` should return `403` status code when user has not permission, but is returning 401
- [ ] I could comment on a private project that I was not allowed/admin/owner
- [ ] `/api/v2/projects/{project_id}/comments/`, `/api/v2/projects/{project_id}/activities/` and `/api/v2/projects/{project_id}/contributions/` should be limited to allowed users too (currently they are public)
- [ ] `/api/v2/projects/{project_id}/` when the request does not have `Authorization` header it should return 401code

ramyaragupathy:
All endpoints trouched by this PR:

Class|Verb
-----|-----
ProjectsRestAPI|POST, HEAD, PATCH, DELETE
ProjectsQueriesOwnerAPI|GET
ProjectsQueriesNoTasksAPI|GET
ProjectsActionsTransferAPI|POST
ProjectsActionsMessageContributorsAPI|POST 
ProjectsActionsFeatureAPI|POST
ProjectsActionsUnFeatureAPI|POST
ProjectsActionsSetInterestsAPI|POST
ProjectsCampaignsAPI|POST, DELETE
TasksActionsMapAllAPI|POST 
TasksActionsValidateAllAP|POST 
TasksActionsInvalidateAllAPI|POST 
TasksActionsResetBadImageryAllAPI|POST 
TasksActionsResetAllAPI|POST 
OrganisationsRestAPI|PATCH, DELETE
CampaignsAllAPI|POST
CampaignsRestAPI|PATCH, DELETE
InterestsAllAPI|POST, PATCH, DELETE

Issue #1521, opened 2019-04-25
Show user token on the user profile page


willemarcel:
In order to use the API, a user needs to know his/her authorization token, it would be helpful to add it on the user profile page (only for the current logged in user).

uwaiszaki:
@willemarcel Can we use local storage for session token?

willemarcel:
@uwaiszaki I think it's ok, although I'm currently not the best person to answer this question.

It seems we have a `SessionToken`, which is already saved on the local storage and an `Authorization Token`, that can be seen when inspecting the network requests on the browser.

xamanu:
Just connecting some things: This might be related to https://github.com/hotosm/tasking-manager/pull/1407 and https://github.com/hotosm/tasking-manager/pull/1037

uwaiszaki:
@xamanu #1407 is using auth token to verify, generate and return new token.

uwaiszaki:
@ethan-nelson can u please guide me.

xamanu:
@willemarcel is this related to #202?

willemarcel:
I couldn't understand what #202 is about...

I saw some related issues like #1197, but I didn't get why it was included on task annotation and which solution was implemented.

willemarcel:
Besides exposing the token to the user, in the new React frontend implementation I didn't succeed to use the current Authentication view. One of the problems I noticed is that the authentication API uses flask sessions, that way the frontend needs to be served by the backend.

In my opinion, the backend and the frontend should be totally independent. The redirection of the user to the OSM website should be done by the frontend, not by the backend view.

My proposal is to rewrite the authentication view with https://requests-oauthlib.readthedocs.io/, so we can have an authentication API totally independent of the frontend.

fitoria:
WIP: https://github.com/hotosm/tasking-manager/commits/feature/redesign-login

willemarcel:
I created PR with @fitoria work + improvements: https://github.com/hotosm/tasking-manager/pull/1769

willemarcel:
Fixed with #1913

PR #1913, opened 2019-10-09
show api key card to users with expertMode enabled

willemarcel:
... and add `FormattedMessages` to forms

Fixes https://github.com/hotosm/tasking-manager/issues/1521


willemarcel:
@ramyaragupathy @xamanu I fixed the problem @thadk mentioned. Could you run the server and test to check if you notice some problems? You just need to navigate to `/welcome` and `/settings`

PR #1918, opened 2019-10-11
Allow PMs to configure available iD editor presets

nrotstan:
> Note: this is intended for TM v3

* When editing the Metadata section of a project, offer PMs the option
of limiting available iD editor presets based on the mapping types the
PM has indicated for the project

* If "other" mapping type is selected, remaining iD presets that aren't
assigned to any other mapping type are offered

* Include typeahead filter to help narrow down the list of presets

* Add `idPresetsLimitedByDefault` configuration option that controls
whether iD preset limiting is enabled or disabled by default for new
projects

* Don't let projects be saved if iD preset limiting is enabled but no
presets have been selected

* When a user opens iD from TaskingManager, include any preset limits in
the URL

* Do not apply configured iD preset limits to users in expert mode

* Add new id_presets column to projects

* Add gulp task that fetches the iD preset categories straight from the
iD source repo on Github and then converts that into an Angular config
module. There is a plan for the iD team to break out the presets into
their own NPM module that could be imported into other projects, at
which time this gulp task will no longer be necessary


xamanu:
Thanks for the contribution, @nrotstan!

* I tested it today, but it doesn't seem to restrict the presets on iD. Also in the URL is calls I can't see anything about it (switched off expert mode for my user).
* Can we hide the functionality to select this on project creation for now and only make available to PMs with "Expert mode" switched on?
* Where can `idPresetsLimitedByDefault` be set through the UI? 
* I would rather keep the default as disabled for now and once more people are using it, it got into documentation and project managers are happy with it we can move to make it enabled.

> Do not apply configured iD preset limits to users in expert mode
This makes sense. But also causes a bit of unexpected behaviour. I think I would abstain for now with this and make the presets part of the project if selected, also for users with "expert mode" on. What do you and others think?

nrotstan:
Thanks for taking a look, Felix. I should have provided better information.

`idPresetsLimitedByDefault` is a front-end configuration option living in the `client/taskingmanager.config.json` file. The setting itself is not configurable through the UI, but it only represents a default that can be altered for each project during creation/editing (see below).

iD Presets are setup for a project in the Metadata section, where a new option is found to limit iD Editor Presets. It's default value (checked or unchecked) corresponds to the `idPresetsLimitedByDefault` configuration setting, but can be manually turned on or off by the PM for their project. If we don't wish to have the "Limit iD Editor Presets" checked by default, we can simply set the `idPresetsLimitedByDefault` configuration option to false.

limit_presets_option

if iD Editor Presets are enabled for a project, the PM can then choose a mapping type on the left, and this will cause corresponding iD presets to be offered on the right. The PM can choose the presets they wish iD to offer to mappers. If they don't wish to limit presets, the PM can simply uncheck the "Limit iD Editor Presets" option.

selecting_presets


When a (non-expert) mapper opens iD for a task in that project, it will apply the preset restrictions.

limited_presets_in_id

**If a mapper is in expert mode, no iD preset limits will be applied**. If you're not seeing the limited presets in iD after setting them up on a project, it's definitely worth double-checking your expert-mode setting. This was added to address the concern that advanced users would get frustrated with the limits, but could certainly be changed.

xamanu:
Hi Neil. Thanks for the further explanation! No worries, it was already quite understandable from the beginning.

I don't know why it wasn't restricting when I tested it earlier. Now it does work well.

> we can simply set the idPresetsLimitedByDefault configuration option to false.

Yes, please let's do this.

> If a mapper is in expert mode, no iD preset limits will be applied

With the expert mode, I think we should consider two potential changes:

* Show only the "iD Editor Presets" section for project creation, if the project manager has "expert mode" on. This allows us to deploy it already without updating documentation on several places.
* I would like to think openly whether it wouldn't make sense to apply the preset restrictions to all users, regardless of the "expert mode". I understand that it might be useful, but also hard to expect from the user side.

nrotstan:
I've committed these changes:

* Disable iD preset limits in projects by default

* Only offer iD preset limits during project creation/editing to PMs in
expert mode unless limits are already enabled in that project

* When the iD editor is opened, apply any configured iD presets to all
users, including users in expert mode

PR #1919, opened 2019-10-16
Adds ml-enabler integration enpoints to TM.

fitoria:
Adds automatic prediction generation on project creation, ml assisted
task split and task annotations creation based on predictions.

WIP for ml-enabler integration

adds automatic task annotation creation on project create

Skip ml-enabler test temporarly


server/config.py
```diff
@@ -28,6 +28,30 @@ class EnvironmentConfig:
     # A freely definable secret key for connecting the front end with the back end
     SECRET_KEY = os.getenv("TM_SECRET", None)
 
+    ML_ENABLER_URL = os.getenv("TM_ML_ENABLER_URL", "https://ml-enabler.hotosm.org/v1")
```
xamanu: Can we have a config file without a fallback to our resources? Ideally we specify them in the config file ourselves, but we don't want to expose them/

ramyaragupathy:
Stale

PR #1924, opened 2019-10-18
Custom Editor

### LINKED ISSUE #1916: Custom Editor Options
I'd like to discuss a new feature that will allow Project Managers to add a custom editor endpoint to a project that will be added as an additional option in the dropdown list:
image

The plan is to add it as a "Custom" option to the editor options in the "Settings" section of the `edit-project` page. 

cc @xamanu

zlavergne:
>Note: This PR requires a migration

With this PR, a PM with expert mode enabled can add a custom editor to a project.

# Demos
## Project edit
custom-editor_edit
## Mapping
custom-editor_mapping

**Note:** This will currently only work with iD-ish urls or JOSM urls that don't require parameters.

cc @xamanu @russdeffner 

closes #1916


zlavergne:
@fitoria figured it out. Long story short, since I was using an uninitialized `ng-model` (`projectCtrl.customEditor`) when rendering, the checkbox was set to `null`. When the rest of the data was input but the checkbox wasn't toggled, nothing triggered `projectCtrl.customEditor.enabled` to ever be created. 

I decided to fix this on the backend by:
1. changing the default for the column in the model to `False`
2. adding the same default to the DTO object

xamanu:
Rerolled after a conflict with newest `master` after PR #1918 and catched up to newest migration. Tested it again and it is good to go.

Issue #1881, opened 2019-09-30
New notifications endpoint


ramyaragupathy:
From https://github.com/hotosm/tasking-manager/issues/1644

thadk:
To support #1655 we would need to index the text in subject, body, and ID of the task/project so it can have full text search like we have in "Explore Projects" with the PostgreSQL GIN index.
https://www.postgresql.org/docs/11/textsearch-controls.html
There are a few migrations in the codebase for the `projects` table we could probably copy to create the needed columns for metadata storage.

ramyaragupathy:
Basic functionality implemented through #1928

PR #1928, opened 2019-10-21
Redesign - Notifications API

ramyaragupathy:
* [x] TM notification/message plus email sent to a project's contributors list(whoever has done a state change to a task) whenever there is a task comment
* [x] When there is a project chat - TM notification/message plus email sent to a user list who has favorited a project 
* [x] Recent activities from favorite and contributed projects
* [x] Introduce a notification model
* [x] Notification class and service to update and fetch unread count for a user id in the backend
* [x] Sort the 10 or 11 notification types into Messages, Tasks, Projects. Allow these to be queried from frontend in batches.

Move to another issue:
 Notifications searchable by task ID, project ID or contents by a single text field (<a href="https://github.com/hotosm/tasking-manager/issues/1881#issuecomment-565045081">Full Text Search</a>)

Frontend
* [x] Pagination
* [x] Basic cards
* [x] Basic page design, including routes for messages
* [x] Category Tabs with API integration
* [x] Allow message bodies to be viewed along lines of design
* [x] Allow user to see that they have new unread messages
* [x] More styling of cards to match design, especially distinguishing types of messages.
* [x] Add avatars component to notification card once they're merged into develop for messages from people.
* [x] Notifications popdown (rehash most of the above)
* [x] "2 hours ago" in whatever the final relative date library we are using is (react-intl 3 or 2?).
* [x] Allow marking deleting messages with confirmation (need frontend API).
* [x] Customize the project widgets to notifications query parameters (search box, sort box, etc)
* [x] Use Redux to store notifications for popdown and page.

image


thadk:
@willemarcel I just noticed that popdown from the bell. It seems like a good use of Redux to cache notifications since they're accessible from every page.

willemarcel:
@thadk yes, we can use redux for it. Also, move the bell to the right, in order to put it on the left side of the user avatar.

willemarcel:
@ramyaragupathy @thadk Some comments on the notifications. You can merge it if you want and fix those issues later.
- [ ] on the backend, use the `ListedUser` DTO instead of just the name of the user. We don't need all the fields of this DTO, but the id, username and picture_url. This will avoid extra API requests to get the picture_url of the users
- [ ] on the backend, when the user doesn't have notifications, return an empty array instead of a 404 error on the list endpoint.
- [ ] The search on the notifications could be by projects and users (the api already supports it) But I think we could let it for a next sprint and just hide the search field for now.
- [ ] Align the `clear filters` link with the search text input
- [ ] Align better the bell icon with the username text (if the bottom border is a problem for the alignment, we could drop it).
- [ ] When we click on the bell, open the tooltip instead of navigating to the notifications page
- [ ] The red ball indicating that I have non read notifications didn't show for me.
- [ ] On the `NotificationBodyModal`, do as much as possible to replace hard coded colours by those defined on our css palette.
- [ ] Remove the string indicating the message type (between the eye and trash icon on the notification card)
- [ ] on the NotificationsResults, I think we could try to use the `state.notifications` instead of (or together with) the `state.isLoading`on the ready field of the ReactPlaceholder in order to avoid it showing and hiding anytime a new request is made (maybe this will need the second topic solved)

thadk:
Ready for review!

frontend/src/store/actions/notifications.js
```diff
@@ -0,0 +1,5 @@
+export const types = {
+  FETCH_INIT: 'FETCH_INIT',
+  FETCH_SUCCESS: 'FETCH_SUCCESS',
+  FETCH_FAILURE: 'FETCH_FAILURE',
```
willemarcel: @thadk would it be ok if we replace `FETCH` by `NOTIFICATIONS` here?

thadk: This makes more sense in Redux! Thanks.

xamanu:
Squashed and cleaned up the commits. Please check whether all is there, and have a quick look at the conflict with `develop`. Thanks.

Issue #1715, opened 2019-06-24
Project ownership


xamanu:
In some rare cases it can be observed that the owner of a project is not the one that created this. 

For example:

* https://tasks.hotosm.org/project/6167
* https://tasks.hotosm.org/project/6176
* https://tasks.hotosm.org/project/6202
* https://tasks.hotosm.org/project/6290
* https://tasks.hotosm.org/project/6381
* https://tasks.hotosm.org/project/6480
* https://tasks.hotosm.org/project/6497
* https://tasks.hotosm.org/project/6498
* https://tasks.hotosm.org/project/6539
* https://tasks.hotosm.org/project/6680
* https://tasks.hotosm.org/project/7027
* https://tasks.hotosm.org/project/7030
* https://tasks.hotosm.org/project/7143
* https://tasks.hotosm.org/project/7147
* https://tasks.hotosm.org/project/7183

xamanu:
We investigated in the database logs that the wrongly used owner is the subject of the query before creating the project:

* Some task activity: `2019-06-19 04:09:48 UTC:172.31.71.2(56460):taskingmanager@tm:[3615]:LOG: statement: INSERT INTO task_history (project_id, task_id, action, action_text, action_date, user_id) VALUES (6128, 6, 'STATE_CHANGE', 'VALIDATED', '2019-06-19T04:09:48.649496'::timestamp, 1004928) RETURNING task_history.id` 
* Right afterwards the project creation `2019-06-19 04:09:50 UTC:172.31.71.2(55114):taskingmanager@tm:[11189]:LOG: statement: INSERT INTO projects (status, created, priority, default_locale, author_id, mapper_level, enforce_mapper_level, enforce_validator_role, allow_non_beginners, private, entities_to_map, changeset_comment, due_date, imagery, josm_preset, last_updated, license_id, geometry, centroid, task_creation_mode, mapping_types, organisation_tag, campaign_tag, mapping_editors, validation_editors, total_tasks, tasks_mapped, tasks_validated, tasks_bad_imagery) VALUES (2, '2019-06-19T04:09:50.689893'::timestamp, 2, 'en', 1004928, 1, false, false, false, false, NULL, NULL, NULL, NULL, NULL, '2019-06-19T04:09:50.598505'::timestamp, NULL, ST_SetSRID(...`

clairedelune:
I'm facing the same issue with this project that I had created and cannot edit anymore: https://tasks-assisted.hotosm.org/project/56

[...]

PR #1957, opened 2019-11-06
Confirm project author is PM

zlavergne:
This PR is a simple way to help mitigate the project ownership bug discussed in #1715

### DIFF
--- a/server/services/project_admin_service.py
+++ b/server/services/project_admin_service.py
@@ -38,6 +38,10 @@ class ProjectAdminService:
         :raises InvalidGeoJson
         :returns ID of new draft project
         """
+        # First things first, we need to validate that the author_id is a PM. issue #1715
+        if not UserService.is_user_a_project_manager(draft_project_dto.user_id):
+            raise(ProjectAdminServiceError(f'User {UserService.get_user_by_id(draft_project_dto.user_id).username} is not a project manager'))
+
         # If we're cloning we'll copy all the project details from the clone, otherwise create brand new project
         if draft_project_dto.cloneFromProjectId:
             draft_project = Project.clone(draft_project_dto.cloneFromProjectId, draft_project_dto.user_id)
@@ -222,7 +226,7 @@ class ProjectAdminService:
     def get_projects_for_admin(admin_id: int, preferred_locale: str):
         """ Get all projects for provided admin """
         return Project.get_projects_for_admin(admin_id, preferred_locale)
-        
+
     @staticmethod
     def transfer_project_to(project_id: int, transfering_user_id: int, username: str):
         """ Transfers project from old owner (transfering_user_id) to new owner (username) """
@@ -246,4 +250,3 @@ class ProjectAdminService:
             project.save()
         else:
             raise Exception("Normal users cannot transfer projects")
-        


russdeffner:
I'm not sure this solves it; as the person creating of course is a PM or would not have the option to create a project. It is more like a transfer is happening just after the creation step. If this also checks for PM role for any/all project transfers then maybe it will mitigate the problem.

That said, I'd rather not see us put band-aids on issues but resolve to find the source of the bug and fix the underlying issue. Who knows, maybe this bug is also causing/potentially causing other issues and putting a 'mitigation' so we don't see it anymore doesn't help get the root code up to par.

ramyaragupathy:
Thanks for the support @zlavergne - Steps to test:

* Generate a PM token
* Manually update the user id in https://github.com/hotosm/tasking-manager/blob/master/server/services/users/authentication_service.py#L40 to a non-PM ID
* send create project API call with PM token
* Now this should fail with an error message `User is not a PM`

To clarify @russdeffner's doubts:

> I'm not sure this solves it; as the person creating of course is a PM or would not have the option to create a project. It is more like a transfer is happening just after the creation step. If this also checks for PM role for any/all project transfers then maybe it will mitigate the problem.

Following are the internal steps to create a project:
* A project manager clicks on create-project from the user menu dropdown
* Project creation workflow kicks in on the frontend - PM provides a bunch of data and finally clicks on `CREATE PROJECT`
* Now this triggers an API call and all the data provided by PM plus the PM details are sent along to the server
* Backend code on the server, verifies if the user is a PM and then creates a project

In the case of project ownership issue - mix up happens in the last step. There is no deliberate transfer happening after the project creation, rather the user ID is mixed up right before the data insertion.

To detail the last step further: 
* PM verification is the first step before any other function is performed on project creation action from the user. So for the problematic projects in the first step right user ID is supplied to the code and the user is authorised to create a new project. 

* Right after this authorisation check, we've the user ID picked up by the code for the author information. It is here the problem occurs. A different user ID suddenly comes up for the project and then subsequently the project is added to our database with a piece of different user information.

Detailing it with lines from our codebase:
* This line https://github.com/hotosm/tasking-manager/blob/master/server/services/users/authentication_service.py#L30 returns the token and the OSM User ID for the logged-in user. 
* And this block https://github.com/hotosm/tasking-manager/blob/master/server/services/users/authentication_service.py#L35-L38 checks for PM/Admin access. 

If the decoding is going wrong or a wrong user ID is supplied, then project creation should fail in the first step. But in all the mixed up projects it has gone past this first verification, which means the actual project manager(who has initiated the project creation) ID is generated here. So the token generation and the subsequent decoding is done right.

It could be 
* https://github.com/hotosm/tasking-manager/blob/master/server/services/users/authentication_service.py#L40 
* https://github.com/hotosm/tasking-manager/blob/master/server/api/project_admin_api.py#L80

where the ID is mixed up. When the right user id is supplied for the role verification in the first step, a different user ID is assigned in one of these lines.

After this ID assignment, actual DB insertion happens. @zlavergne has provided an additional verification right before the insertion, to make sure the user ID supplied for this step has PM access. This will mitigate the problem to a great extent. When this issue occurs on the backend, PM would get an error message, but re-triggering should work.

While so far we've been able to narrow down the problematic block in our code, we're still working to understand why this mix up occurs. Here's all the steps we've tried so far:
* From DB logs, check if there were deliberate actions from a user to transfer project. No such activity noticed. But from our logs for certain flagged projects, we saw some activity performed by related user IDs right before the project creation and this led us to our current research direction.
* We then tried reproducing the issue using four user logins(3 mappers and 1 PM) on local server setup and initiate random actions plus project creation from each of the users simultaneously. The issue was not reproducible.  
* To reduce the time difference for user actions further, we set up a script to send parallel requests to the server from multiple users with different levels of access performing task lock, unlock, and project creation. Still, this issue was not reproducible.

So far this issue has been really hard to reproduce on a local setup. However, we know where to add in a check to make sure non-PM IDs are not picked for author field during project creation. We've about  [15 projects on main instance](https://github.com/hotosm/tasking-manager/issues/1715#issue-459816495) reported for this issue. And in all these cases the mixup so far has been with non-PM user IDs. What if the mixed up user ID happens to be another PM(which has not happened so far)? then the project creation gets through but with the mixed up ID of another PM.  From the projects reported, this mixup has been only with non-PM IDs, but that doesn't eliminate the possibility of PM/Admin IDs getting mixed up at this stage.

So yes - this is not a permanent solution, but this will definitely reduce the number of such occurrences, while we investigate the root cause and ideate on including more data points to track user login.

PR #1960, opened 2019-11-06
fixes for saving project

zlavergne:
Quick fixes for problems when saving the project after adding, removing, adding custom editors


zlavergne:
@fitoria can you give more detail to repro? I was able to create a square task project without error.

PR #1961, opened 2019-11-06
Fix migration tables, campaigns table queries and tests

JorgeMartinezG:
This PR needs revisions since there are some variables that are removed which i am not sure if they are necessary.

1. fix migration file 0eee8c1abd3a_.py. Maybe Typo error.
2. Fix campaigns model table names and relationships. The table name **must** be changed within this file instead of the migration one. The reason is that whenever a new postgis model change is done, alembic will create campaign table, causing inconsistencies.
3. Fix campaigns table queries.
4. Set organisation name column without unique constraint. This creates new revision file f86698c827cc_.py. Same reason as 2


ramyaragupathy:
Per chat w/ @xamanu - removing lines that drops`organisation_tag` from projects table. This will be handled in a separate PR

**What does the migrations in this PR do?**

1. Creates new table for `campaigns` and `organisations`
2. Drops `organisation_tag` column from `projects` table
3. Populates all campaign and organisation related information from `tags` table into their respective tables
4. Removes `campaign_tag` field from `projects` table. Noen of the projects in our production DB has any campaign info attachedS

Steps 1-3 is handled by `e3282e2db2d7` and 4th step is from `f86698c827cc`

**What should be done differently to retain data?**

1. Creates new table for campaigns and organisations
2. Populates all campaign and organisation related information from `tags` table
3. Map the existing `organisation_tag` from `projects` table  to `organisation_id` from the new `organisations` table
4. Delete the `organisation_tag` from `projects` table

PR #1963, opened 2019-11-07
Fixes #1953 allow to clear field

### LINKED ISSUE #1953: api/v2/users/actions/set-user/ doesn't allow to clear a field
If I send a PATCH request with `{"name": ""}` or {"name": null} as the payload, the API returns a 200 status code, but doesn't update the name field. The same happens with other fields.

It would be good to have a better error reporting system. For example, if I send the payload with a unsupported content-type, it returns 200, but didn't update the user.

fitoria:


willemarcel:
It worked well with the other fields when we pass the value `""`

fitoria:
This is a bit tricky because if the DTO does not gets a value on a nullable field it will set to null by default, so right now I check for those nulls and it won't be possible to set to null but it will work with empty string and false.

Issue #1965, opened 2019-11-07
Migrate organisations and campaigns


xamanu:
We removed the migrations of content for organisations and campaigns from #1858 in #1961. Let's bring this back in a new migration, which includes a logic for consolidating several organisations with different names, such as HOT, HOTOSM, and Humanitarian OpenStreetMap Team.

This code snipped might be a good starting point:

```

    conn = op.get_bind()

    # Content migration: Migrate the campaigns tag in campaigns table
    campaigns = conn.execute(
        "select campaigns from tags where campaigns is not null"
    ).fetchall()

    # This will be used to consolidate the data in tags table
    dictionaries = {"HOTOSM": {"HOTOSM", "HOT-OSM"}, "ABC": {"abc", "ABc"}}

    for campaign in campaigns:
        result = campaign[0]
        for campaign_key, campaign_values in dictionaries.items():
            if campaign[0] in campaign_values:
                result = campaign_key

        query = "insert into campaigns(name) values('" + result + "')"
        op.execute(query)

    # Migrate the organisations tag in organisations table
    organisations = conn.execute(
        "select organisations from tags where organisations is not null"
    ).fetchall()

    # This will be used to consolidate the data in the tags table
    org_dictionaries = {"HOTOSM": {"HOTOSM", "HOT-OSM"}, "ABC": {"abc", "ABc"}}

    for org in organisations:
        result = org[0]
        if result.startswith("'") or result.startswith('"'):
            print(result)
            result = result[2:]
        for org_key, org_values in org_dictionaries.items():
            if result in org_values:
                result = org_key

        query = "insert into organisations(name) values ('" + result + "')"
        op.execute(query)
```

willemarcel:
@xamanu Is there any block to push a solution to this issue?

xamanu:
No blocker, but also not highest priority. This can be addressed while TM4 is in the testing phase.

xamanu:
Quick update here:

* @ramyaragupathy has implemented a first logic for the migration https://github.com/hotosm/tasking-manager/pull/2260
* Currently, we are actively going through all the organizations and finding duplicates, grouping some of them, reaching out to contact persons, etc. This will take another week approx. Then we can continue here to bring this data in.

xamanu:
@ramyaragupathy, please use the current state of editing for testing and completing the migration of organizations.

ramyaragupathy:
[...]

PR #1966, opened 2019-11-07
Organisation endpoint related

ramyaragupathy:
Related to @thadk's message from Slack -
>noticed the /api/v2/organisations/ is giving HTML instead of JSON on staging for the last day or three, breaking the contribute page


thadk:
Hi @ramyaragupathy @willemarcel – this looks okay, though my local database still has broken migrations.

Also, maybe I will consolidate the #1981 commit into this, I didn't see this until now.

thadk:
This works well but it still is returning `401 Unauthorized` on `http://127.0.0.1:5000/api/v2/organisations/` when the user is logged out and empty `tags:[]` when they are logged in. 
Are either of these intentional behavior or just pending tweaks?

willemarcel:
> This works well but it still is returning `401 Unauthorized` on `http://127.0.0.1:5000/api/v2/organisations/` when the user is logged out and empty `tags:[]` when they are logged in.
> Are either of these intentional behavior or just pending tweaks?

@thadk I asked Felix and Ramya and they confirmed the organisations endpoints should require authentication. The organisations is empty because the data migration is not ready https://github.com/hotosm/tasking-manager/issues/1965

Issue #1853, opened 2019-09-02
Project stats exceeding the max percentage


ramyaragupathy:
From project [#6317](https://tasks.hotosm.org/project/6317/dashboard), [#6387](https://tasks.hotosm.org/project/6387/dashboard), [#6305](https://tasks.hotosm.org/project/6305/dashboard)  [#6313](https://tasks.hotosm.org/project/6313/dashboard). 
![image](https://user-images.githubusercontent.com/12103383/64120015-c8ea8500-cdb8-11e9-8708-666fae52d399.png)

![image](https://user-images.githubusercontent.com/12103383/64120082-f0415200-cdb8-11e9-8f73-8d2241bc644a.png)

PR #1986, opened 2019-11-13
Fixes for #1853 bad counting on master branch

fitoria:


xamanu:
Just one question: we need to run a command (`python3 manage.py refresh_project_stats`) on the main deployment here. Can we do this, @dakotabenjamin? Or shall we put the refreshing into a migration?

dakotabenjamin:
Can you clarify what you mean by migration? Do you mean when `python manage.py db upgrade` is run?

So you could add it on a new line here: https://github.com/hotosm/tasking-manager/blob/master/devops/cloudformation/tasking-manager.template.js#L229 , but note that the command will run _each time a new instance is added to the application stack_. When `db upgrade` is run, for example, it will check if a migration is needed and exit gracefully if not.

xamanu:
> Do you mean when python manage.py db upgrade is run?

Yes. We run `python manage.py db upgrade` on every deployment (as you indicated), right? Within this PR we need to refresh some statistics in the database. This can be triggered either with a time (manual) call to `python3 manage.py refresh_project_stats`, as proposed by @fitoria. Or we could trigger this through a alembic migration so that it is getting called with our regular `manage.py db upgrade` call. What is preferred for the roll-out?

dakotabenjamin:
If you want it to only run once, then I suggest not putting it into a migration. `db update` runs not on every deployment, but on every instance initialization during autoscaling.

server/services/stats_service.py
```diff
@@ -19,32 +24,50 @@
...
+        user_id: int,
+        last_state: TaskStatus,
+        new_state: TaskStatus,
+        action="change",
+    ):
         """ Update stats when a task has had a state change """
 
-        if new_state in [TaskStatus.READY, TaskStatus.LOCKED_FOR_VALIDATION, TaskStatus.LOCKED_FOR_MAPPING]:
+        if new_state in [
+            TaskStatus.READY,
```
ramyaragupathy: May be we should also include `AUTO_UNLOCK_FOR_MAPPING` and `AUTO_UNLOCK_FOR_VALIDATION` here, since there is no state change

fitoria:
Added a new migration file, please note that this will take a long time depending on the db size.

server/services/stats_service.py
ramyaragupathy: Just realised after AUTO UNLOCK  task status is set to READY - so we need not include these here

Issue #1984, opened 2019-11-12
Update organisations list API


willemarcel:
Update `/api/v2/organisations/` endpoint https://github.com/hotosm/tasking-manager/blob/49861fe3ef64afa0c8b23d49dc99344c8ac25139/server/api/organisations/resources.py#L258

- [x] `managers` field is returning a null value, should return the usernames of the administrators/managers of the org. Actually the ideal would be having the pictureUrl too, like: `[{username: 'Admin_1', pictureUrl: null }, {username: 'B', pictureUrl: null }]`
- [x] restrict organisation creation to users with `admin` role

willemarcel:
- [ ] allow admins to filter by managers `/api/v2/organisations/?managers=<uid>`
- [ ] allow logged in users to filter only with their own uid, with the same url param `/api/v2/organisations/?managers=<uid>`

xamanu:
Assigning @fitoria, as he has been continuing on this issue. What is the current status here?

fitoria:
Work in #2029

PR #1987, opened 2019-11-14
[WIP] Improving organisations

xamanu:
This addresses #1984 

So far:

* Renamed in the context of organizations misleading `admin` to `manager`.
* Removed visibility option on organizations
* Made /organisations endpoint to give the full list to all users. The own organisation to logged-in users requesting their own user id. And organisations of any particular user, if the requested by a logged-in admin.
* Removed logic and endpoints for unused functionality


server/api/projects/campaigns.py
```diff
@@ -125,57 +125,7 @@ def delete(self, project_id, campaign_id):
...
-
-
-class ProjectsCampaignsActionsRemoveAPI(Resource):
-    @token_auth.login_required
-    def post(self, project_id):
-        """
-        Unassign campaign(s) for a project
-        ---
-        tags:
-          - campaign
```
willemarcel: @xamanu what do you think about renaming the tag to project campaigns or just campaigns to have it together with the other endpoints on swagger?

PR #1989, opened 2019-11-15
Add frontends views to manage teams, organizations and campaigns

willemarcel:


willemarcel:
@JorgeMartinezG Thanks for the review. I fixed all issues, except: 

> When selecting and org, I could not find a way to return to their manager. Same with teams and campaigns. Do you think it would be good to create a button to go back?. The position would be next to delete. Maybe this could be addressed in another issue.

I'll discuss with @hexiaok later. I thought about adding the submenu to all management views

PR #1991, opened 2019-11-15
fix set-interest for user URL

willemarcel:
The interests of a user can only be changed by themselves, so we don't need the `user_id` param on the URL. I updated it from `/api/v2/users/<int:user_id>/actions/set-interests/` to `/api/v2/users/actions/set-interests/`.

### DIFF
--- a/frontend/src/store/actions/auth.js
+++ b/frontend/src/store/actions/auth.js
@@ -26,7 +26,7 @@ export const updateUserEmail = (userDetails, token, relevant_fields) => dispatch
     }, {});
   const payload = JSON.stringify(filtered);
 
-  pushToLocalJSONAPI(`users/actions/set-user/`, payload, token, 'PATCH').then(() => {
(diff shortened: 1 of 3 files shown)


xamanu:
Shouldn't we use `/api/v2/users/me/actions/set-interests`

willemarcel:
Currently we have:

/api/v2/users/actions/set-user/
/api/v2/users/myself/actions/verify-email/

Let's decide on a pattern. I prefer `me` than `myself` (https://www.openstreetmap.org/user/myself/ we even has an OSM user named myself!)

xamanu:
Haha, not a very active user though. Yes, let's go with `me` and big :heavy_plus_sign: for doing it in one consistent way.

willemarcel:
@xamanu I changed the URLs to the pattern with `/me/actions/`

ramyaragupathy:
`me` seems consistent with many of the Rest examples I've come across @willemarcel @xamanu

PR #1995, opened 2019-11-20
Update develop on master changes

xamanu:


willemarcel:
Replace `op.drop_table("organisation_admins")` by `op.drop_table("organisation_managers")` on `/migrations/versions/e3282e2db2d7_.py`

willemarcel:
After fixing the error I mentioned on the last comment, I had to downgrade until `4b9fb4d3f349` to make all the migrations be applied. If master is on `4b9fb4d3f349`, I believe we will not have issues. We could try to import a new dump on staging and run the migrations to check if it works.

xamanu:
Yes, this PR was made with a dump from live. And as it rebases on the changes from `master` the migration `51ccc46f1b8a` was insertet right after `4b9fb4d3f349`. Live is on `51ccc46f1b8a`.

xamanu:
Fixed the dropping of table in the migration. Thanks for spotting that, @willemarcel. This can be reviewed again.

PR #2032, opened 2019-12-05
Embed iD editor within Tasking Manager

willemarcel:
Some issues we need to solve:

- [ ] iD css is overriding our app style.
Screenshot_2019-12-05 Tasking Manager


willemarcel:
https://www.npmjs.com/package/react-frame - this lib could be a solution to the above issue

willemarcel:
@xamanu @JorgeMartinezG I rebased it with develop, removed the `/editor` url and added iD in the /map and /validate projects views with the sidebar. I couldn't solve the problem of the CSS overriding.

To test it, select a task to map/validate and select the iD Editor, you''ll be redirected to `/projects/<id>/map/?editor=iD Editor`

Issue #1941, opened 2019-10-30
Add legend to tasks map


willemarcel:
The task maps (both on task selection and on the project detail pages) needs a legend.

More information on https://github.com/hotosm/tasking-manager/issues/1257

xamanu:
To me it seems this has been completed. Is this correct, @willemarcel ?

willemarcel:
@xamanu yes, closing

Issue #1257, opened 2018-10-26
Revisit Task Colors


russdeffner:
This issue is to combine #1256 and #246 with a general recoloring of the task status. In the Activation Working Group we have noted that the 'red' for invalidated is maybe causing mappers to stay away from those tasks. We may also want to take into consideration color impaired vision to ensure all our mappers can easily understand what tasks are available to them.

smit1678:
Thanks @russdeffner for bringing that up. One quick question: does an invalidated task even need to be colored? 

I can see the case for a project manager or validator to maybe know, but for a mapper it doesn't make a difference right?

russdeffner:
I think there is value in having some noticeable difference in a project that is invalidated - i.e. a bit of warning that it might be fixing stuff rather than just drawing new features.

xamanu:
Probably related to this: #1523.

zlavergne:
I've started using icons instead of additional colors. Here's an example of my implementation of task assignment: 
![Screen Shot 2019-05-13 at 1 03 14 PM](https://user-images.githubusercontent.com/9849118/57713987-c3620700-7630-11e9-9e0a-210a5a733e45.png)

For times when colors don't appropriately portray the status, I think icons could be super useful.

ethan-nelson:
I am a big fan of icons :+1: we've struggled for a while with all the different colors. I tried to add pin-striping in TM2 to indicate a checked out task, but ran into issues with the map library at that point. An icon might be a way to indicate the status in a manner different than colors without being too overbearing. Furthermore, being able to see at a glance what state of task someone has checked out (e.g. mapping versus validating) could be beneficial.

zlavergne:
I have the implementation for OpenLayers if you want to implement icons soon. But as @xamanu mentions in #1541 we may be switching from OL soon. If that's the case, it may be better to make the switch for the map library before adding icons. From what I've seen, not many people are making changes that rely heavily on the map so switching the map library might be safer than changing the whole front end framework. 

Thoughts @xamanu?

xamanu:
On the colors, we got some input from our redesign. Instead of adding more colors we were encouraged to reduce the states to what matters to the users:

![image](https://user-images.githubusercontent.com/97706/58168807-ae175900-7c7e-11e9-98d2-0e8ac8ab3a35.png)

What do you think about these states?

I like the idea of using icons for further indication, like user to tasks assignments in #1575. And i guess we will have more good use for this for future features.

[...]

PR #2035, opened 2019-12-05
Tasks select and map/validate views

willemarcel:
- Solves #1257 , https://github.com/hotosm/tasking-manager/issues/1941
- Rename `projects/:id/map` URL to `projects/:id/tasks`


xamanu:
Please run `black` to fix the failing ci

PR #2040, opened 2019-12-10
adds filters to team endpoints

### LINKED ISSUE #2023: Add filters to team list endpoints
We need to filter teams using some conditions:

- teams that a user is member
- teams that a user is manager
- teams that a user requested to join, but approval is pending
- by team name
- by team function

The two first filters are the most urgent.

fitoria:
Fixes #2023

### DIFF
--- a/server/api/teams/resources.py
+++ b/server/api/teams/resources.py
@@ -283,6 +283,32 @@ class TeamsAllAPI(Resource):
           - teams
         produces:
           - application/json
+        parameters:
+            - in: query
+              name: team_name
+              description: name of the team to filter by
+              type: str
+              default: null
+            - in: query
+              name: member
+              description: user id to filter teams that the users belongs to, user must be active.
+              type: str
+              default: null
+            - in: query
+              name: manager
+              description: user id to filter teams that the users has MANAGER role
+              type: str
+              default: null
+            - in: query
+              name: member_request
+              description: user id to filter teams that the user has send invite request to
+              type: str
+              default: null
+            - in: query
+              name: team_role
+              description: team role for project
+              type: str
+              default: null
         responses:
             201:
                 description: Team list returned successfully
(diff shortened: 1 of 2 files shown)


server/services/team_service.py
```diff
@@ -176,10 +176,47 @@ def delete_team_project(team_id, project_id):
...
-        query = Team.query.distinct()
+    def get_all_teams(
+        team_name_filter: str = None,
+        team_role_filter: str = None,
+        member_filter: int = None,
+        member_request_filter: int = None,
+        manager_filter: int = None,
+    ) -> TeamsListDTO:
+
+        query = db.session.query(Team).join(TeamMembers).join(ProjectTeams)
```
willemarcel: maybe the problem is on that line. I noticed that only the team that was assigned to a project was on the results

server/api/teams/resources.py
```diff
@@ -293,8 +319,37 @@ def get(self):
...
+        filters["team_name_filter"] = request.args.get("team_name")
+        try:
+            member_filter = request.args.get("member")
+            filters["member_filter"] = int(member_filter) if member_filter else None
+        except ValueError:
+            pass
+
+        try:
+            manager_filter = request.args.get("manager")
+            filters["manager_filter"] = int(manager_filter) if manager_filter else None
```
willemarcel: Would it cause an error if we don't convert to int? When I pass a string instead of returning 0 results, it didn't filter

server/api/teams/resources.py
```diff
@@ -293,8 +319,37 @@ def get(self):
             500:
                 description: Internal Server Error
         """
+
+        filters = {}
+        filters["team_name_filter"] = request.args.get("team_name")
```
willemarcel: Would it be possible to filter by the presence of the string instead of an exact match?

server/services/team_service.py
```diff
@@ -176,10 +176,47 @@ def delete_team_project(team_id, project_id):
...
+        team_role_filter: str = None,
+        member_filter: int = None,
+        member_request_filter: int = None,
+        manager_filter: int = None,
+    ) -> TeamsListDTO:
+
+        query = db.session.query(Team).outerjoin(TeamMembers).outerjoin(ProjectTeams)
+
+        if team_name_filter:
+            query = query.filter(Team.name.contains(team_name_filter))
```
willemarcel: @fitoria Do you think it'd be safe to use it to make the name query case insensitive?
`query = query.filter(Team.name.ilike("%{}%".format(team_name_filter)))`

PR #2049, opened 2019-12-10
Enhancement in project creation view

### LINKED ISSUE #2018: Project creation enhancements
- [x] Replace alert errors by a message above the back / next buttons
- [x] If you split the tasks, advance to the next step and then go back to split step, the split work is lost
- [x] Make possible to split the tasks more than one time using the click/draw functions
- [x] Update `tasksNo` after each split operation
- [x] Replace click/draw to split buttons by switch toggles (it makes more sense, as we are activating mouse functions, not properly executing actions)
- [x] Fix initial task size (in some cases, the initial task is too much larger than the project AoI)

JorgeMartinezG:
This PR closes #2018 
**Features:**
- Error notification as messages instead of alerts
Screen Shot 2019-12-10 at 1 10 27 PM
- Allow to return without resetting taskgrid.
go_back
- Fix for Tasks counter
- Updated intiial zoom level to set taskgrid.
- Split multiple tasks 
multiple_split

### DIFF
--- a/frontend/src/components/projectCreate/create.js
+++ b/frontend/src/components/projectCreate/create.js
@@ -103,7 +103,7 @@ const ProjectCreate = props => {
 
   return (
     <div className="cf bg-tan pb3 pl4">
-      <div className="fl vh-75-l pt3 w-30">
+      <div className="fl pt3 w-30">
         <h2 className="f2 fw6 mt2 mb3 ttu barlow-condensed blue-dark">
           {cloneProjectName === null ? (
             <FormattedMessage {...messages.createProject} />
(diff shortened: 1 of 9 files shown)


willemarcel:
Thanks @JorgeMartinezG ! I rebased and made some improvements on error messages and clone projects. Some details we could fix:

- [ ] When uploading a invalid file, like a JPG, the app breaks. We need to validate if it's a valid source before trying to add it to the map.
- [ ] If you draw an area bigger than allowed, try to go to the next step, then delete the area, the error message doesn't disappear. Maybe we need to move the validation to the create.js and disable the next button if the aoi is not valid.
- [ ] The split function is yet active on the trim task step and causes some errors,

willemarcel:
@JorgeMartinezG I'm merging it as @xamanu requested to have the updates deployed. I added the remaining issues to #2018 . You can open a new PR when you have the solution.

JorgeMartinezG:
Thank you @willemarcel . Will do!

PR #2061, opened 2019-12-12
Replace screenshot with TM4

xamanu:
Started working on flipping the documentation of TM slightly in preparation of the new release.


xamanu:
Currently waiting for the testing instance to come up to change the screenshot of the tasks view.

PR #2069, opened 2019-12-16
Upgrade localization in frontend to React-Intl v4

thadk:
This will allow us to use the changes specified in https://github.com/formatjs/react-intl/blob/master/docs/Upgrade-Guide.md like `useIntl` instead of injectIntl HOC.

**As of Dec 16 tests are blocked by known temporary babel error with all NodeJS versions**
```
/home/circleci/app/frontend/node_modules/react-intl/dist/index.js: The macro imported from "@formatjs/macro" must be wrapped in "createMacro" which you can get from "babel-plugin-macros". Please refer to the documentation to see how to do this properly: https://github.com/kentcdodds/babel-plugin-macros/blob/master/other/docs/author.md#writing-a-macro
```

This issue is preventing us from migration to react-intl v3 I think – **all of our localized tests fail** even with a new react-scripts which has the new babel-plugin-macros `https://github.com/formatjs/react-intl/issues/1511`
Blocked by https://github.com/kentcdodds/babel-plugin-macros/issues/131

The code itself works though. I moved it cleanly onto develop: https://github.com/hotosm/tasking-manager/commit/f0222a47f99d4851b214e5e5d99fad6d57922da3

There is a new util library react-intl team made to make it very straightforward to migrate, which was not available before.


thadk:
Still broken with `react-scripts:"3.3.0-next.80"` and `3.3.0`. I put a comment on the issue.

PR #2080, opened 2019-12-19
open id editor when user locks a task for mapping/validation

willemarcel:
It's missing to add other editors, but iD is ready. @xamanu feel free to merge it now or wait for the other editors. I think I can implement the other editors this Thursday.


xamanu:
Tested it and works. Wondering why testing fails though.

PR #2081, opened 2019-12-19
Improve reliability of task splitting

nrotstan:
* Split non-square gridded tasks the same way as cells created with
arbitrary geometry. One side-effect is that any x/y/zoom values are
discarded during split

* Add a sanity check to task splitting to ensure new tasks intersect the
original task, raising an exception if not

* Ensure new tasks from a split are fully cleaned up if deletion of the
original task fails for some reason


fitoria:
hi @nrotstan what's a good way to test this properly?

nrotstan:
Hi @fitoria the PR does three main things. First, it simplifies task splitting for non-square tasks (which can sometimes be problematic) by reusing the existing splitting code used for arbitrary geometries. There is a side effect to this, which is that the x/y/zoom values are discarded for those tasks, but I don't believe those are really used other than during task splitting, and they'll no longer be used to split non-square tasks. So if there are other places where the x/y/zoom could be used, that would be something to look at (but tasks created from arbitrary geometries don't ever get x/y/zoom values to begin with).

Second, a sanity check is added to ensure that the newly-split tasks actually intersect the original parent task. At one point in time there were some scenarios where the x/y/zoom values could be weird and lead to bizarre splitting. I belive it was because OpenLayers on the front-end could sometimes produce fractional zoom levels if someone used the scroll wheel on a mouse to zoom in, but the server always assumed zooms are whole-number integers. I don't know if that's still an issue, but this sanity check will at least cause an exception to be generated should the split result in a geographically nonsensical new task, and it shouldn't hurt if all is well.

And finally, if there is a failure during task split -- for whatever reason -- then any newly created tasks are explicitly deleted so that the system is left in a clean and consistent state following the failure.

I think the most important thing to test would be the splitting of non-square tasks to make sure you're happy with how it's working, as that would impact current behavior, and that the discarded x/y/zoom values aren't resulting any any adverse side-effects. The other two changes are really intended to help eliminate undesirable child tasks in edge cases when things don't go as planned.

zlavergne:
@xamanu is there anything else blocking this PR from merging?

xamanu:
Nothing. Merged. We need to take care to bring this into TM4, as we have diverging branches.

xamanu:
Thank you everyone! @nrotstan for providing the PR, @ramyaragupathy and @JorgeMartinezG for your review.

I left the related issue open (https://github.com/hotosm/tasking-manager/issues/1282), so we port it to TM4 - isn't going to be to hard, maybe we can even cherry-pick and merge.

PR #2082, opened 2019-12-22
add functions to open JOSM, Potlatch2 and FieldPapers editors

willemarcel:


willemarcel:
@xamanu I added the functions to open JOSM, Potlatch and FieldPapers. The JOSM part has some improvements in relation to TM3, like loading the boundary of the task even if the user is mapping/validating just one task.

@JorgeMartinezG @ramyaragupathy I would like your help to test if JOSM is called correctly from Safari (it seems that there are some CORS problems).

willemarcel:
Added also a modal to let the user know when he has already a task locked on the same or in another project.

xamanu:
Rebased it on latest `develop` and checked it, it works well - didn't test it with safari. There are some minor details:

* Some tests are failing
* Can we preselect as a fallback iD editor for mapping in case a user hasn't specified a default editor in the profile or the pm on the project?
* On the one task view, I don't get the zoom into the task anymore.

I had also difficulties to reopen the task in case anything goes wrong or I accidentally close the editor. So I'm wondering whether we want to have a way to open the task again from the task view.

willemarcel:
@xamanu I'll fix the tests 

> Can we preselect as a fallback iD editor for mapping in case a user hasn't specified a default editor in the profile or the pm on the project?
I'll handle it on https://github.com/hotosm/tasking-manager/issues/2109, we had some confusion between the values we use on the db to the editors.

> On the one task view, I don't get the zoom into the task anymore.

I've noticed something similar and just opened an issue to it: https://github.com/hotosm/tasking-manager/issues/2112

Issue #2057, opened 2019-12-12
Avoid sending out emails about tasks states without comment


xamanu:
We received input in the chat:

> hey team, i’m starting to get complains about the volume of automated emails from some of our more active mappers.  if a task is validated, but there is no mapper feedback, do we really need to email all the mappers who worked on those squares and let them know?
> for example, when we have a task that’s almost finished, we might ask a more experienced mapper to finish up all the last little bits which might mean just reviewing and marking as completely mapped. that could take say 20 mins.  if we are starting validation right after, they are ending up with 5+ automated emails a day from the TM in the following week and starting to gloss over them. this means they’ll miss actual important emails due to overload.

Ideally we would continue having this as a notification, but not sending emails out for this minor information. Can we do this as a hotfix for TM3 and for TM4?

d3netxer:
I don't think there should be any automated messages or notifications, at least until there are settings where users can adjust what notifications they want to receive.

xamanu:
Probably we do want to send emails in the meanwhile when people actually write to somebody. In our case this is a direct mention or comment on a task one has been mapping on. People seem to expect that directed communication would go through

fitoria:
![image](https://user-images.githubusercontent.com/9713/70798145-cb6f5a00-1d6b-11ea-923f-d6fb2a498d12.png)

It's configurable on TM3 maybe the way to go is by creating a better default on this case for new users and do a mass update for existing ones?

xamanu:
I wouldn't enter the topic of making this configurable in TM3. Then in TM4 (#2058) we can plan and figure out the best way of offering this to the users.

PR #2083, opened 2019-12-23
add more options to configure email notifications

fitoria:
Adds new field to user profile to avoid sending emails on
validation state change as detailed on: https://github.com/hotosm/tasking-manager/issues/2057

fixes: 2057


willemarcel:
That issue was solved in another PR

PR #2089, opened 2019-12-24
update nodejs to 13.x on circle-ci

willemarcel:


willemarcel:
We found that probably the reason why some frontend tests were failing was an bug on some nodejs versions: https://github.com/facebook/jest/issues/8769#issuecomment-560034405

So I updated the node version on the build job of circle-ci to 13.x

@dakotabenjamin @eternaltyro could you take a look and check if it's ok doing this update? Do you think we could update the deploy node version too?

PR #2090, opened 2019-12-25
Create Contact frontend

### LINKED ISSUE #1635: Create contact form
Create a form to allow users to send general inquiries to the main admin of the Tasking Manager.

Design reference: https://projects.invisionapp.com/share/VCQTFR6YSJR#/screens/357652189

thadk:
closes #1635. 

Needs POST-capable backend on `contactUs` for `{"name":"Thad","email":"contact@openstreetmap.org","description":"An example message"}`

image
image

image

### DIFF
--- a/frontend/src/App.js
+++ b/frontend/src/App.js
@@ -24,6 +24,7 @@ import { SelectTask } from './views/taskSelection';
 import { MapTask, ValidateTask } from './views/taskAction';
 import { EmailVerification } from './views/verifyEmail';
 import { ProjectEdit } from './views/projectEdit';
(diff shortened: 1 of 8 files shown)


willemarcel:
@xamanu what backend solution to send emails do you prefer to it? Add an endpoint that sends an email to a configured email address? Use some external service?

xamanu:
I think the easiest implementation would be to just send out the email.

willemarcel:
@ramyaragupathy @xamanu I added a new endpoint that allows us to send messages to the system admin. It requires TM authentication. If the user is not logged in, we call his email client with a link.

PR #2091, opened 2019-12-26
Add redux provider to taskSelection/footer tests

willemarcel:


willemarcel:
Merging this without an additional review as it only changes the tests

PR #2093, opened 2019-12-26
serialize team dto for project update

### LINKED ISSUE #2086: Project update fails when it's associated with a team
When making a PATCH update request to a project that is associated with a team, it fails. I can see this error on the backend console:

`[2019-12-23 08:52:33,840] CRITICAL in resources: Project PATCH - unhandled error: 'teamId'`

JorgeMartinezG:
This PR closes #2086.

### DIFF
--- a/frontend/src/components/deleteModal/index.js
+++ b/frontend/src/components/deleteModal/index.js
@@ -10,7 +10,7 @@ import { DeleteButton } from '../teamsAndOrgs/management';
 import { Button } from '../button';
 import { AlertIcon } from '../svgIcons';
 
-export function DeleteModal({ id, name, type, className }) {
+export function DeleteModal({ id, name, type, className }: Object) {
   const token = useSelector(state => state.auth.get('token'));
   const [deleteStatus, setDeleteStatus] = useState(null);
   const [error, setErrorMessage] = useState(null);
--- a/frontend/src/components/deleteModal/messages.js
+++ b/frontend/src/components/deleteModal/messages.js
@@ -1,7 +1,7 @@
 import { defineMessages } from 'react-intl';
 
 /**
- * Internationalized messages for use on header.
+ * Internationalized messages for use on delete modal.
  */
 export default defineMessages({
   projects: {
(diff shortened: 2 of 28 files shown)


willemarcel:
Added some docstrings to endpoints that were missing and added the role field to the team-project creation.
We should modify the permissions in order to match with: https://github.com/hotosm/tasking-manager/issues/2099

willemarcel:
@JorgeMartinezG to solve this issue we need:

- [ ] modify the projectTeams field on the project detail endpoints to return something like `"teams": [ { id: 1, name: "My team name", "role": "MAPPER" }, { id: 2, name: "Other team", "role": "VALIDATOR" } ]`
- [ ] Make that field read only, so it will not be used on the project update operation. as we manage the team-project assignment with other endpoints.

JorgeMartinezG:
@willemarcel  i've just updated and rebased with latest develop. it is now included the frontend component. However I would like a backend review for this line: https://github.com/hotosm/tasking-manager/commit/5dfb193a6fd5bd0045f4df173330f236a2e6b69e#diff-2c8b707bbb933d57e9f6c708e9edb5e5R448

To be more specific the second condition, which I removed in the commit: Basically, the **get_project_teams** method  returns a dictionary with data read from the **teams** attribute. So, this will always be true if and only if there are elements within the teams attribute. If there are no teams associated with the project, it is impossible to associate teams with projects, since the condition is `[] != []`.

willemarcel:
@JorgeMartinezG I reviewed the work done until now and it worked well! On the frontend I think we could add some placeholders or labels on the teams form to show the user what select is the team, what is the role and, on the organisation select, change the placeholder to something like "Filter teams by organisation"

JorgeMartinezG:
@willemarcel placeholders included and modified placeholder for orgs

willemarcel:
@JorgeMartinezG thanks, I'm working on it

willemarcel:
@JorgeMartinezG I've made some improvements to the project edit in general. I also replaced the Organisation select by `react-select` and add a query to filter the organisations the user is part of. Would you like to take a look on it before we merge?

JorgeMartinezG:
@willemarcel good to go

PR #2105, opened 2019-12-31
Send arbitrary changeset tags to an editor

taylorsmock:
Currently only JOSM supports arbitrary changeset tags (since JOSM r15316 -- for reference, 78% or so of current JOSM users are on r15492+, as of 12/23/2019, stats taken from 12/01/2019 to 12/23/2019).

Future potential use cases:
* Add `import=yes`
* Add `tm_url=tasks.hotosm.org`, `tm_project=1`, and `tm_task=5` to the changeset.

This pull request also changes `LANDUSE` to `LAND_USE` (there was only one instance of the former) and updates `black` and `flake8` to the current latest versions.

~I was unable to test remote control, since it appears to not currently be present in TM4.~
EDIT: I implemented JOSM remote control (see #2106 ) so I could test that it worked properly.

For documentation, see https://josm.openstreetmap.de/wiki/Help/RemoteControlCommands#load_and_zoom

### DIFF
--- a/frontend/src/components/projectEdit/instructionsForm.js
+++ b/frontend/src/components/projectEdit/instructionsForm.js
@@ -41,8 +41,8 @@ export const InstructionsForm = ({ languages }) => {
         <input
           className={styleClasses.inputClass}
           type="text"
-          value={projectInfo.changesetComment}
-          name="changesetComment"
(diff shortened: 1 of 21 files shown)


xamanu:
Rerolled this on latest `develop` and with merged commits: #2229

PR #2106, opened 2019-12-31
Implement JOSM remote control

taylorsmock:
This includes and depends upon #2105.

This ~adds~ _modifies_ the calls to `openEditor.js` for JOSM. ~It does want to open the URL's in separate tabs which will trigger pop-up blockers.~

Also, is the new URL for GPX layers stable? ~If so, I'll file a patch on JOSM Trac.~ _I've filed a bug on JOSM trac (~https://josm.openstreetmap.de/ticket/18504~), but I would like to know if the URL is stable._

I do have an open bug with patch to take the name from a GPX file, when a name isn't otherwise provided (see https://josm.openstreetmap.de/ticket/18494 ), so I do not need to do what TM3 did in regards to naming the GPX layer, assuming that the patch is pulled.

I also modified the GPX generation to reduce console spam from JOSM (`INFO: Waypoint time value unset`). Its not a super-serious issue, but it is something that was relatively easy to change.

### DIFF
--- a/frontend/src/components/projectEdit/instructionsForm.js
+++ b/frontend/src/components/projectEdit/instructionsForm.js
@@ -41,8 +41,8 @@ export const InstructionsForm = ({ languages }) => {
         <input
           className={styleClasses.inputClass}
           type="text"
-          value={projectInfo.changesetComment}
-          name="changesetComment"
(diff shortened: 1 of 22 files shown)


xamanu:
Rerolled this on latest `develop` and with merged commits: #2229

PR #2124, opened 2020-01-03
Move API Version to a environment variable

ramyaragupathy:
* Include API Version in Cloudformation
* Update server config to pick TM_APP_VERSION
* Update frontend config to pick TM_APP_VERSION
* Handle BASE URL with a trailing slash


ramyaragupathy:
sure, please go ahead @dakotabenjamin

PR #2125, opened 2020-01-03
add team membership view

willemarcel:

### DIFF
--- a/frontend/src/App.js
+++ b/frontend/src/App.js
@@ -17,7 +17,7 @@ import { Welcome } from './views/welcome';
 import { Settings } from './views/settings';
 import { Management } from './views/management';
 import { ListOrganisations, CreateOrganisation, EditOrganisation } from './views/organisations';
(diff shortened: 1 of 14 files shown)


willemarcel:
Screenshot_2020-01-06 Tasking Manager

Added views to list the user's teams and another that allows the user to join or leave a team.

xamanu:
Thanks @willemarcel!

Just checked it an in principle it works very good and is a useful page. I have a few comments:

* I would like to add  the organization to the list of "my teams". Especially outside of the context of an organization it is useful to have.
* The "Add" button leads to page not found. Here I'm wondering whether it wouldn't be better to take it out. As long as there is an "Add" button I would expect in this section I could also manage the teams, but when I click on them I get to a regular team view, and not the manage view.

I encountered some problem, of everybody being deleted from the team. Trying to reproduce this again. In any case this seems to be unrelated, or more a bug type.

PR #2146, opened 2020-01-08
change user's default_editor values

willemarcel:
* use the same values on frontend for the user default_editor option and the project editors for mapping and validation 
* define the selected option automatically according to the user preference and with the options available to the action the user is taking.


willemarcel:
@JorgeMartinezG @ramyaragupathy whenever you have some time, can you check and propose a solution to the migration. Maybe we could merge it first to improve the user experience and then fix the data on the db

PR #2154, opened 2020-01-10
frontend review

willemarcel:
- Don't check for new notifications if the user is not logged in
- Remove unneeded code on My Contributions
- Compile translation messages
- Prettier code


willemarcel:
> IMO, a better option would be to make the check before performing the function call. Like:
> if (token) { UseFetch(...) } and if (token) { UseFetchIntervaled(...) }
> With this, we would not need to define a default parameter to the functions. What do you think?

It would be a good solution, but the problem is that the hooks can not be called inside an `if`

JorgeMartinezG:
I see. However, It might be confusing for people to send the token variable as trigger, having the token being read inside the _useFetch_ and _useFetchIntervaled_ hooks. What do you think about this?

Within frontend/src/components/header/notificationBell.js we can do this: Before calling the hooks, we can make the validation, like this
```
let trigger = true;
if (!token) { 
  trigger = false;
}

// then we can call here the hooks with the trigger variable.
useFetch(url, trigger);
useFetchIntervaled(url, trigger);

```

willemarcel:
@JorgeMartinezG yes, it sounds much better!

willemarcel:
@JorgeMartinezG I pushed a solution based on your idea

Issue #2157, opened 2020-01-10
TM4 console errors on home page


ramyaragupathy:
Noted the following console errors while loading https://tm4.hotosm.org:

`Mixed Content: The page at 'https://tm4.hotosm.org/' was loaded over HTTPS, but requested an insecure XMLHttpRequest endpoint 'http://tm4.hotosm.org/api/v2/projects/queries/featured/'. This request has been blocked; the content must be served over HTTPS.`

`SyntaxError: Unexpected token < in JSON at position 0` at `stats.js: 69`

cc @willemarcel

ramyaragupathy:
`Mixed Content: The page at 'https://tm4.hotosm.org/' was loaded over HTTPS, but requested an insecure XMLHttpRequest endpoint 'http://tm4.hotosm.org/api/v2/projects/queries/featured/'. This request has been blocked; the content must be served over HTTPS.` -  fixed through #2186

ramyaragupathy:
Still seeing `SyntaxError: Unexpected token < in JSON at position 0` at `stats.js: 69`

willemarcel:
@ramyaragupathy could you paste here the response you're getting on `/api/v2/system/statistics/`?
It's failing to respond on staging server and returning that error, but it works on my machine.

ramyaragupathy:
Here you go @willemarcel 

```
{
    "mappersOnline": 77,
    "totalArea": 35201389.11327167,
    "tasksMapped": 2219998,
    "tasksValidated": 1140807,
    "totalMappers": 204760,
    "totalValidators": 9968,
    "totalProjects": 7233,
    "totalMappedArea": 8370361.934781266,
    "totalValidatedArea": 4211516.536374155,
    "totalOrganizations": null,
    "totalCampaigns": 0,
    "organisations": null,
    "campaigns": [
        {
            "campaign": "Unassociated",
            "projectsCreated": 7233
        }
    ]
}
```

willemarcel:
It's not happening on production neither in staging.

PR #2186, opened 2020-01-13
fix featured api url / solves #2157

willemarcel:


ramyaragupathy:
Don't see the following error anymore on homepage:

`Mixed Content: The page at 'https://tm4.hotosm.org/' was loaded over HTTPS, but requested an insecure XMLHttpRequest endpoint 'http://tm4.hotosm.org/api/v2/projects/queries/featured/'. This request has been blocked; the content must be served over HTTPS.`

Issue #2158, opened 2020-01-10
Map centroid loading inconsistent


ramyaragupathy:
Sometimes see the following message in the project load page within Chrome browser and the map centroid does not show up. I was able to see the centroid on the map once but after reload, there was no centroid and there was this console error.

``Error {message: "Input data given to 'tasks-centroid' is not a valid GeoJSON object."}
message: "Input data given to 'tasks-centroid' is not a valid GeoJSON object."
__proto__: Object``

xamanu:
This has been merged. Let's see whether it is fixed. In the contrary case, please reopen.

PR #2187, opened 2020-01-13
Assure centroid data is present before adding map layer

willemarcel:
solves #2158

### DIFF
--- a/frontend/src/components/taskSelection/map.js
+++ b/frontend/src/components/taskSelection/map.js
@@ -187,7 +187,7 @@ export const TasksMap = ({
         });
       }
 
-      if (map.getSource('tasks-centroid') === undefined && taskBordersMap) {
+      if (map.getSource('tasks-centroid') === undefined && taskBordersMap && taskCentroidMap) {
         map.addSource('tasks-centroid', {
           type: 'geojson',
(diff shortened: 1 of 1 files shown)


willemarcel:
@ramyaragupathy are you seeing the icon on the map? Could you check if the `aoiCentroid` field is present and filled on `/api/v2/projects/<id>/statistics/`

ramyaragupathy:
No there is no icon on the map and I still see the console error. Also centroid is returned through stats endpoint.

willemarcel:
@ramyaragupathy does that error happen on all projects? Could you investigate if it occurs only with projects with a big number of tasks or a complex project/tasks geometry?

willemarcel:
@xamanu could you test this PR, so I can have an additional error/fix case?

ramyaragupathy:
> @ramyaragupathy does that error happen on all projects? Could you investigate if it occurs only with projects with a big number of tasks or a complex project/tasks geometry?

Randomly tested projects and it happens every time

xamanu:
I had difficulties reproducing the issue. Code looks good, and with @ramyaragupathy we decided to merge it it, then we see whether others still have this problems during testing.

Issue #2191, opened 2020-01-15
Better error handling when user can not lock a task


willemarcel:
I tried to lock a task for validation that was mapped by me. It's returning a 500 status code with the following content `{"Error": "Unable to lock task"}`.

On the backend console I can see:

```
127.0.0.1 - - [15/Jan/2020 09:08:05] "OPTIONS /api/v2/projects/9/tasks/actions/lock-for-validation/ HTTP/1.1" 200 -
[2020-01-15 09:08:05,871] CRITICAL in actions: Validator Lock API - unhandled error: USER_ALREADY_HAS_TASK_LOCKED
```

It should return a 403 status code with more information about why the task couldn't be locked. Something like: `{"Error": "You can not validate a task mapped by you."}`.

When the user has not the permissions to map/validate a task it should be: `{"Error": "You don't have the right permissions to map this project."}` or `{"Error": "You don't have the right permissions to validate this project."}`. That way we can give the right information to the user.

willemarcel:
@ramyaragupathy #2206 was a good improvement, but we yet have a generic error message: `{"Error": "Unable to lock task"}`. Do you think it would be possible to add more details to each error case?

PR #2206, opened 2020-01-16
Update locked tasks DTO field check

ramyaragupathy:
From https://github.com/hotosm/tasking-manager/issues/2191 - issue arises because of a missing field check. While [introducing additional fields for user's locked tasks DTO](https://github.com/hotosm/tasking-manager/pull/2107), we have the task list included in a new array field called `locked_tasks`. However [`is_user_permitted_to_validate`](https://github.com/hotosm/tasking-manager/blob/develop/server/services/project_service.py#L230) - still has conditional checks for the old DTO, updated the condition check field.

### DIFF
--- a/server/services/project_service.py
+++ b/server/services/project_service.py
@@ -243,7 +243,7 @@ class ProjectService:
 
         tasks = Task.get_locked_tasks_for_user(user_id)
 
-        if len(tasks) > 0:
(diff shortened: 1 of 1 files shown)


ramyaragupathy:
If a user is a PM or Admin - then this restriction is not applicable, can you confirm that you're a non-admin in your local DB?

PR #2225, opened 2020-01-20
Countries most contributed endpoint

### LINKED ISSUE #2207: List number of tasks mapped and validated per country per user
We need to create a backend endpoint which returns for a given user, the number of tasks mapped and validated. The result must be grouped per country and ordered by the total amount in descending order.

For instance. When hitting `GET api/v2/users/<string:username>/countries-most-mapped/`, the server should return:

```
"countries": [
  {
    "name": "null island",
    "centroid": "0, 0",
    "mapped": 100,
    "validated": 200,
     "total": 300
  }, ...
]
```

Is is required for the user profile page: https://projects.invisionapp.com/share/VCQTFR6YSJR#/screens/357652191

JorgeMartinezG:
Closes #2207 . Removed the centroid from the server response

### DIFF
--- a/server/models/dtos/user_dto.py
+++ b/server/models/dtos/user_dto.py
@@ -91,6 +91,28 @@ class UserDTO(Model):
         return value
 
 
+class UserCountryContributed(Model):
+    """ DTO for country a user has contributed """
+
+    name = StringType(required=True)
+    mapped = IntType(required=True)
+    validated = IntType(required=True)
+    total = IntType(required=True)
+
+
+class UserCountriesContributed(Model):
+    """ DTO for countries a user has contributed """
+
+    def __init__(self):
+        super().__init__()
+        self.countries_contributed = []
+
+    countries_contributed = ListType(
+        ModelType(UserCountryContributed), serialized_name="countries"
+    )
+    total = IntType()
+
+
 class UserStatsDTO(Model):
     """ DTO containing statistics about the user """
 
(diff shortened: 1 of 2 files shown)


willemarcel:
@JorgeMartinezG I tested and it worked well. On `/api/v2/users/{username}/statistics/` we have:

```
{
  "totalTimeSpent": 493687,
  "timeSpentMapping": 483846,
  "timeSpentValidating": 9841,
  "projectsMapped": 4,
  "countriesMapped": [
    "Brazil",
    "Cuba",
    "South Africa",
    "Uruguay"
  ],
  "tasksMapped": 1333,
  "tasksValidated": 5
}
```

Maybe, instead of adding a new endpoint, it would be better to replace the `countriesMapped` field on that endpoint with the new data. What do you think, @ramyaragupathy and @xamanu ?

xamanu:
Are there any performance concerns to bring them all into one endpoint? If not, I would follow @willemarcel suggstion to combine them.

JorgeMartinezG:
Checking the code https://github.com/hotosm/tasking-manager/blob/develop/server/services/users/user_service.py#L268, the service does multiple queries to the database. In my local instance, the endpoint to return the countries and numbers, takes around 900 ms.

JorgeMartinezG:
Rebased and included changes:
```
{
    "totalTimeSpent": 10748,
    "timeSpentMapping": 10407,
    "timeSpentValidating": 341,
    "projectsMapped": 2,
    "countriesContributed": {
        "countries": [
            {
                "name": "Colombia",
                "mapped": 1,
                "validated": 0,
                "total": 1
            },
            {
                "name": "Bolivia",
                "mapped": 1,
                "validated": 0,
                "total": 1
            }
        ],
        "total": 2
    },
    "tasksMapped": 2,
    "tasksValidated": 0
}
```

JorgeMartinezG:
@willemarcel Do you get 200 without this commit?

willemarcel:
> @willemarcel Do you get 200 without this commit?

no, it fails too. Do you think we could fix this issue in another PR?

JorgeMartinezG:
please check https://github.com/hotosm/tasking-manager/pull/2234. It fails in a jest test

PR #2276, opened 2020-02-03
Users contributions from project frontend

JorgeMartinezG:
This PR includes the user contributions for each project.
pDs6YKDKRF

### DIFF
--- a/frontend/src/components/checkCircle.js
+++ b/frontend/src/components/checkCircle.js
@@ -1,8 +1,8 @@
 import React from 'react';
 import { CheckIcon } from './svgIcons';
 
-export const CheckCircle = () => (
-  <span className="br-100 bg-red white h1 w1 ph1 mr2">
+export const CheckCircle = ({ className }: Object) => (
+  <span className={`br-100 h1 w1 ph1 mr2 ${className}`}>
     <CheckIcon height="10px" width="10px" />
   </span>
 );
--- a/frontend/src/components/taskSelection/action.js
+++ b/frontend/src/components/taskSelection/action.js
@@ -347,15 +347,15 @@ function CompletionInstructions() {
         </span>
         <div className="blue-grey">
           <p>
-            <CheckCircle />
+            <CheckCircle className="bg-red white" />
             <FormattedMessage {...messages.instructionsSelect} />
           </p>
           <p>
-            <CheckCircle />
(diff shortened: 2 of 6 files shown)


willemarcel:
@xamanu @JorgeMartinezG Improvements I've made:
- allow to select the tasks by status and update the action button depending on the status
- activate the instructions tab if the current user has not contributed to the project yet
- some style changes on the contributions tab

Would you like to take a look before merging?

JorgeMartinezG:
@willemarcel I have some opinions:

1. When user has a value within mapped or validated higher than 99 happens an overflow within the div. Like this:
image 
We could decrease the font size or increase the width on the div. The issue would be for mapping level going down because of a similar overflow.

2. To highlight tasks, IMO we could optimize a bit: The backend could be modified to return mapped and validated taskids separated. Then, the frontend renders for each user, a handler that receives the id of the array and the option (mapped, validated or total) to get the TaskIds. The function just returns the required ids and updates the map. In this way we do not need a new state variable and the loop to filter among the whole TaskIds array for a given user.

Let me know your opinions

willemarcel:
> When user has a value within mapped or validated higher than 99 happens an overflow within the div. Like this:

@JorgeMartinezG  good catch! I'll test some solutions

> To highlight tasks, IMO we could optimize a bit: The backend could be modified to return mapped and validated taskids separated. Then, the frontend renders for each user, a handler that receives the id of the array and the option (mapped, validated or total) to get the TaskIds. The function just returns the required ids and updates the map. In this way we do not need a new state variable and the loop to filter among the whole TaskIds array for a given user.

Yes, it would be a good improvement... Another problem that I see right now is that we are showing the number of mapped tasks, but we are highlighting only the ones that are yet on the MAPPED state. I choose this way because the goal here is to allow a validator to validate all the tasks. I was thinking if we should change the text to `x ready for validation`. I see that the mapped number is counting tasks locked for mapping and set as bad imagery.

So applying the solution you proposed, I suggest to have an array for tasks that are on the MAPPED state, another that are on VALIDATED state and another to all tasks the user has contributed.

willemarcel:
@JorgeMartinezG I pushed some improvements on responsiveness

JorgeMartinezG:
@willemarcel Good to go.

PR #2280, opened 2020-02-05
Contributions nav in user profile

### LINKED ISSUE #2275: Restructure 'My contributions' section
- My contributions nav bar link should open the current user detail page (the url could be `/contributions/` or `/users/:username/`)
- In the case the current user is seeing the details of themselves, we show a submenu like this above the `HeaderProfile`: 
image
with the links My projects, My tasks, My stats
  - when we click on My project, load the content of `/manage/projects/` with that `HeaderProfile`
  - when we click on My tasks, load the current content of `/contributions/` with that `HeaderProfile`

JorgeMartinezG:
closes #2275

### DIFF
--- a/frontend/src/App.js
+++ b/frontend/src/App.js
@@ -23,7 +23,7 @@ import { Authorized } from './views/authorized';
 import { Login } from './views/login';
 import { Welcome } from './views/welcome';
 import { Settings } from './views/settings';
-import { Management } from './views/management';
+import { ManagementPageIndex, ManagementSection } from './views/management';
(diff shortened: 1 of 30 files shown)


willemarcel:
@xamanu I've made some changes on top of @JorgeMartinezG work. Would you like to take a look on it?
- replaced React Suspense by PlaceHolder on user detail page
- reorganized the management views using nested routes
- reorganized my contributions views using nested routes + added more consistency to the urls
- fixed broken links
- added buttons to project cards in the manage section, so the user can chose to go to project editing

PR #2290, opened 2020-02-07
avoid login popup blocking

willemarcel:
+ increase popup window height


willemarcel:
to test it, I suggest to use Firefox. If you have already allowed your TM dev instance to open popups, remove that permission and then click on the login button.

PR #2292, opened 2020-02-08
update locale select to have all languages that TM supports

willemarcel:


frontend/src/index.js
```diff
@@ -3,13 +3,34 @@ import ReactDOM from 'react-dom';
 import { Provider } from 'react-redux';
 import WebFont from 'webfontloader';
 import { addLocaleData } from 'react-intl';
-import de from 'react-intl/locale-data/de';
+
```
xamanu: Oh wow, we won't have all the languages at the beginning. Planned are Spanish, French, Turkish, Filipino, Indonesian, Swahili to start with.

willemarcel:
@xamanu I just added support to Swahili, Turkish and Filipino (Tagalog) that were missing on TM. Furthermore, I disabled the selection of the languages that we are not supporting in the first release and fixed a bug in the map localization when the locale code is bigger than 2 chars.

willemarcel:
@xamanu I renamed "Tagalog" to "Filipino"

PR #2310, opened 2020-02-12
Include locales for project GET request

JorgeMartinezG:


server/models/postgis/project.py
```diff
@@ -994,6 +994,8 @@ def as_dto_for_mapping(self, locale: str, abbrev: bool) -> Optional[ProjectDTO]:
             project_dto.organisation_name = project.organisation.name
             project_dto.organisation_logo = project.organisation.logo
 
+        project_dto.project_info_locales = ProjectInfo.get_dto_for_all_locales(self.id)
```
willemarcel: @JorgeMartinezG thanks, it worked well. I think we can erase the lines 989 and 990, right?

JorgeMartinezG:
That is a good question @willemarcel . I am not sure if the projectInfo field is used in any part inside the frontend. I will check and reply soon

willemarcel:
@JorgeMartinezG I thought that we are querying the same information two times, but if they are different information, I think we should not remove it

PR #2318, opened 2020-02-14
Fix a few task lock related issues

ramyaragupathy:
* Remove an extra argument from validator function
* ~Include a validator check to handle initial rollout of teams~
* Update error handling for task split

### DIFF
--- a/server/services/project_service.py
+++ b/server/services/project_service.py
@@ -284,10 +284,7 @@ class ProjectService:
 
     @staticmethod
     def evaluate_validation_permission(
-        project_id: int,
-        user_id: int,
-        validation_permission: int,
-        user_mapper_level: int,
+        project_id: int, user_id: int, validation_permission: int
     ):
         allowed_roles = [TeamRoles.VALIDATOR.value, TeamRoles.PROJECT_MANAGER.value]
         # validation_permission = 1(level),2(teams),3(teamsAndLevel)
--- a/server/services/project_service.py
+++ b/server/services/project_service.py
@@ -300,7 +300,8 @@ class ProjectService:
                 for team_dto in teams_allowed
                 if TeamService.is_user_member_of_team(team_dto.team_id, user_id)
             ]
-            if len(user_membership) == 0:
+            # Check for validator role for initial roll out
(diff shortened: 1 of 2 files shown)


server/services/project_service.py
```diff
@@ -303,7 +300,8 @@ def evaluate_validation_permission(
                 for team_dto in teams_allowed
                 if TeamService.is_user_member_of_team(team_dto.team_id, user_id)
             ]
-            if len(user_membership) == 0:
+            # Check for validator role for initial roll out
+            if not (len(user_membership) or UserService.is_user_validator(user_id)):
```
willemarcel: This logic is wrong. If `len(user_membership)` is positive and the user is validator it would return the MappingNotAllowed exception.

Furthermore, we can not support the `is_user_validator` any more, as it would make TM not follow the teams restriction.

server/services/project_service.py
```diff
@@ -325,7 +323,8 @@ def evaluate_validation_permission(
...
```
willemarcel: same here

server/services/project_service.py
willemarcel: Furthermore the frontend is not allowing a validator role user to make that request, so it would not be useful.

server/services/project_service.py
```diff
@@ -284,10 +284,7 @@ def _is_user_intermediate_or_advanced(user_id):
 
     @staticmethod
     def evaluate_validation_permission(
-        project_id: int,
-        user_id: int,
-        validation_permission: int,
-        user_mapper_level: int,
+        project_id: int, user_id: int, validation_permission: int
```
willemarcel: that part is good!

ramyaragupathy:
@willemarcel @xamanu -we handle split service-related errors for the following scenarios:
* Task not locked, so split fails - 403 (Forbidden)
* Attempting to split a task locked by another user - 403 (Forbidden)
* Split failing for some random reason - 500 (Internal server error)
* Split failing because new task geometry does not intersect with the parent task - 200 
 
Let me know if you think we need to update error codes here

PR #2326, opened 2020-02-17
Textstrings English language review

xamanu:


xamanu:
@willemarcel I also removed a couple of lines, as suggested by the reviewers. Can you please check, if this was ok? Thanks.

PR #2327, opened 2020-02-17
Check list endpoints to not return 404 errors

### LINKED ISSUE #2047: Check list endpoints to not return 404 errors
Some endpoints like, `/api/v2/projects/{project_id}/comments/`, `/api/v2/projects/{project_id}/annotations/` are returning 404 errors if there aren't, respectively, comments and annotations.

These endpoints should return 404 code only if the project doesn't exist. If there aren't entries of the entities that should be listed, the result should contain an empty array and a 200 status code. The same is happening in some other list endpoints.

joaovitor3:
Closes #2047 

Fix `Activities` and `Contributions` endpoints which were returning 404 errors if the project exists, but without activity or contribution associated to them.

Following the work done by @fitoria, I've added this if statement in the `ProjectsContributionsAPI` class in the `/server/api/contributions.py` file

```py
class ProjectsContributionsAPI(Resource):
    def get(self, project_id)
        if not ProjectService.exists(project_id):
            return {"Error": "Project not found"}, 404
```

And the same for the  `ProjectsActivitiesAPI` class in the `/server/api/contributions.py` file.

```py
class ProjectsActivitiesAPI(Resource):
    def get(self, project_id):
        if not ProjectService.exists(project_id):
            return {"Error": "Project not found"}, 404
```
https://github.com/hotosm/tasking-manager/blob/4ff9c4c1653b8c3f2b6024b8867f9dcb0b71e840/server/services/stats_service.py#L113-L135

Also, I thought that the Exception `Not Found` found in both `ProjectsContributionsAPI`, `ProjectsActivitiesAPI` classes should be removed, as this part of the code would be unreachable.

https://github.com/hotosm/tasking-manager/blob/4ff9c4c1653b8c3f2b6024b8867f9dcb0b71e840/server/api/projects/activities.py#L37-L38

https://github.com/hotosm/tasking-manager/blob/4ff9c4c1653b8c3f2b6024b8867f9dcb0b71e840/server/api/projects/contributions.py#L34-L35

### DIFF
--- a/frontend/src/components/localeSelect.js
+++ b/frontend/src/components/localeSelect.js
@@ -48,9 +48,6 @@ const mapStateToProps = state => ({
   token: state.auth.get('token'),
 });
 
-const LocaleSelector = connect(
-  mapStateToProps,
-  { setLocale },
-)(LocaleSelect);
+const LocaleSelector = connect(mapStateToProps, { setLocale })(LocaleSelect);
 
 export { LocaleSelector };
--- a/frontend/src/components/taskSelection/contributions.js
+++ b/frontend/src/components/taskSelection/contributions.js
@@ -6,8 +6,6 @@ import { injectIntl } from 'react-intl';
 import messages from './messages.js';
 import { UserAvatar } from '../user/avatar';
 import { CheckCircle } from '../checkCircle';
-import ProjectProgressBar from '../projectcard/projectProgressBar';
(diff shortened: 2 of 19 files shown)


joaovitor3:
Since there were many changes in multiple frontend files and some stylesheets weren't been caught by eslint, I will close this PR and open another one.

PR #2328, opened 2020-02-17
general update of locales files

willemarcel:


willemarcel:
We don't need it, as transifex already handles the update of the files

Issue #2099, opened 2019-12-31
Review project edit permissions


willemarcel:
With the teams & organisations feature, we need to review the permissions to edit a project.

A project should be modified by:
- Any `ADMIN` user
- The user that created the project
- The managers of the organisation assigned to the project
- The team members of a team assigned to the project with the role `PROJECT_MANAGER` (id=2)

PR #2351, opened 2020-02-18
Update project edit permissions 

ramyaragupathy:
From #2099


server/services/project_admin_service.py
```diff
@@ -118,18 +152,18 @@ def update_project(project_dto: ProjectDTO, authenticated_user_id: int):
         if project_dto.license_id:
             ProjectAdminService._validate_imagery_licence(project_dto.license_id)
 
+        is_allowed_user = None
         if project_dto.private:
             ProjectAdminService._validate_allowed_users(project_dto)
+            is_allowed_user = username in project_dto.allowed_usernames
```
willemarcel: Currently the allowed users can only map and validate, right? So on that case, I think we shouldn't give them permissions to edit the project.

willemarcel: On that same file, let's update the `create_draft_project` function and allow also the organization members to create projects.

server/models/dtos/project_dto.py
```diff
@@ -197,12 +195,12 @@ class ProjectDTO(Model):
     )
     mapping_permission = StringType(
         required=True,
-        serialized_name="mapping_permission",
+        serialized_name="mappingPermission",
```
willemarcel: I would suggest to make that change... after your fixes, I'll commit a change in the frontend

server/services/project_admin_service.py
ramyaragupathy: > Currently the allowed users can only map and validate, right? So on that case, I think we shouldn't give them permissions to edit the project.

Right - looking at the current implementation, I can say this allowedUser list is only for mapping/validation. While editing(update_project), there is a check to validate:
* if there is atleast one user in the allowedUser list
* and if the username is valid (present in the DB)

ramyaragupathy: > On that same file, let's update the `create_draft_project` function and allow also the organization members to create projects.

Per discussion letting org managers create projects. Found a few issues on organisation endpoint. Sending that fix in a separate PR.

PR #2354, opened 2020-02-18
Add react matomo components

### LINKED ISSUE #2352: Add piwik-react-router to TM4 for Matomo Tracking
While most of our static sites are able to use the generic tracking script for integrating Matomo, TM4's react frontend requires a more specific solution. I have already found that solution in [OAM-Browser](https://github.com/hotosm/oam-browser/pull/298) and need to port that here. 

Components: 
1. Adding matomo script directly in the index.html
2. Adding a banner component and styles for the opt-out agreement
3. Adding a new variable for the matomo site_id

dakotabenjamin:
This is my first time writing a react component- so any suggestions to improve or do things differently greatly appreciated. 

Resolves #2352 

Based on the generic script used on our other static sites, this PR adds the matomo tracking script with two variables (already present in the deployment from TM3) and a banner component which helps the user opt-in or out of tracking if they wish. 

image


willemarcel:
@dakotabenjamin I pushed some improvements to the style and to the text, replacing the styles by the css classes provided by our css framework and enabling translations.

Are you sure it's working? Do we need to add something to the App.js Router?

Probably we don't need to add the history dependency as on TM4 we use another router that already includes history: https://reach.tech/router/api/createHistory

willemarcel:
@dakotabenjamin I also added a condition to only show the banner if there is a MATOMO_ID env var set, and modified the Privacy Policy link to use the url we have configured by ENV var. This will make it easier to other organisations to customize the use of Matomo.

xamanu:
@dakotabenjamin please rebase.

dakotabenjamin:
@willemarcel @xamanu finally got this rebased and working- I was missing some components. I need to update some environment variables in CircleCI before we merge so let's coordinate after :+1: review.

frontend/package.json
```diff
@@ -36,7 +36,6 @@
     "marked": "^0.8.0",
     "node-sass": "^4.13.0",
     "osmtogeojson": "^3.0.0-beta.3",
-    "react": "^16.13.1",
```
willemarcel: I think it's not good to remove react here. We need to control its version.

dakotabenjamin: sorry I don't think this was intended. Will fix

dakotabenjamin:
@willemarcel I didn't find any issues with the CSS,but also I did not notice the tachyons classes. Do you want me to reconfigure?

willemarcel:
> @willemarcel I didn't find any issues with the CSS,but also I did not notice the tachyons classes. Do you want me to reconfigure?

@dakotabenjamin yes, I would prefer to use the tachyons CSS for it

dakotabenjamin:
@willemarcel done- let's not forget to squash and merge

PR #2361, opened 2020-02-19
[FIX]: This PR fixes the malfunctioning of project search on entering…

### LINKED ISSUE #2358: [BUG]: Search task by project id.
__BUG__:
As we are supposed to enter an integer value in 'project id' placeholder on '/contributions/tasks', if I enter any non-integer value the error thrown is unhandled and the component explodes.
@willemarcel please correct me if I am wrong !
I wanna work on this issue and it would be really helpful if someone guides me how to startoff !

ritvikjain99:
The problem was not with backend, as it returned 500, internal server error, as it was supposed to.
The problem was that in file `taskResults.js` on line 42 `messages.myContributions` is used but not defined in `messages.js`, which fed empty value to a `<FormattedMessage/>` causing it to throw error. !
Fixes #2358 
@xamanu please review!

### DIFF
--- a/frontend/src/components/contributions/messages.js
+++ b/frontend/src/components/contributions/messages.js
@@ -8,6 +8,10 @@ export default defineMessages({
     id: 'mytasks.mainSection.title',
     defaultMessage: 'My Tasks',
   },
+  myContributions: {
+    id: 'mytasks.filter.my_contributions',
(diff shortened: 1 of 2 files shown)


willemarcel:
@ritvikjain99 Thank you very much for finding this error! I'll merge it through #2370

PR #2369, opened 2020-02-20
Improvements on TranslatedMessages

willemarcel:
- rename project interests to categories
- fix pagination count messages
- fix delete modal messages to avoid wrong article gender in other languages


willemarcel:
@xamanu let me know if you think that, under the `Manage` section, we should rename `Interests` to `Categories` too.

xamanu:
Yes, it would be good to call them categories under `manage`, too.

xamanu:
Looks good. I would just rename all to Categories, also in the "Manage" section and reexport the locales please. Then we can merge it. Many thanks.

PR #2375, opened 2020-02-22
stick footer position to bottom

### LINKED ISSUE #2374: [BUG]: Footer position.
__BUG__:
The footer should be at bottom of the page but if there are no projects in 'Explore Project' tab it moves right next to the message 'Showing 0 projects of 0'.

ritvikjain99:
The footer should be at the bottom if the content is shorter than the page. This PR ensures that the footer stays at its ideal position, and doesn't move up if there is shortage of content.
FIxes #2374


ritvikjain99:
@willemarcel please review !

ritvikjain99:
@willemarcel when the page content is smaller the footer should still be at the foot of the page.
That's the ideal behaviour of a footer, which this PR tries to achieve.
Do you want something different ?
Can you please clarify !

ritvikjain99:
@willemarcel I have checked it behaves ideally, that is if content is shorter than page then footer should still stick to the bottom of page !

PR #2384, opened 2020-02-24
[FIX]: User redirected to the url he hit.

### LINKED ISSUE #2376: Redirect to project page after login
Often times in mapathons, people give out the project URL and expect people to sign up or login from there. In TM3 people are redirected after login to the project page again. It would be good to implement this for TM4. Probably only when starting from the project pages only.

ritvikjain99:
@willemarcel forgot to include the `redirectTo` parameter which I think had the same intention.
__ENHANCEMENT__:
If user hits an invalid url then we should redirect him/her to `/welcome` route. For this we need an array of available valid routes. This can be done in two ways:
1. Make a file where all routes are listed as a raw array.
2. Use any third party package which lists the routes.
Please suggest how to do this !
Closes #2376

### DIFF
--- a/frontend/src/utils/login.js
+++ b/frontend/src/utils/login.js
@@ -44,6 +44,7 @@ export const createLoginWindow = redirectTo => {
           username: res.username,
           session_token: res.session_token,
           picture: res.picture,
+          redirect_to: redirectTo,
         }).toString();
         let redirectUrl = `/authorized/?${params}`;
         window.location.href = redirectUrl;


ritvikjain99:
@willemarcel please look at the proposal above it's really necessary.
Coz redirecting a user to wrong route and then throwing 'page not found' is not good.
Rather we should direct him to /welcome !

willemarcel:
@ritvikjain99 I don't see it as necessary. If a user tried to access a page that doesn't exist, they're already seeing the not found page, so they can go away from that page before login.

PR #2410, opened 2020-02-27
Update user role and mapping level in frontend

JorgeMartinezG:


JorgeMartinezG:
@willemarcel I modified the css to show the popup menu in the center for mobile devices.

image

PR #2418, opened 2020-02-29
Fix projects PATCH method

### LINKED ISSUE #2218: PATCH method on projects is not properly working
The PATCH method should allow us to update any individual field on a project.

When I tried to update just these two fields: `{"restrictMappingLevelToProject": true, "restrictValidationRole": true}`, it failed with:

```
[2020-01-17 11:27:44,798] ERROR in resources: Error validating request: {"status": ["This field is required."], "projectPriority": ["This field is required."], "defaultLocale": ["This field is required."], "mapperLevel": ["This field is required."], "private": ["This field is required."], "taskCreationMode": ["This field is required."], "mappingEditors": ["This field is required."], "validationEditors": ["This field is required."]}
127.0.0.1 - - [17/Jan/2020 11:27:44] "PATCH /api/v2/projects/6/?projectStatuses=0 HTTP/1.1" 400 -
```

It's not blocking the project editing with the frontend because we are sending all the projects fields in the request, but we should make it work properly in the future.

joaovitor3:
Closes #2218 

I updated the `ProjectDTO` fields to be optional, set the `ProjectDTO` attributes in the `update` method and add in `if` statement for each field. As seen in the sniped below.

```project_dto.py```
```py
# before
default_locale = StringType(
        required=True, serialized_name="defaultLocale", serialize_when_none=False
)

# after
default_locale = StringType(
        serialized_name="defaultLocale", serialize_when_none=False
)
```

```project.py```
```py
project_dto_data = self._get_project_and_base_dto()[1]
updated_fields = [key for key, value in project_dto.items() if value is not None and value]
[setattr(project_dto, key, value) for key, value in project_dto_data.items() if key not in updated_fields]

if project_dto.project_status:
            self.status = ProjectStatus[project_dto.project_status].value
```

All validations in the `updated_fields` line are necessary, due to some fields have the `False` and empty list values, which wasn't caught by the `if value` condition.


server/models/postgis/project.py
```diff
@@ -382,26 +382,53 @@ def get(project_id: int):
...
+            ].value
+        if project_dto.validation_permission:
+            self.validation_permission = ValidationPermission[
+                project_dto.validation_permission.upper()
+            ].value
+        if project_dto.enforce_random_task_selection:
+            self.enforce_random_task_selection = (
+                project_dto.enforce_random_task_selection
+            )
+        if project_dto.private:
```
willemarcel: that line is avoiding to update the project to `private = False`

joaovitor3: Thanks for the review!

I will try to fix all the problems.

As soon as I find a solution I comment here

joaovitor3: I changed the logic to get the fields to update, removed the if statement for the `changeset_comment` and `private` fields and left a comment explaining that these fields don't have a comparison because it would affect when changed for `False` or an empty string.

Get updated fields line
```py
updated_fields = [
        key
        for key, value in project_dto.items()
        if value is not None and project_dto_data[key] != value
]
```

`changeset_comment` and `private` fields

```py
# Doesn't check if exists because the boolean False when modified affects comparisons
self.private = project_dto.private
if project_dto.mapper_level:
    self.mapper_level = MappingLevel[project_dto.mapper_level.upper()].value
if project_dto.entities_to_map:
    self.entities_to_map = project_dto.entities_to_map
# Doesn't check if exists because empty strings when modified affects comparisons
self.changeset_comment = project_dto.changeset_comment
```

ramyaragupathy:
@joaovitor3 - would be helpful if you could recreate this fix in a separate branch as multiple commits have gone in here and it is hard to evaluate, thank you

PR #2419, opened 2020-02-29
[ENHANCEMENT]: Return empty result instead of error.

### LINKED ISSUE #2111: Improvements to /api/v2/users/{user_id}/tasks/ endpoint
- [ ] add pagination
- [ ] add project name field on the results
- [ ] boolean filter by archived project status

ritvikjain99:
The result is now empty array instead of an exception if user sends project_id which is not number.
Closes #2111  
@xamanu @willemarcel please review !


ritvikjain99:
Sorry for multiple pushes :)

PR #2438, opened 2020-03-04
Edit aoi in project creation view

JorgeMartinezG:


willemarcel:
@JorgeMartinezG The AOI edit part worked well. I noticed yet an issue when we switch the basemap and move the AOI.
image

The grey shadow remains on the same place and the AOI is moved. If we switch the basemap again, the shadow is updated

xamanu:
This is also now having some conflicts. Please check @JorgeMartinezG. Thank you!

JorgeMartinezG:
I fixed the bug related to moving the aoi. Rebased and fixed conflicts generated wrt develop.

PR #2413, opened 2020-02-28
Notif popup

ritvikjain99:
What these changes intend to do:
[Commit 1]: Corrects the wrong datetime format fed to `<FormattedRelative />`.
[Commit 2]: The `useOnClickOutside` function initially checked if the click was outside popup, so clicking the bell was considered out-click. Which toggled the 'isOpen` state twice.

Now if the popup is opened and the bell icon is clicked, it closes !
Fixes #2412 
@willemarcel  please review !

PR #2439, opened 2020-03-04
Remove additional string for date in notification card mini

JorgeMartinezG:
Fixes **invalid date** error in notitfication card

image


ritvikjain99:
@JorgeMartinezG #2413 's commit 1 does the same, also this notif popup has one more problem, which is that it doesn't close when bell icon is clicked again, which is also fixed by the mentioned PR :)

willemarcel:
@JorgeMartinezG Thanks! I noticed some problems with the text inside the notification popup, so I pushed some improvements. Could you just double check if it works well for you?

PR #2460, opened 2020-03-07
remove unused field last_validated_date from user table

### LINKED ISSUE #2448: Remove last_validation_date field from the User model
We have a `last_validation_date` field on the User model that is updated every time a task mapped by that user is validated. We don't use that information and it's one additional database operation that is executed on some task status change.

The only endpoint that uses it is `/api/v2/projects/{project_id}/tasks/queries/mapped/`, but we are not using it on the frontend.

Related to https://github.com/hotosm/tasking-manager/issues/2003

vegeta999:
Closes #2448 .
Changes made in migrations, so user is supposed to upgrade the db first !
@willemarcel please review !!!!!!!!


ritvikjain99:
@ramyaragupathy I have done the requested changes, please review !
Point I wanna make:
--> The purpose behind altering earlier migrations was that if a field is of no use then maintaining a history that we once created it, and now we are deleting it dint make sense to me. Also running migrations has to perform two extra operations.

ritvikjain99:
@ramyaragupathy any further changes am I supposed to make ?

ramyaragupathy:
stale, no next actions

PR #2462, opened 2020-03-08
hash navigation

### LINKED ISSUE #2441: add anchors to each project create and edit sections
Allow the users to navigate throw the project create and edit sections throw hash anchors. For example: `/manage/projects/2#description`, `/manage/projects/2#instructions`.

Assure that if the user press the back button after going to a second section the form data will not be lost and it will just enable the previous section.

ritvikjain99:
Closes #2441 
The desired features are working well!
Just need one help:
I want to call a function whenever the `props.location` changes which I did in changes.
But it throws warnings please suggest what to do ?
@willemarcel


willemarcel:
@ritvikjain99 I believe this can help:

```
  useLayoutEffect(() => {
    const hash = props.location.hash && props.location.hash.slice(1);
    const isValid = elements.some(element => element.value === hash);
    if (isValid && hash) {
      setOption(hash);
    } else {
      setOption('description');
    }
  }, [props.location, elements]);
```

ritvikjain99:
@willemarcel for every re-render of component, the reference for `elements` changed, leading to invoking of above `useLayoutEffect` every time, hence prevented change in option.
To solve this I used caching the `elements` constant using `useMemo`. Now everything works fine, thanks for assistance !

Issue #2454, opened 2020-03-06
Inconsistent totalContributors results in /projects endpoint


willemarcel:
We have two projects on TM4 instance that has not received contributions yet, but the `api/v2//projects` endpoint is returning randomly 7 and 71 for those projects:

![image](https://user-images.githubusercontent.com/666291/76083309-1a541680-5f8c-11ea-9433-d4a3625c4b48.png)

Some additional details:
```
contrib_counts = ProjectSearchService.get_total_contributions(
            paginated_results.items
        )
```

Maybe it would be better if we add the `get_total_contributions` as a method of the `Project` model and replace the code above by something like: `[project.get_total_contributions() for project in paginated_results.items]`

The query on the projects table seems being executed two times without need
- here we query for the projects: https://github.com/hotosm/tasking-manager/blob/develop/server/services/project_search_service.py#L148-L150
- and it's repeated here: https://github.com/hotosm/tasking-manager/blob/develop/server/services/project_search_service.py#L133-L136

PR #2484, opened 2020-03-10
Fix contributor count on  project sort 

ramyaragupathy:
From #2454:
* Introduced a new function to compute per project distinct user ID count
* Commented out earlier method call which returns an ascending list of contributor count irrespective of the sort order in the API call. This resulted in count mixup between projects

### DIFF
--- a/backend/models/postgis/project.py
+++ b/backend/models/postgis/project.py
@@ -827,6 +827,20 @@ class Project(db.Model):
         )
         return project_info.name
 
+    @staticmethod
+    def get_project_total_contributions(project_id: int) -> int:
+
+        project_contributors_count = (
+            TaskHistory.query.with_entities(TaskHistory.user_id)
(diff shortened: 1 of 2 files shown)


server/models/postgis/project.py
```diff
@@ -826,6 +826,18 @@ def get_project_title(self, preferred_locale):
         )
         return project_info.name
 
+    @staticmethod
+    def get_project_total_contributions(project_id: int) -> int:
+
+        project_contributors_count = (
+            TaskHistory.query.with_entities(TaskHistory.user_id)
+            .filter(TaskHistory.project_id == project_id)
```
willemarcel: I think we should exclude from the results the `COMMENT` actions.

ramyaragupathy: Agreed, good catch - implementing the change

PR #2488, opened 2020-03-11
Update language translations

xamanu:
Updating and adding languages from the translations.


server/config.py
```diff
@@ -74,11 +74,11 @@ class EnvironmentConfig:
     SUPPORTED_LANGUAGES = {
         "codes": os.getenv(
             "TM_SUPPORTED_LANGUAGES_CODES",
-            "ar, cs, de, en, es, fr, id, it, ja, mg, ml, nl_NL, pt, pt_BR, sw, tl, tr, uk",  # noqa
+            "ar, cs, de, en, es, fa, fa_IR, fr, he, hu, id, it, ja, mg, ml, nl_NL, pt, pt_BR, sw, tl, tr, uk, zh_TW",  # noqa
         ),
         "languages": os.getenv(
             "TM_SUPPORTED_LANGUAGES",
-            "Arabic, Česky, Deutsch, English, Español, Français, Indonesia, Italiano, 日本語, Malagasy, Malayalam, Nederlands, Português, Português (Brasil), Swahili, Filipino (Tagalog), Türkçe, Українська",  # noqa
+            "Arabic, Česky, Deutsch, English, Español, Persian, Persian (Iran), Français, Hebrew, Hungarian, Indonesia, Italiano, 日本語, Malagasy, Malayalam, Dutch (Nederlands), Português, Português (Brasil), Swahili, Filipino (Tagalog), Türkçe, Українська, Chinese (Taiwan)",  # noqa
```
willemarcel: With most languages we followed the pattern of writing the language name in its own language, like "日本語" and "Español", probably we should follow the same patter to Arabic, Persian, Chinese and all others. The same on frontend/src/utils/internationalization.js

xamanu: Good point. I added language names in their languages.

Issue #2247, opened 2020-01-23
Fix notifications page


xamanu:
I just observed some issues on TM4 with the notifications:

![image](https://user-images.githubusercontent.com/97706/72440716-4380d380-37a1-11ea-9754-69ad67e95588.png)

* [ ] Instead of showing the username it shows some id, when the notification reaches the same user who made the comments
* [ ] Avoid sending a notification if you comment on your project
* [ ] Redirect to notifications page after delete a notification

thadk:
@xamanu I suspect those are back-end bugs but I don't have relevant notification data in my own backend to test. Can you share the JSON response from the notification API from your dev tools?

willemarcel:
> Instead of showing the username it shows some id

I'm noticing that problem as well.

- The backend is creating a message to notify the same user that posted the comment (what is not useful...)
- The problem with the id instead of the user name is on the backend as well

On the frontend we need to replace the username (`@[username]`) on the message body by the user page link and format it as `@username`.

ritvikjain99:
@willemarcel I can do this !

xamanu:
Some more good input from the testers:

> I noticed that when deleting notifications I get redirected to the home page instead of staying 8n the notification page. Also, when new notifications come in there is no indicator showing that there are new notifications. An orange bubble or new notification count would be useful.

xplora1a:
Also having this. This might be sensible in the event of the last notification is deleted, but it also happens if I have remaining messages.

xamanu:
Fixed with #2623

PR #2493, opened 2020-03-11
review username tagging in messages

ritvikjain99:
Now we can tag user using just '@' !
@willemarcel @xamanu  please reveiw !
Related #2247


ritvikjain99:
@willemarcel I noticed that on frontend it's just conversion of @[username] -> **username**, without the check whether the user was actually tagged. Which needs to be changed !
I am working on it, will make a PR soon :)

ritvikjain99:
@willemarcel please review !

tests/backend/unit/services/messaging/test_messaging_service.py
```diff
@@ -7,12 +7,14 @@ class TestMessagingService(unittest.TestCase):
     def test_message_service_identifies_all_users(self):
         # Act
         usernames = MessageService._parse_message_for_username(
-            'Hello @[Iain Hunter] and "[LindaA1]'
+            'Hello @Iain Hunter and "@LindaA1 @@@Cersie @ @Peter_123'
         )
 
         # Assert
-        self.assertEqual(usernames[0], "Iain Hunter")
+        self.assertEqual(usernames[0], "Iain")
```
xamanu: Are we dropping support for usernames with a space in between? Like `Iain Hunter`?

ritvikjain99:
@xamanu As far as I have seen usernames don't have spaces, and underscores are the substitutes for spaces, the reasons I see are:
1. We'll face problem sending username as query parameter.
For instance I have username "ritvik rajesh jain":
Then param would have to be: ?username=ritvik%20rajesh%20jain, which is not so readable.
2. For instance there are three persons having usernames differing just by whitespaces you can't differentiate between two people based on number of white spaces they have in their usernames.
E.g. "   ritvik", "  ritvik", "ritvik ", "ritvik    " , "hobs shaw", "hobs   shaw", "hobs  shaw" , all look the same, yet are different, they are really hard to track.
3. If someone uses tab('\t') in username then it would be really problematic as tab size can be altered.

Please correct me if my arguments are wrong !

willemarcel:
I'm solving it in another PR

PR #2492, opened 2020-03-11
Accomodate org manager permissions

ramyaragupathy:
Endpoints touched:

Class|Verb
-----|-----
ProjectsRestAPI|POST, HEAD, PATCH, DELETE
ProjectsQueriesOwnerAPI|GET
ProjectsQueriesNoTasksAPI|GET
ProjectsActionsTransferAPI|POST
ProjectsActionsMessageContributorsAPI|POST 
ProjectsActionsFeatureAPI|POST
ProjectsActionsUnFeatureAPI|POST
ProjectsActionsSetInterestsAPI|POST
ProjectsCampaignsAPI|POST, DELETE
TasksActionsMapAllAPI|POST 
TasksActionsValidateAllAP|POST 
TasksActionsInvalidateAllAPI|POST 
TasksActionsResetBadImageryAllAPI|POST 
TasksActionsResetAllAPI|POST 
OrganisationsRestAPI|PATCH, DELETE
CampaignsAllAPI|POST
CampaignsRestAPI|PATCH, DELETE
InterestsAllAPI|POST, PATCH, DELETE

PR #2496, opened 2020-03-13
Rename server directory to backend

xamanu:
After we renamed the `client` directory in TM4 to `frontend`, now it is time to do the same for the `server` directory and rename it to `backend`.


xamanu:
Rebased on merged PR #2492

ritvikjain99:
@xamanu project/{id} throws error on my client, please correct me if this is just with me !

PR #2389, opened 2020-02-25
Remotes/origin/bugfix/test function fix

DeboDevelop:
I have changed the name of the conflicting method and run a test. 
The Output is:

Ran 133 tests in 9.385s

FAILED (errors=37)

I need to look further into the errors. This is the solution to #2151 .
Please check whether the problem persist in your side.

PR #2504, opened 2020-03-16
fix tests with duplicated function name

xamanu:
Replaces #2389


xamanu:
The failing of the CI seems not to be related to the introduced changes...

Aadesh-Baral:
Addressed in #5129

PR #2508, opened 2020-03-17
Redirect to projects page after signup.

### LINKED ISSUE #2505: Redirect to project page after sign-up
When a user starts from a specific project page and clicks sign-up, the project should be saved in the database (together with the other pre-saved data such as name and email) and after the first login the user should be redirected to this specific project.

The standard behaviour of the welcome page should persist, if a user comes from any other page, e.g. the explore project page. Then no redirect shall happen.

ritvikjain99:
@xamanu please see if this helps !
Closes #2505

### DIFF
--- a/frontend/src/components/header/signUp.js
+++ b/frontend/src/components/header/signUp.js
@@ -198,7 +198,7 @@ export const SignUp = ({ closeModal }) => {
 
   const login = () => {
     closeModal();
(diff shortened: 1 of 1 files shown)


xamanu:
Sorry, the original issue had no real description, I added it there. The fix is a bit more complex. Closing here. In case you want to take it, please open a new PR.

PR #2509, opened 2020-03-17
Fix mixed up filter query parameters.

### LINKED ISSUE #2172: Notification Page: sort query parameter mixed up
Notification search by project and sorting works fine on the backend. But the frontend seems to be pick wrong parameters.

eg :`/api/v2/notifications/?sortDirection=asc` becomes `/api/v2/notifications/?sortDirection=projects&sortBy=ASC` from the client, so effectively the sorting doesn't happen

ritvikjain99:
Closes #2172 
@willemarcel please review !


xamanu:
Works like a charm. Thank you.

PR #2512, opened 2020-03-17
Embed id editor within Tasking Manager

willemarcel:
Co-Authored-By: Jorge Martínez Gómez <JorgeMartinezG@users.noreply.github.com>


xamanu:
Current status update:

* [x] iD is embedded
* [x] TM shares the OSM authentication with iD (user doesn't have to login twice)

To-dos: 

- [ ] Sidebar toggle
- [ ] Rename "start another editor" to "reload editor"
- [ ] Rename "edit status" to "Submit Task"
- [ ] Move split and select another buttons (padding to "submit task" button)
- [ ] Handle save on "submit task" or other action

Eventually we should look into the cursor not showing up as a cross, when selecting any of the objects (point, line, polygon) for mapping. It would be also good to backport some of the code to iD, in order to support better our use case of embedding it.

Issue #2498, opened 2020-03-14
Non existent users can be tagged


vegeta999:
__BUG__:
Even if the user is non-existent the frontend renders the name under the trigger characters `@[<username>]` in bold with link to a non-existent user's profile page.
Also the comment rendered is that input by user, which should rather first be sent to backend then rendered

__PROPOSED SOLUTION__:
We can send an array of tagged and verified usernames from backend for each comment, which in turn is verified at frontend whether the name tagged actually exists . And we render it accordingly !

![coment_bug](https://user-images.githubusercontent.com/34532173/76681493-6c92d880-6619-11ea-85c8-8e7262478ebf.png)

PR #2529, opened 2020-03-19
Preventing the tagging of non-existent users.

vegeta999:
This PR prevents the tagging of non-existent users, changes made:
1. Everytime method `get_message` is called, it parses the usernames tagged, checks whether they exist and sends a list of validated usernames to frontend along with message.
2. Frontend while link-ifying the username checks whether this username is present in validated list.

I thought of a bit optimized approach in which we store the tagged usernames in db for every message, and just iterate over usernames(rather than whole message) to check existence on every GET.
Also I felt checking user's existence for every GET is necessary because the user might not be existing while comment was created but exists now !

@ramyaragupathy @willemarcel please review !


ritvikjain99:
Closes #2498

xamanu:
Hey, thanks for this! Tested this and behaviour is fine.

I agree that we somehow have to cover the edge case of people linking to users that appear later on. However I do think this is the edge case, and doing this check on each `get_message` seems to be quite complex and i fear this has negative effects on the performance.

I'm wondering whether it would make sense to do the following:
* When saving a comment check if a @[...] to a non-existing user is inside the text. If this is the case set a flag on the db for this message.
*  Only run the check within `get_message` on the ones that have this flag set, but not on all.

@ramyaragupathy, @willemarcel what do you think is a good solution for this problem?




we can only do the check, if the message contains a not existing user

ritvikjain99:
@xamanu, thanks for your attention. I have one more solution:
Add a many-to-many field to message to the linked users, simple, clean, and fast.
Also the above solution has better average time complexity, but same worst case complexity.

PR #2544, opened 2020-03-22
Add 404 handling to /projects/{id} and /users/{id}

### LINKED ISSUE #2499: Add handling for route /user/{id} and 404.
If you hit the url on frontend with a non-existent user-id the page keeps loading for infinite time.
The profile page with wireframes is rendered and never unmounted, user never gets to know that user is non-existent.

For reference:
ezgif com-video-to-gif(3)

ritvikjain99:
Closes #2499


xamanu:
This is much better, thanks @ritvikjain99 ! Can we make this look like our other "page not found" styles:

image

xamanu:
Please excuse. This was a mistake on my side. It looks great. merging now. Thanks again!

PR #2548, opened 2020-03-23
Consistency in table names

ramyaragupathy:
* Introduce a migration to rename tables:
    * `users_license` -> `user_license`
    * `projects_interests` -> `project_interests`
    * `users_interests` -> `user_interests`

### DIFF
--- a/migrations/versions/2d6a7ad4e1eb_.py
+++ b/migrations/versions/2d6a7ad4e1eb_.py
@@ -1,7 +1,7 @@
 """empty message
 
 Revision ID: 2d6a7ad4e1eb
-Revises: c40e1fdf6b70
+Revises: 14340f1e0d6b
 Create Date: 2020-03-23 12:10:20.120500
 
 """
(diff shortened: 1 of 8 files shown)


migrations/versions/2d6a7ad4e1eb_.py
```diff
@@ -0,0 +1,23 @@
...
+
+# revision identifiers, used by Alembic.
+revision = "2d6a7ad4e1eb"
+down_revision = "c40e1fdf6b70"
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    op.rename_table("users_licenses", "user_licenses")
```
xamanu: Shouldn't this also migration `projects_interests`?

migrations/versions/2d6a7ad4e1eb_.py
```diff
@@ -0,0 +1,23 @@
...
+
+
+def downgrade():
+    op.rename_table("user_licenses", "users_licenses")
```
xamanu: same as above

PR #2562, opened 2020-03-24
Fix task lock API errors

ramyaragupathy:
* There was an argument mismatch in `get_task_details_for_logged_in_user` function resulting in 500 on `tasks/queries/own/locked/details`
* Removed the project ID from the url since we're implementing an instance wide check now. It is not just tied up with projects


backend/__init__.py
```diff
@@ -453,7 +453,7 @@ def add_api_endpoints(app):
     )
     api.add_resource(
         TasksQueriesOwnLockedDetailsAPI,
-        format_url("projects/<int:project_id>/tasks/queries/own/locked/details/"),
+        format_url("projects/tasks/queries/own/locked/details/"),
```
xamanu: I think the `projects/tasks/`isn't fitting that well into our current practice of naming the API. But this is also a special case. I would consider naming it:

* `/api/v2/users/queries/tasks/locked` (similar to the existing `/api/v2/users/queries/favorites/`)
* `/api/v2/projects/all/tasks/queries/own/locked/details/`(includes `all` for the resource)

I kind of prefer the first one. But would also move both api endpoints into the users section. What do you think, @ramyaragupathy ?

willemarcel: @xamanu I like `/api/v2/users/queries/tasks/locked`. Another solution would be `/tasks/queries/own/locked/details/`

xamanu: Yes, both would work. But as we have already  another `/api/v2/users/queries/...` so i would go with this one.

ramyaragupathy: `/api/v2/users/queries/tasks/locked` seems nifty and consistent with our naming convention. Going with this

ramyaragupathy:
Changes made:

* ProjectFavoriteAPI -> ProjectsFavoritesAPI
* TasksQueriesOwnLockedAPI -> UsersQueriesOwnLockedAPI
* TasksQueriesOwnLockedDetailsAPI -> UsersQueriesOwnLockedDetailsAPI
* UserFavoritesAPI -> UsersQueriesFavoritesAPI
* UserInterestsAPI -> UsersQueriesInterestsAPI
* UserRecommendedProjectsAPI -> UsersRecommendedProjectsAPI

PR #2584, opened 2020-03-24
Set environment variable for contact form

### LINKED ISSUE #2581: Fix email from contact fornm
The email is sent according to the TM_EMAIL_FROM_ADDRESS environment variable. But the contact send to address is different than the from address (noreply@.... usually). Lets introduce another environment variable here and separate the two.

JorgeMartinezG:
closes #2581

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -112,6 +112,8 @@ jobs:
         type: env_var_name
       email_from_address:
         type: env_var_name
+      email_contact_address:
+        type: env_var_name
       smtp_host:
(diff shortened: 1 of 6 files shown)


xamanu:
Please have a look on the tests, @JorgeMartinezG

JorgeMartinezG:
Fix for tests included . @xamanu  I think it would be necessary to set the environment variable in the CF instance cc @dakotabenjamin

xamanu:
@dakotabenjamin can you please set the variable for the CI?

PR #2596, opened 2020-03-27
Fix projects favorited by me filter

### LINKED ISSUE #2593: Project favorites filter
`projects/?favoritedByMe=true` filter should return results based on auth token. However, at present, this filter is returning all projects in published state.

joaovitor3:
Closes #2593 

Fix `favorited_by_me filter, which was always returning all projects in `PUBLISHED` state.
All the discussion is located in the [issue](https://github.com/hotosm/tasking-manager/issues/2593).

### DIFF
--- a/backend/services/project_search_service.py
+++ b/backend/services/project_search_service.py
@@ -148,7 +148,6 @@ class ProjectSearchService:
         all_results, paginated_results = ProjectSearchService._filter_projects(
             search_dto
         )
-
         if paginated_results.total == 0:
             raise NotFound()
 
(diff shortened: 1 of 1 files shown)


willemarcel:
@joaovitor3 thank you, but it didn't solve the problem... it continues to return all projects if a user does not have favourite projects.

When the request is not authenticated and the `favoritedByMe=true` is on the url, we should not return projects.

joaovitor3:
Thanks for the feedback!
I will try to fix this and as soon as I finish I comment here.

joaovitor3:
Thank you @ramyaragupathy and @willemarcel

Issue #2577, opened 2020-03-24
Mark tasks as impossible to map


xamanu:
When you click on “It is not possible to map this task” the task doesn’t change color and stay white on the map. This one should become grey, check also if it is not selectable anymore.

ritvikjain99:
@xamanu could you tell which roles have rights to mark it impossible?

willemarcel:
@xamanu I tested it on develop now and it seems to be fixed.

Issue #2572, opened 2020-03-24
Rename “select another task” to “stop validating” on validation


xamanu:
Feedback from the validators:

> I have to quit but cannot find that option during validation . Only Validate and mark as finished or More mapping is needed.

JorgeMartinezG:
Issue solved in https://github.com/hotosm/tasking-manager/pull/2600

PR #2600, opened 2020-03-28
change elements of map/validate sidebar

willemarcel:
Change reload editor, completion instructions and add a tasks map modal (only when iD is enabled)


willemarcel:
Also resolves #2577 and #2572

PR #2602, opened 2020-03-29
Fix resume mapping and task status updation.

### LINKED ISSUE #2577: Mark tasks as impossible to map

### LINKED ISSUE #2601: Task actions are not woking properly.
If you mark any of the three options in task and submit it, the action shows 'Resume mapping' even if the task is marked as unavailable or mapped. But after refresh the they work fine.

ritvikjain99:
Closes #2601 : 
After mapping task as either of ['BADIMAGERY', 'READY', 'MAPPED'] the variable `lockedTasks` must update. But the function `fetchLockedTasks` which is responsible for the above, is never invoked due to which `Resume Mapping` wasn't working.

Closes #2577 : 
**Initially**, the api calls were as follows : 
- BADIMAGERY: hit `/stop-mapping` and restore the previous status.
- READY: hit `/stop-mapping` and restore the previous status.
- MAPPED: hit `unlock-after-mapping` and unlock it.
Here `/stop-mapping` prevented the updation in status to ['BADIMAGERY', 'READY'] coz it just restored the previous status.

@xamanu @willemarcel please correct me if I missed something or am wrong !

### DIFF
--- a/frontend/src/components/taskSelection/action.js
+++ b/frontend/src/components/taskSelection/action.js
@@ -20,6 +20,7 @@ import { TaskHistory } from './taskActivity';
 
 import DueDateBox from '../projectcard/dueDateBox';
 import { UserFetchTextarea } from '../projectDetail/questionsAndComments';
+import { useFetchLockedTasks } from '../../hooks/UseLockedTasks';
 
 const Editor = React.lazy(() => import('../editor'));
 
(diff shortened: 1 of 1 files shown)


ritvikjain99:
@willemarcel I just wanna confirm if the api calls are correct or should they be amended ?

ritvikjain99:
Some of the targeted changes are achieved by #2600 which hits `/stop-mapping` on selecting the option `Task partially mapped, but needs more work` .
Which is also achieved by the changes above. But the issue of 'Resume mapping' still persists.
If the above fix isn't satisfactory we are good to close this !

ritvikjain99:
@willemarcel does this work fine of should we think of other solution ?

willemarcel:
Thanks, @ritvikjain99

PR #2628, opened 2020-04-02
Optimize autoscaling for TM4

### LINKED ISSUE #2557: Configure Autoscaling to use a smaller threshold for tracking metric
Current threshold is 600 Requests per Target in 1 minute intervals. Users experience slow loading and 500 errors before this threshold is reached. I have tested changing to 250 requests which seems to work, but I need to do a bit more load testing to tune that number.

dakotabenjamin:
Closes #2557

After running some load tests on the TM4 instance, I think the following configuration will work well for the release- it should more than handle a user load similar to the current capacity on TM3, with room to improve on after launch. 

Production Capacity (number of instances):
- min 2
- max 9

Autoscaling Tracking Target:
RequestsPerTarget: 500

Cooldown: 300s

The cooldown reduction from 600 should speed up the autoscaling process.

### DIFF
--- a/scripts/aws/cloudformation/tasking-manager.template.js
+++ b/scripts/aws/cloudformation/tasking-manager.template.js
@@ -11,7 +11,7 @@ const Parameters = {
   AutoscalingPolicy: {
     Type: 'String',
     AllowedValues: ['development', 'demo', 'production'],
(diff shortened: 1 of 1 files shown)


dakotabenjamin:
@xamanu please merge *after* tomorrow's mapathon

PR #2633, opened 2020-04-03
Add clear filter button to users list

ritvikjain99:
@JorgeMartinezG some edits to the code you authored, please review !


ritvikjain99:
@JorgeMartinezG would be good if clear button is only visible only if filters are different from initial filters. Shall I that that ?

willemarcel:
@ritvikjain99 Thank you very much! I pushed some additional changes on #2647

Issue #2614, opened 2020-03-31
html title


xamanu:
Improve title shown in browser tabs:

![image](https://user-images.githubusercontent.com/97706/78052913-50a86a00-736f-11ea-958c-04780eb78445.png)

* [x] Make html title configurable additionally to title (may differ because of the logo). Environment variable with the default/fallback `Tasking Manager`
* [x] Include environment variable for our deployment with `HOT Tasking Manager`
* [ ] Fix the title tag when displaying iD to have the same value

willemarcel:
With https://github.com/hotosm/tasking-manager/pull/2637, I couldn't prevent iD of setting the title tag, but I set the title tag on all project pages. So the title tag is updated when the user finished mapping on iD and go back to the task selection page.

I also added the username to the title tag when on `/users/username/`. I propose we set different title tags to each page, like, explore, about, learn, my contributions... It's not difficult to do it

xamanu:
Wille is going to check on titles in general to be meaningful.

xamanu:
Closing this. For the one remaining item i created a follow up issue: #2658

PR #2637, opened 2020-04-04
Set title tag in all views

willemarcel:
Solves partially https://github.com/hotosm/tasking-manager/issues/2614


xamanu:
Giving back to @willemarcel for some additional work on top here. Thanks!

PR #2644, opened 2020-04-06
Project and Team doesn't update some values 

### LINKED ISSUE #2643: Project and Team doesn't  update some values
In the manager section:
In Projects:
When you want to update the "Types of mapping" with the `landuse` option it doesn't work, all the other options work fine .
In teams:
When you want to update the `Organisation` with other value it doesn't update.

NoemiNahomy:
when changing the key from "LANDUSE" to "LAND_USE" use the one that was most used in the project.
Closes #2643
Please review @xamanu or @JorgeMartinezG


xamanu:
The `LAND_USE` change is verified and works. The other change needs some more review. I suggest we split the PR into two, so each issue is treated separately. @NoemiNahomy can you do that?

xamanu:
@NoemiNahomy Thanks a lot. Can you please squash the commits into one?

PR #2656, opened 2020-04-06
Include actions made by users on each task

### LINKED ISSUE #2177: Distinguish between actions and comments on My Tasks page
On My contributions page, if a user has only posted a comment to a task, it shows the text: "You worked on this task...". We should distinguish between task status changes and comments on this text.

JorgeMartinezG:
closes #2177


JorgeMartinezG:
Yes. What is happening is that for a comment task, it involves three rows in the database. For a task without, it will be two rows. COMMENT will always be the second insertion and STATE_CHANGE will be the third one. So, returning the last action will always be state_change no matter if it was commented or not. This is why I return all actions that the user has performed within a task

xamanu:
Closing as discussed in the dev meeting today.

PR #2665, opened 2020-04-07
Fix project filters

### LINKED ISSUE #2663: Filters on Manage Projects
When logging in as an organization admin, and then create a project. For example leave it in `draft` state and go to `Manage`->`Projects` the filters aren't showing the project under `Created by me`, nor `Draft`.

Short description of what is expected for each of the filters:

* Active (all active and published projects a user can manage)
* Draft (all draft projects a user can manage)
* Archived (all archived and published projects a user can manage)
* Created by me (all projects created by one user no matter which state these are in)

ritvikjain99:
Closes #2663 

Here is the description of changes made:
1. The `manageByMe` filter should also include the **projects created by a person**. They were missing, I appended them.
2. The `Active` tab should list 'PUBLISHED' projects as per the instructions by @xamanu in #2663, I added the status filter to query param.

Also if no status is mentioned then all projects should be listed irrespective of statuses rather than just the 'PUBLISHED' ones that is **no status param means no status filter**.
@xamanu Please correct me if I am wrong !

### DIFF
--- a/backend/services/project_search_service.py
+++ b/backend/services/project_search_service.py
@@ -192,14 +192,12 @@ class ProjectSearchService:
         if search_dto.project_statuses:
             for project_status in search_dto.project_statuses:
                 project_status_array.append(ProjectStatus[project_status].value)
-        else:
-            project_status_array = [ProjectStatus.PUBLISHED.value]
+            query = query.filter(Project.status.in_(project_status_array))
 
         if search_dto.interests:
             query = query.join(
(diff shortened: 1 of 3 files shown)


ritvikjain99:
@willemarcel there is some problem with the newly added tests could you please checkout I am unable to solve it.
Related to env variable !

willemarcel:
It's integrated on https://github.com/hotosm/tasking-manager/pull/2666

PR #2371, opened 2020-02-21
Add loader

ritvikjain99:
Initially all the loading and rendering after loading happening was visible to user.
After addition of LOADER before any loading related activity, especially LOGIN the loader is visible.
This loader can be extended to use with fetching of 'my tasks' and 'my contribution' related data.
Fixes #2167 
Before Loader:

ezgif com-video-to-gif-1

After Loader:

ezgif com-video-to-gif

PR #2681, opened 2020-04-08
Add loader to iD editor

### LINKED ISSUE #2680: Add a loading indicator while iD is loading
While iD is loading we could show a loading indicator instead of a blank area.

vegeta999:
Closes #2680 
Added preloader to the iD editor.
Further enhancements:
1. We can give the `transition` property a thought to avoid sudden disappearance of the loader.
2. This is similar to the loader I suggested to add to login process, so if in future we add it then these files should be edited to avoid redundancy, mentioning it here #2371 .

@willemarcel please review !


xamanu:
@willemarcel has this been fixed with #2753?

ritvikjain99:
@xamanu I'll have to do some changes in this to import component made in the PR you mentioned, I'll get it done today !

ritvikjain99:
I have incorporated the preloader of 2753 to id editor, works fine !

ramyaragupathy:
stale

PR #2682, opened 2020-04-08
Remove GET project cache

### LINKED ISSUE #2669: Caching on Manage Projects page
If a user changes the status of a project (from published to draft or from draft to published) it still shows up as the previous state in the `Manage Projects` page.

ritvikjain99:
Fixes #2669 
Removed the project cache. 
Question: Should it be removed or shall we decrease the time of caching ?
@xamanu please review !


ramyaragupathy:
A cache should be present - and we have a 30s cache. Seems like the results persist beyond that and this has to be investigated. Closing it here, thanks for looking at this @ritvikjain99

PR #2702, opened 2020-04-09
change spelling of 2 words to be consistent

wireguy:
prefered changed to preferred

organisation changed to organization
's' changed to 'z',  there were 2 strings that used 's' and 22 strings that used 'z', so changed the 2 strings to be consistent with the other 22.   Note, I did not change any string names or variables with the name orgaisation


willemarcel:
Replaced by #2707

Issue #2710, opened 2020-04-12
Check on provided social media links in user profile


wireguy:
Would be nice if the LinkedIn field:

- could parse the input and strip everything up to the last '/' 
users seem to grab the url of their linkedin id and paste it as input causing tm to make the link:
https://www.linkedin.com/https://www.linkedin.com/gg
instead of 
https://www.linkedin.com/gg

- would be nice if the url could be validated (pinged?) and inform the user if the link is bad (gets a 404)

xamanu:
Lets check here on other social media links, too.

eternaltyro:
Twitter link does not show `https://` explicitly. The rest of the social media links do.

willemarcel:
I added a validation on the user settings form that avoid users from inserting urls https://github.com/hotosm/tasking-manager/pull/3684

PR #2718, opened 2020-04-13
Check on provided social media links in user profile

Nikitag0329IITkgp:
This is for issue https://github.com/hotosm/tasking-manager/issues/2710.


xamanu:
I think it would be better to check the input than the output. This is causing an error when visiting the "My contributions" page: `TypeError: Cannot read property 'split' of undefined`

ritvikjain99:
@Nikitag0329IITkgp you should add handling for the case when the value is empty that is when we want to return base URL itself and dont wanna append anything at the end !

Nikitag0329IITkgp:
Hey, @xamanu can you have a look at it now?

ramyaragupathy:
@Nikitag0329IITkgp - would you be able to create a separate PR here as this branch is unmanageable now with too many commits to follow?

PR #2727, opened 2020-04-14
Add error handling while creating and updating organisation

### LINKED ISSUE #2717: Show error message on Organisation creation
If for some reason the organisation creation fails, it's breaking the app. Instead, it should show a message on the footer with the reason the request failed.

ritvikjain99:
Currently the error message is visible in alert modal.
@willemarcel please suggest whether this is fine should I be thinking about some error message box.
Closes #2717


ritvikjain99:
@willemarcel I have done it like you wanted just wanna know how do we show the message describing the error. The error received in catch just gives the technical description like BAD REQUEST, INTERNAL SERVER ERROR.
PS: I have checked this works fine in mobile view !

ritvikjain99:
The same occurred if I updated the org name to the one that already exists in 'edit organisation' section, fixed it !

HelNershingThapa:
The issue has been addressed in https://github.com/hotosm/tasking-manager/pull/5252.

PR #2735, opened 2020-04-15
Prevent creation of campaign with same names

### LINKED ISSUE #2728: There can be multiple campaigns with same name
I can create multiple campaigns with same name, not really sure about whether this is feature or not.
If this is intended then we should have an distinction between two with same name.

Screenshot_2020-04-14 Tasking Manager
 @xamanu please confirm this so that I can solve if this is really an issue!

ritvikjain99:
Closes #2728 
This PR aims to prevent creation of multiple campaigns with same name and handle the error on frontend.
To achieve this I have applied a check to ensure that campaign with same doesn't exist.
The same goes for the patch method so that user cant set the name same as an already existing campaign.
Current behavior:

ezgif com-video-to-gif

@xamanu @willemarcel please review !


ramyaragupathy:
Moved it to https://github.com/hotosm/tasking-manager/pull/3119

Issue #2622, opened 2020-04-01
Style issue on the  project manage page


NoemiNahomy:
There are some minor styling issues on the project manage page( `/manage `):

- The link "go to tasks" has less height than "edit project"
![project_manage](https://user-images.githubusercontent.com/1789542/78153351-f78d1480-7408-11ea-9433-7e5a5e102435.jpg)

- In the equipment section, there is no space between the properties: "Invite only" and
"Public", but this only happens in languages ​​other because the text is longer.

![image](https://user-images.githubusercontent.com/1789542/78154925-d4636480-740a-11ea-8021-686acc6f8252.png)

NoemiNahomy:
Also in the manager in the team section using the **all** filter, you can see there is a little style problem.
Url: `https://tm4.hotosm.org/manage/teams/?manager={id}`

![imagen](https://user-images.githubusercontent.com/1789542/78512268-0d843780-7771-11ea-8084-bdfe95e2ff3b.png)

Issue #2563, opened 2020-03-24
Styling issue on the project overview page


xamanu:
There are some minor styling issues on the project overview page:

------

![image](https://user-images.githubusercontent.com/97706/77432516-13276800-6dd6-11ea-91a1-d159cd95b8b6.png)

------

* It seems text is cut off above "TYPES OF MAPPING" icons, if the short description is too long.
* The scrollbar doesn't make much sense/looks good.
* The line spacing /styling is also different in the description at the top vs. below under "Description"

See [testing project #18](https://tm4.hotosm.org/projects/18) for an edge case example.

xamanu:
Another small issue about a strange distortion of profile pic on the project page:

![image](https://user-images.githubusercontent.com/97706/77447780-f7c55880-6de7-11ea-8a28-feb0ad5c9b71.png)

LeenasPingveenas:
> 
> 
> Another small issue about a strange distortion of profile pic on the project page:
> 
> ![image](https://user-images.githubusercontent.com/97706/77447780-f7c55880-6de7-11ea-8a28-feb0ad5c9b71.png)

It also allows me to drag text box over other content:
![Capture](https://user-images.githubusercontent.com/58010959/77448832-2c85df80-6de9-11ea-87ac-bd56c5bb2455.PNG)

LeenasPingveenas:
Bug that allows text box to be dragged over other UI items.

![Capture](https://user-images.githubusercontent.com/58010959/77449008-622ac880-6de9-11ea-984b-cfb67ad4ac93.PNG)

ritvikjain99:
@xamanu comment box has been amended in #2585 .

willemarcel:
Most issues were solved. What's missing:

- Detect when the short description is too long for the screen space and assure the "Read more" button will be always visible.
- The organization information is mostly not visible. We could add it on the end of the page or next to the teams information.

xamanu:
Fixed the distortion of profile pic problem in #2739

willemarcel:
This issue has been solved.

PR #2739, opened 2020-04-16
improve TeamCards and UserAvatar style

willemarcel:
- avoid distortion on UserAvatar
- show correct number on UserAvatarList (contributors section on project detail page)
- makes height of TeamCard fixed in order to avoid break the lines

Solves partially #2622


willemarcel:
It also solves some of the problems described on #2563

PR #2749, opened 2020-04-17
Fix add button evaluating to 0

ritvikjain99:
If user doesn't have privileges to add project, '0' is rendered instead of 'Add Project' button.
Without Admin and orgManager privileges:

add-project-0


willemarcel:
Fixed by https://github.com/hotosm/tasking-manager/pull/2821

Issue #2679, opened 2020-04-08
Task activity modal skip to another task when tasks are refreshed


willemarcel:
> I opened the 'Task Activity' for proj 40 task 25, and looked at my comment. just waited there about 30 seconds, didn't do anything, and then the content changed to task 21's content

Issue #2678, opened 2020-04-08
Task refresh error when a task is splitted by other user


willemarcel:
I just let the task selection page opened for some minutes, when I returned I saw a blank page with that error:

```
TypeError: "l.features.filter(...)[0] is undefined"
    gi taskList.js:341
    gi taskList.js:333
    React 7
    unstable_runWithPriority scheduler.production.min.js:19
    React 4
    unstable_runWithPriority scheduler.production.min.js:19
    React 3
    F scheduler.production.min.js:17
    onmessage scheduler.production.min.js:14
```
The error points to that line:

`tasksGeoJSON.features.filter(task => task.properties.taskId === item.taskId)[0].geometry`

I could reproduce that problem and it's caused by a task split made by other user, what makes the task id disappear.

PR #2751, opened 2020-04-18
Solves errors caused by the periodic tasks updates

willemarcel:
Resolves https://github.com/hotosm/tasking-manager/issues/2679
Resolves https://github.com/hotosm/tasking-manager/issues/2678

Please review carefully.

Wait 60s to check if the tasks get updated (both list and map). With another user in another browser window, start mapping a task and check if the update on the first browser occurs correctly. Split a task on the second window and see if the map and list get correctly updated on the first.

Do the same with a task activity modal open.

### DIFF
--- a/frontend/src/components/taskSelection/index.js
+++ b/frontend/src/components/taskSelection/index.js
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, Suspense } from 'react';
+import React, { useState, useEffect, useRef, Suspense } from 'react';
 import { useSelector, useDispatch } from 'react-redux';
 import Popup from 'reactjs-popup';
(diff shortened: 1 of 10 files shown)


xamanu:
it works fine on the tasks list. but i was able to observe some problems:

* task history modal doesn't update if I mark it as "Ready for validation"
* task history modal reloads to a white screen if i split the task one is looking at
* contributions tab doesn't update on new information

willemarcel:
> task history modal reloads to a white screen if i split the task one is looking at
I have solved it... the remaining issues needs more time to work.

xamanu:
Just created a follow up issue for the remaining issues: https://github.com/hotosm/tasking-manager/issues/2794

PR #2759, opened 2020-04-20
add perTaskInstructions + handle better multiple tasks validation

willemarcel:
- add perTaskInstruction info on mapping/validation page
- hide history tab when validating multiple tasks
- format markdown content correctly across all system


xamanu:
This issue is connected to #2754

PR #2771, opened 2020-04-21
Organisation logo upload

willemarcel:
Add endpoint and update frontend to allow organisation logo image upload

@ramyaragupathy could you review the backend part (and also the frontend if you want :)?

@dakotabenjamin I added two ENV VARs to configure the upload service, if there is the need to change the cloudformation template, you can add commits here.


xamanu:
Ping @eternaltyro

PR #2772, opened 2020-04-21
fix project detail page redirect

willemarcel:
@xamanu do you think it would be good to add also a redirect from /contribute to /explore?


xamanu:
Yes! This would be nice!

willemarcel:
@xamanu it's updated!

Issue #2251, opened 2020-01-24
Change count logic on /api/v2/projects/{id}/contributions/queries/day/


willemarcel:
I was testing the map all, invalidate all, validate all and reset all functions on this project. So I mapped all one time, invalidate them, then mapped all, validated all and then reset the status of all tasks to ready to map. This endpoint `/api/v2/projects/{id}/contributions/queries/day/` is returning that result:

` {"stats": [{"date": "2020-01-24", "mapped": 224, "validated": 224, "cumulative_mapped": 224, "cumulative_validated": 224, "total_tasks": 112}]}`

We should count the number of tasks that were mapped/validated, not the number of state changes on the tasks.

joaovitor3:
Hi
I'm trying to start contributing to the project, could I pick this issue or this is too hard for a new contributor?

willemarcel:
@joaovitor3 welcome to tasking manager!

Maybe it's a bit complex to understand the tasks history... I think those two issues would be better for a new contributor:

- https://github.com/hotosm/tasking-manager/issues/2132
- https://github.com/hotosm/tasking-manager/issues/2047

joaovitor3:
Thanks for the quick response @willemarcel, I will look at these two issues.

About being a new contributor, I would like to enter the HOT slack chat as mentioned in the [working groups section](https://github.com/hotosm/tasking-manager/blob/develop/docs/working-groups.md), but after answering the form and receiving the link to access the channel, the [invitation link](https://hotosm.slack.com/join/shared_invite/enQtNDkyMDA4MjQ1MDEyLWNmMjU4ZmEyNzI5MTc2MmJkODU5ZTA0ZmFhYWMwZmYyZjdhYThkMTI2OWJjNDZmMTlhMjBlYzYwMjdmYjNjZDg) does not seem to be active.

Am I doing something wrong? If not, should I open an issue about it?
Thanks for the feedback

willemarcel:
@joaovitor3 could you try to register again at http://slack.hotosm.org/? The invitation link has been updated.

joaovitor3:
I tested the new link and it's properly working.

I am already in the hot slack chat, thanks!

willemarcel:
Another problem I noticed on tm4 instance:

On that project: https://tm4.hotosm.org/projects/18, the contributions chart shows almost 50% of mapping completion, but we only have 17 tasks mapped and 1 validated.

[...]

PR #2774, opened 2020-04-21
Fix daily and absolute count values on contribution endpoint

ramyaragupathy:
From https://github.com/hotosm/tasking-manager/issues/2251 - updated cumulative  and per day count logic for validated and mapped tasks.

So for the example in the 2251 ticket in which tasks wer `mapped all one time, invalidate them, then mapped all, validated all and then reset the status of all tasks to ready to map` - stats would return:

```
{"stats": [{"date": "2020-01-24", "mapped": 0, "validated": 0, "cumulative_mapped": 0, "cumulative_validated": 0, "total_tasks": 112}]}
```

Project 8213 on this branch:
image

Project 8213 on staging:
image

### DIFF
--- a/backend/services/project_service.py
+++ b/backend/services/project_service.py
@@ -1,4 +1,3 @@
-import datetime
 from cachetools import TTLCache, cached
 from flask import current_app
 from backend import db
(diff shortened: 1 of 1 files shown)


xamanu:
Just realized @willemarcel had brought this up and has surely much better ways to review it. Moving this to you. If you want me to look at it, please just assign it back. thanks.

PR #2780, opened 2020-04-21
improvement to the installation script

### LINKED ISSUE #2779: improvement to script instalation
After installing TM it is loaded with the test data.
For a new installation, I suggest that these lines be commented by default.

NoemiNahomy:
Closes #2779 
Review please!!

### DIFF
--- a/scripts/install/install_on_debian_10_buster.sh
+++ b/scripts/install/install_on_debian_10_buster.sh
@@ -57,13 +57,13 @@ fi
 
 
 # Populate database
-cd tests/database/ &&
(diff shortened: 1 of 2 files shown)


xamanu:
Can you describe a bit the thinking around the improvement here? 
If we want to give an empty database shouldn't we just remove the commands and maybe make sure they are in the docs? Or do we want them to be in the script?

NoemiNahomy:
I think it could be documented in [setup-development](https://github.com/hotosm/tasking-manager/blob/develop/docs/setup-development.md ) and as an example keep the commented lines

PR #2782, opened 2020-04-21
Improve project search backend

JorgeMartinezG:
Fixes local search for projects.

**How to test**
- In explore page, use different keywords or project id in the search box to get different projects that match


ritvikjain99:
One more thing, currently the project search match whole keywords like if I have typed 'abc' then project named abcd is not shown but when I type abcd it gets retrieved. I tried to fix it but couldn't, thought it should come to your knowledge !

PR #2784, opened 2020-04-22
allow admin, org managers & PMs to review validated and badimagery tasks

### LINKED ISSUE #2583: Review state of an already marked task
For power validators sometimes it is necessary to check an already validated task . f.i. When a non experienced mapper validates tasks I feel the need to check the task to see if the validating has been properly. In projects where no validation qualification is required  you find often such situations. Also when  a validator makes a mess of it,( I've seen some) it makes sense to check their work.

This issue is to add a no possibility to review a already validated task.

willemarcel:
Resolves #2583


xamanu:
Just had a look and i expected them to be in the task activity modal, but nothing shows up. Tried with a validator and an admin and with expert more - but wasn't able to spot the functionality.

willemarcel:
Changed permission of `/projects/{projectId}/tasks/actions/undo-last-action/{taskId}/` to allow the user that made the last action + users that can validate the project.

On the frontend, only users that can manage the project are allowed to review validated and badimagery tasks

PR #2798, opened 2020-04-22
Make image upload service envvars variables instead of strings

dakotabenjamin:
I originally had the environment variables be passed into the JSON file as strings instead of variables. 

@eternaltyro do we have a better solution for this for the CodeDeploy work? I have now on multiple occasions made an unforced error on this line of code.


eternaltyro:
@dakotabenjamin I would investigate AWS Parameter Store (Part of SSM) . I think that's the best way forward. The other would be AWS config but I've never used it .

PR #2801, opened 2020-04-23
Update task count on undo

ramyaragupathy:
Per chat w/ @willemarcel, there is a discrepancy in numbers for a project on TM4 instance.

* Project 31 has a total of 83 tasks - 
     * 76 validated
     * 2 bad imagery
    * 3 mapped
That gets the percentages to - 95% mapped, 91% validated and 2% bad imagery

However the project ID endpoint retruns:

```
"percentMapped": 87,
"percentValidated": 96,
"percentBadImagery": 4,

```

Because  validated  + bad imagery = 100%, contribute button is not enabled on the frontend.

On closer investigation, `update_stats_after_task_state_change` method invokes the `_update_tasks_stats` only if TaskStatus is not in READY, LOCKED_FOR_MAPPING and LOCKED_FOR_VALIDATION. For undo eg: when a BADIMAGERY is set to READY, per current logic this function is not invoked and hence the bad imagery count on projects table never gets decremented resulting in inflated percentage values.

For existing projects, I have a script integrated to updated tasks count on projects table, using tasks current status. I have already run this against our TM4 Database. Queuing this to our TM3 deployment list. cc @dakotabenjamin @xamanu 

Also updated the  denominator for percentage computation in project/id GET endpoint.

State|Current % logic for project 31 | Percentage after changes in this PR
------|-----------------------------|-------------------------------------
Mapped |97|95
Validated|94|91

@willemarcel and I will discuss this in the working group meet today to arrive at an agreement on excluding bad imagery tasks from validated count.

### DIFF
--- a/backend/services/stats_service.py
+++ b/backend/services/stats_service.py
@@ -43,7 +43,6 @@ class StatsService:
         """ Update stats when a task has had a state change """
 
         if new_state in [
-            TaskStatus.READY,
             TaskStatus.LOCKED_FOR_VALIDATION,
             TaskStatus.LOCKED_FOR_MAPPING,
         ]:
--- a/frontend/src/components/projectDetail/index.js
+++ b/frontend/src/components/projectDetail/index.js
@@ -158,6 +158,7 @@ export const ProjectDetailLeft = (props) => {
             className="pb2 bg-white"
             percentMapped={props.project.percentMapped}
             percentValidated={props.project.percentValidated}
+            percentBadImagery={props.project.percentBadImagery}
           />
           <div className="cf pb1 bg-white">
             <MappingLevelMessage
--- a/frontend/src/components/projectcard/messages.js
+++ b/frontend/src/components/projectcard/messages.js
@@ -20,6 +20,10 @@ export default defineMessages({
     id: 'project.card.percentValidated',
     defaultMessage: 'Validated',
   },
+  percentBadImagery: {
+    id: 'project.card.percentBadImagery',
+    defaultMessage: 'Bad imagery',
+  },
   projectPriorityURGENT: {
     id: 'project.card.projectPriorityUrgent',
     defaultMessage: 'Urgent',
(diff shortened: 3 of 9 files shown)


ramyaragupathy:
From the working group meeting today:

_Is Bad Imagery considered validated?_
* _No, validation is a metric of the mapping_
* _Bad Imagery should remain as part of the total # but not validated/mapped_

xamanu:
I think the bad imagery ones should be taken out of the percentage calculations.

We are looking into the progress bar of a project. And I would consider it completed with 100%. Leaving the projects with 90% something, is very confusing, when going through the list of projects. A mapper would have to go into each one of them, to see if there is a last task that requires mapping of leave disappointed because all the remaining tasks are classified as `bad imagery`.

Of course, in project statistics we want to line out all types of states. But the progress bar should show how much of the work that actually can be done, has been done.

willemarcel:
@xamanu I agree with you!

giblet22:
Thanks for this conversation. I think @xamanu makes a good point about a mapper's expectation when selecting a project. The other side of that is a validator's expectation when selecting a project to review. 

I think as long as %bad imagery is readily available on the project page, and the %mapped and validated are correct relative to each other and the total _map-able_ tasks that can be a good compromise. Example:

- Project has 500 tasks
- 50 tasks are marked as bad imagery (10% bad imagery, 450 tasks remain)
- 400 tasks are mapped (88.8% mapped of 450 tasks)
- 200 tasks (of those 400 mapped) are validated (44.4% validated of 450 tasks) 

From the validator's perspective they need to know what percentage is bad imagery (so they can confirm and ensure it was not incorrectly marked). So if the validator confirms it was actually fine to be mapped, and can be reset, the denominator of tasks would increase by however many are reset to their original state.

willemarcel:
@xamanu I added the bad imagery information to the progress bar tooltip. I'm not sure if we should user "Bad imagery" or "Unavailable". Let me know what do you think...

PR #2816, opened 2020-04-28
Docker improvements

JorgeMartinezG:
For tasking manager 4, since frontend and backend are separated within the project, this PR makes two dockerfiles with their own runtimes and environments. Also, the docker-compose file includes a reverse proxy configurable within the same file.
**Backend:** Image size reduced from 1.0Gb to 400Mb
- Modified python runtime from buster to python-3.7-alpine.
- main command set to gunicorn.

**Frontend:**
- The development dockerfile comes with node.
- The production dockerfile compiles the source code to a bundle which is served by nginx.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,6 +13,8 @@ build:
 	docker-compose build --no-cache app
 
 up:
+	docker network inspect tm-web >/dev/null 2>&1 || \
+		docker network create tm-web
 	docker-compose up -d
 
 down:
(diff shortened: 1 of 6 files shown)


xamanu:
Getting this error on a first `make up`:
> Error response from daemon: could not find an available, non-overlapping IPv4 address pool among the defaults to assign to the network

Issue #2833, opened 2020-04-30
I can't create a private project


NoemiNahomy:
When I activate this option, I cannot save the project because of a server error.

In the console: Must have at least one allowed user on a private project

I can't assign at least one user to a project

![Screenshot(5)](https://user-images.githubusercontent.com/1789542/80722637-9753c600-8acd-11ea-9a64-aaba40c5f62b.png)

ramyaragupathy:
Related https://github.com/hotosm/tasking-manager/issues/2817

PR #2838, opened 2020-05-01
remove allowed_users validation when setting a project as private

willemarcel:
Resolves https://github.com/hotosm/tasking-manager/issues/2833


xamanu:
Thanks for the backend code, @willemarcel . Works well.

PR #2848, opened 2020-05-04
Fixes in project creation view

JorgeMartinezG:
- Fixes bug that does not render taskgrid on a  basemap change in trim project and set project name steps for project creation.
- Fixes the taskgrid not reset after being clipped or trimmed and **previous** button is clicked.


willemarcel:
@JorgeMartinezG It works well. I found another issue (it's happening on develop too): https://github.com/hotosm/tasking-manager/issues/2850

PR #2849, opened 2020-05-04
add video background to main banner section

### LINKED ISSUE #1959: Switch to animated banner for frontpage
For the new redesign it was planned to use this banner on the frontpage: 

Banner-Map-a-task

This issue is to include it. Do we need the static image as a fallback until the gif loads?

willemarcel:
Closes #1959 

- use a video if the screen width is higher than 824px

- improve images used on the homepage
  - set a better compression
  - load low or high image resolution depending on the browser window width

The video URL is set with the `TM_HOMEPAGE_VIDEO_URL` environment variable. On HOT instances we need to set it as `https://cdn.hotosm.org/tasking-manager/mapping.mp4` (cc @dakotabenjamin )


xamanu:
Minor styling detail on intermediate resolutions - i added a red square to visualize

image

willemarcel:
@xamanu I pushed a solution to that issue.

PR #2854, opened 2020-05-05
Add TM Homepage Video URL

eternaltyro:
Add link to homepage banner video.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -154,6 +154,8 @@ jobs:
         type: env_var_name
       org_github:
         type: env_var_name
+      org_homepage_video_url:
+        type: env_var_name
       log_level:
         type: enum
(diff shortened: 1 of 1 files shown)


example.env
```diff
@@ -26,6 +26,7 @@ TM_ORG_FB=https://www.facebook.com/example
 TM_ORG_INSTAGRAM=https://www.instagram.com/example/
 TM_ORG_YOUTUBE=https://www.youtube.com/user/example
 TM_ORG_GITHUB=https://github.com/example/
+# TM_HOMEPAGE_VIDEO_URL=
```
xamanu, line 29: This is already in the `example.env` (around 20 lines further down) together with its description text

ramyaragupathy: sorry didn't notice, thanks for fixing

PR #2862, opened 2020-05-05
Include private project filter correctly

### LINKED ISSUE #2861: the filter is also returning private projects
In explorer project (https://tasks-stage.hotosm.org/explore) , the filter is also returning private projects, but I can't see any more details of the project.

JorgeMartinezG:
Closes #2861. To test this PR, you would need to disable server cache.

**how to test:**
- Using the example db dump.
- Go to explore page.
- Set projects as private. ie
```update projects set private=true where id in (1, 2);```
- You will not see projects 1 and 2.
```update projects set private=false where id in (1, 2);```
- You will see projects 1 and 2.

### DIFF
--- a/backend/services/project_search_service.py
+++ b/backend/services/project_search_service.py
@@ -83,7 +83,7 @@ class ProjectSearchService:
                 Organisation.name.label("organisation_name"),
                 Organisation.logo.label("organisation_logo"),
             )
-            .filter(Project.private is not True)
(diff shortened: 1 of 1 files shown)


willemarcel:
On the manage section, we need to see the private projects. With this change. I can not see it, but I think it's more important to hide from general users than show it to managers. So, we can merge it and then improve.

Issue #2887, opened 2020-05-07
Project Search: backend failure while using managed by me filter


ramyaragupathy:
From logs noticed this error - `Project GET - unhandled error: (psycopg2.errors.DuplicateAlias) table name "project_teams" specified more than once`. On further investigation this is because of an additional union statement added to the query filter and it is breaking the project search.

Sending a PR

ramyaragupathy:
Fixed through #2888

PR #2888, opened 2020-05-07
Remove author filter from manage section on project search

ramyaragupathy:
From #2887 - removed the redundant filter using author ID


ramyaragupathy:
Enable managed_by filter on Projects GET in the current branch. It should return results. On tasks.hotosm.org this fails.

Issue #2817, opened 2020-04-29
Give access to private projects for all users in an assigned team


ramyaragupathy:
Expecting an input option for allowed user list on private projects.
![image](https://user-images.githubusercontent.com/12103383/80571486-53ec4f80-8a1a-11ea-9be1-8e8c9224c9ad.png)

![image](https://user-images.githubusercontent.com/12103383/80571520-61093e80-8a1a-11ea-966c-42890f0ec590.png)

russdeffner:
Shouldn't this be handled with Teams in TM4? Or at least, add individually or by team?

xamanu:
Yes, i think i remember we said this should be handled by teams, and we support the list of users but only in the backend. Still we need a way to add teams to a private project.

willemarcel:
> Shouldn't` this be handled with Teams in TM4? Or at least, add individually or by team?

Yes, we made the decision to remove the allowed users field from the project edit form.

> Still we need a way to add teams to a private project.

@xamanu we can add teams the same way we do to public projects, can't we?

xamanu:
As far as I see, if a user is part of a team that is assigned to the project, s/he can still not access the project, but this is what I would assume as the behaviour in TM4.

willemarcel:
Resolved by https://github.com/hotosm/tasking-manager/pull/2949

PR #2924, opened 2020-05-11
Validate user role to include private projects in explore page

### LINKED ISSUE #2906: Private projects hidden from Admins and Managers
If I put a project as Private, even as the owner and creator - it does not show up in Explore or Manage sections of the TM, the only way it seems possible to return to it is if you know the project URL. We need to allow admins, project managers and owners to search for private projects.

JorgeMartinezG:
closes #2906

### DIFF
--- a/backend/services/project_search_service.py
+++ b/backend/services/project_search_service.py
@@ -14,6 +14,7 @@ from backend.models.postgis.statuses import (
     MappingLevel,
     MappingTypes,
     ProjectPriority,
+    UserRole,
 )
 from backend.models.postgis.campaign import Campaign
 from backend.models.postgis.organisation import Organisation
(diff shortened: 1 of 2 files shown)


backend/services/project_search_service.py
```diff
@@ -83,10 +84,14 @@ def create_search_query():
                 Organisation.name.label("organisation_name"),
                 Organisation.logo.label("organisation_logo"),
             )
-            .filter(Project.private.is_(False))
             .outerjoin(Organisation, Organisation.id == Project.organisation_id)
             .group_by(Organisation.id, Project.id)
         )
+
+        # Remove private projects if user role is not admin.
+        if user is None or user.role != UserRole.ADMIN.value:
```
willemarcel, line 92: We need to consider here also org manager, project team managers and project authors. Probably using the `is_user_action_permitted_on_project` method of `ProjectAdminService`

JorgeMartinezG: visibility at that level are planned to be handled in this issue: https://github.com/hotosm/tasking-manager/issues/2817. BTW, AFAIK we agreed to set visibility and permissions to teams level only

PR #2947, opened 2020-05-13
Array of polygons support in project creation

### LINKED ISSUE #2919: Area of interest with multiple polygons error
In project creation view, when aoi has multiple polygons...the taskgrid is generated using the first polygon only.

JorgeMartinezG:
closes #2919  Also included fix when setting arbitrary tasks using uploaded file.

### DIFF
--- a/frontend/src/components/projectCreate/index.js
+++ b/frontend/src/components/projectCreate/index.js
@@ -56,7 +56,7 @@ export const addLayer = (layerName, data, map) => {
 const AlertMessage = ({ err }) => {
   if (err.error === true) {
     return (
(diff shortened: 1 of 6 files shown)


willemarcel:
It worked well.

Some points we could improve:
- It's allowing me to draw more than one polygon, but it's considering only one of them when creating the grid.
- When I upload a bad formatted GeoJSON file, the app crashes.

Low priority (we can create an issue if you think we can handle this better on the future):

- It allows me to upload a GeoJSON with a MultiPolygon, but create the grid with only one of the Polygons.

PR #2974, opened 2020-05-19
v4.0.6 release

willemarcel:


dakotabenjamin:
I get some network errors trying to load iD but perhaps that is from the staging instance limitations. I only epxerienced that for a few minutes and had no issues otherwise.

PR #2985, opened 2020-05-20
Restore members when user cancel editing

### LINKED ISSUE #2977: Deleting Members without Saving
Hello,

If I delete someone (maybe by accident) by clicking the "x" on their icon in edit mode of an organizatoin or team - they are gone, immediately. Although there is the Save/Cancel option.  We should probably allow the 'Cancel' to 'undo' the deletion :)

willemarcel:
Resolves #2977

### DIFF
--- a/frontend/src/components/teamsAndOrgs/members.js
+++ b/frontend/src/components/teamsAndOrgs/members.js
@@ -1,4 +1,4 @@
-import React, { useState, useCallback } from 'react';
+import React, { useState, useEffect, useCallback } from 'react';
 import { Link } from '@reach/router';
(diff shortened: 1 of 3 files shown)


willemarcel:
> Everything is working fine for me.
> 
> I just want to confirm if this console log should be present? If so, it is ready to go.
> 
> https://github.com/hotosm/tasking-manager/blob/1a3b0c0bbce49a17d622cfaf706187cce4b3281e/frontend/src/components/teamsAndOrgs/members.js#L29

thanks, @joaovitor3 ! I removed it

Issue #2691, opened 2020-04-09
Explicit URLs for tasks


xamanu:
User feedback:

> On the map on the project page when i click on a task the URL does not reflect the clicked task. Eg when i enter the url https://tm4.hotosm.org/projects/31/tasks, i click a task in the map, the url stays the same. And also (as a consequence?) i cannot select a specific task via the url, as i could in TM3. I do not see any search function to compensate for this

ritvikjain99:
@willemarcel  We can give a thought to the inclusion of hash navigation here ! 
like /project/{id}/tasks#{task-number}

willemarcel:
@ritvikjain99 It's bit more complicated. Right now we have the `?search=` param  than can do it. I tried to add a `?task=` to enable just the selection of the task, but it was affecting other components. I'll try to work on it again after solving some problems on the task list.

tawirth:
I would like to vote this one up. For the Crowd2Map project we used to ask for a url to the project task square so we could validate or provide new mapper feedback. Now we are asking that they provide the task number but we still have to go and search for the osm username to find the task square. From my point of view this is a step backward in the new tasking manager. Thanks.

PR #2989, opened 2020-05-21
allow users to select multiple tasks for validation

### LINKED ISSUE #2580: Select multiple tasks for validation
Feedback from the testing with validators about the need to validate with more than one task (this was possible in TM3):

* Option to select multiple adjoining tasks to download (using a polygon) in JOSM/iD.
* Eventually as clicks or with the  option to draw and select a square/rectangular over tasks to be validated.

willemarcel:
Resolves #2580


willemarcel:
The second commit resolves #2691

PR #2995, opened 2020-05-22
Id presets feature

### LINKED ISSUE #2302: Add iD preset selection to project edit page
Allow the users to select which presets iD will load when mapping a TM project task.

JorgeMartinezG:
closes #2302


willemarcel:
@JorgeMartinezG I removed some of the presets that are not useful for TM. Also fixed a bug when the `project.idPresets` is empty and added the test to `getIdUrl` function.

Although the iD url param is correct, it didn't work to me yet. Maybe we need to do some work on the integrated iD.

I don't feel the react-select component is a good choice, I think we need a way to add, for example, all the buildings presets in one click, otherwise it will be a bit painful for the managers to select all presets the project needs.

JorgeMartinezG:
@willemarcel I decided to use react select because of the search option. I was checking db dump and could not find projects with id presets. So the number of presets on a single project might not be large. 

One option to reduce the number of steps is to enable the select option not to be closed on a single select. The library has this option **CloseOnSelect** and set it to False

willemarcel:
@JorgeMartinezG yes, we can maintain the react-select and wait for more feedback. I noticed the project summary endpoint was returning the idPreset always as `[]`, so I fixed it. However we can not merge this yet, because the integrated iD is not listing all the presets options we pass.

willemarcel:
@JorgeMartinezG as some organisations were in need to that feature, I merged it and opened a follow up issue: https://github.com/hotosm/tasking-manager/issues/3043 . It's something that needs to be changed in iD side.

PR #3001, opened 2020-05-23
use circleci/python:3.7-buster-node image and replace npm by yarn

willemarcel:
- It's a good practice to commit the `yarn.lock` file to assure the correct frontend package versions are instaled
- The `circleci/python:3.7-buster-browsers` is needed only when we need to run selenium tests, which needs a browser. On our case, I believe it's better to user the -node one, which already includes node and yarn packages


willemarcel:
I removed `- frontend/node_modules` from the `cached-dependencies` to avoid problems with previous dependencies installed by `npm`, but we can add it back in a next PR.

eternaltyro:
> I removed `- frontend/node_modules` from the `cached-dependencies` to avoid problems with previous dependencies installed by `npm`, but we can add it back in a next PR.

@willemarcel , this means that between one stage and a different stage of the CI process, the files will not be carried over. It will be as if `yarn` was not run at all, I think. @dakotabenjamin might have a better picture of the behaviour involved.

.circleci/config.yml
```diff
@@ -24,16 +23,15 @@ jobs:
         keys:
         - cached-dependencies
         paths:
-        - frontend/node_modules
```
eternaltyro, line 27: Reconsider this.

willemarcel: I'm not sure about it too. We can do a test without removing it.

willemarcel:
@eternaltyro I reverted the change on the `cached-dependencies` and replaced npm by yarn also on the installation of mapbox/cfn-config and cloudfriend

eternaltyro:
> @eternaltyro I reverted the change on the `cached-dependencies` and replaced npm by yarn also on the installation of mapbox/cfn-config and cloudfriend

@willemarcel thanks.

.circleci/config.yml
```diff
@@ -3,17 +3,16 @@ jobs:
   build:
     working_directory: /home/circleci/app
     docker:
-    - image: circleci/python:3.7-buster-browsers
```
dakotabenjamin, line 6: if we remove `-browsers` will that affect the frontend tests? Or is that overwritten below in line 55?

.circleci/config.yml
```diff
@@ -69,7 +68,7 @@ jobs:
     - store_artifacts:
         path: tests/backend/results
     - save_cache:
-        key: cached-dependencies
+        key: node-v1-{{ .Branch }}-{{ checksum "package-lock.json" }}
```
willemarcel, line 71: replace that line by `key: yarn-packages-v1-{{ .Branch }}-{{ checksum "yarn.lock" }}`

eternaltyro:
@dakotabenjamin looks good to merge to me (LGTMTM)

PR #3047, opened 2020-05-28
Search projects without tables union for correct sorting

### LINKED ISSUE #2869: Fix sorting of the manage projects page
The sorting of the projects in the manage projects page is not having an effect, when choosing the different options.

image

JorgeMartinezG:
closes #2869 .

Search for projects based on user organization moved to global search and removed union query. This fixes the bug related to sorting.

### DIFF
--- a/backend/services/project_search_service.py
+++ b/backend/services/project_search_service.py
@@ -35,7 +35,6 @@ from backend.models.postgis.utils import (
 )
 from backend.models.postgis.interests import project_interests
 from backend.services.users.user_service import UserService
-from backend.services.organisation_service import OrganisationService
 
 
 search_cache = TTLCache(maxsize=128, ttl=300)
@@ -91,9 +90,15 @@ class ProjectSearchService:
         if user is None:
             query = query.filter(Project.private.is_(False))
 
-        # Get also private projects of teams that the user is member.
         if user is not None and user.role != UserRole.ADMIN.value:
+            # Get also private projects of teams that the user is member.
             project_ids = [[p.project_id for p in t.team.projects] for t in user.teams]
+
+            # Get projects that belong to user organizations.
+            orgs_projects_ids = [[p.id for p in u.projects] for u in user.organisations]
+
+            project_ids.extend(orgs_projects_ids)
+
             project_ids = tuple(
                 set([item for sublist in project_ids for item in sublist])
             )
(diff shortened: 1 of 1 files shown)


JorgeMartinezG:
@willemarcel did you remove the cache? I will be checking also

willemarcel:
@JorgeMartinezG Yes, I tested again and assured it was not using cache

JorgeMartinezG:
@ramyaragupathy @willemarcel branch updated and rebased with latest develop

ramyaragupathy:
I can confirm that I don't see archived projects under `My contributions` -> `My projects`. Also, can we directly use the `projects_mapped` field from user model to identify projects that the user has contributed to?

ramyaragupathy:
Sorry about an accidental commit made to this branch

backend/services/project_search_service.py
```diff
@@ -292,16 +297,13 @@ def _filter_projects(search_dto: ProjectSearchDTO, user):
...
-                ProjectTeams.role == TeamRoles.PROJECT_MANAGER.value,
-                ProjectTeams.project_id == Project.id,
-            )
-            user_orgs_list = OrganisationService.get_organisations_managed_by_user(
-                search_dto.managed_by
+            # Get all the projects associated with the user and team.
+            user_teams_ids = [u.team_id for u in user.teams]
+            query = (
+                query.filter(ProjectTeams.project_id == Project.id)
+                .filter(ProjectTeams.role == TeamRoles.PROJECT_MANAGER.value)
```
ramyaragupathy, line 304: This filter overrides the org filter in the `create_search_query` filter. So within `create_search_query` all the proejcts related to user's teams and orgs are collected in one query. Finally within this `managed_by` block when you say filter only those teams that have PM role, all the org related projects are eliminated resulting in null array

backend/services/project_search_service.py
```diff
@@ -292,16 +297,13 @@ def _filter_projects(search_dto: ProjectSearchDTO, user):
...
+                .filter(ProjectTeams.team_id.in_(user_teams_ids))
```
ramyaragupathy, line 305: This filter nullifies the org filter in the `create_search_query` method. So within `create_search_query` all the projects related to user's teams and orgs are collected in one query. Finally within this `managed_by` block when you say filter only those user teams that have PM role, all the org related projects are eliminated resulting in null values

JorgeMartinezG:
@ramyaragupathy I have made changes in the project search service file. The validation for teams is made based on the team membership function of the user, rather than the team role within a project. Also, included the organizations check within the **managed_by_me** condition

JorgeMartinezG:
@ramyaragupathy. Made the change. The validation is done based on the team role within each project rather than team membership

PR #3085, opened 2020-06-04
separate frontend and backend deploy jobs

willemarcel:


willemarcel:
@dakotabenjamin exactly, so if we want to update only the frontend we push only to that branch. We can also add the `deployment/hot-tasking-manager` to the `frontend-production` workflow, so we don't need to push to both when we want to update backend and frontend.

dakotabenjamin:
@willemarcel added that, also made it so the deployment doesn't run `build-deploy` unnecessarily

Issue #2938, opened 2020-05-12
Allow team managers to message all team members


willemarcel:
Add the possibility to message all team members on the manage team page.

PR #3104, opened 2020-06-08
Message all team members 

ramyaragupathy:
From #2938 - just the backend to send message to all team members


backend/api/teams/actions.py
```diff
@@ -234,3 +236,75 @@ def post(self, team_id):
             error_msg = f"TeamMembers DELETE - unhandled error: {str(e)}"
             current_app.logger.critical(error_msg)
             return {"Error": error_msg}, 500
+
+
+class TeamsActionsMessageMembersAPI(Resource):
+    @tm.pm_only(True)
```
willemarcel, line 242: is this decorator useful yet?

backend/api/teams/actions.py
```diff
@@ -234,3 +236,75 @@ def post(self, team_id):
...
+            403:
+                description: Forbidden
+            500:
+                description: Internal Server Error
+        """
+        try:
+            authenticated_user_id = token_auth.current_user()
+            message_dto = MessageDTO(request.get_json())
+            message_dto.from_user_id = authenticated_user_id
+            message_dto.validate()
```
willemarcel, line 293: We need to restrict only to team managers and the managers of the organization of team.

backend/services/team_service.py
```diff
@@ -539,3 +541,26 @@ def check_team_membership(project_id: int, allowed_roles: list, user_id: int):
...
+    def send_message_to_all_team_members(team_id: int, message_dto: MessageDTO):
+        """  Sends supplied message to all contributors in a team.  Message all team members can take
+             over a minute to run, so this method is expected to be called on its own thread """
+        print(team_id, message_dto)
+        app = (
+            create_app()
+        )  # Because message-all run on background thread it needs it's own app context
+
+        with app.app_context():
+            team_members = TeamService._get_team_members(team_id)
```
willemarcel, line 555: we could restrict only to active members, what do you think?

backend/services/team_service.py
```diff
@@ -539,3 +541,26 @@ def check_team_membership(project_id: int, allowed_roles: list, user_id: int):
...
+
+            messages = []
+            for team_member in team_members:
+                print(team_member)
+                message = Message.from_dto(team_member.user_id, message_dto)
+                message.message_type = MessageType.BROADCAST.value
```
willemarcel, line 561: Instead of using "broadcast" we could create a "Team communication" type

ramyaragupathy: Added a new message type

willemarcel:
@ramyaragupathy I rebased it with develop and made some changes:
- add context to message informing who sent the message and which team this message is about.
- remove the sender from the list of recipients
- parse markdown on the message
- added a string to the TEAM_BROADCAST notification type on the frontend

I noticed the backend is accepting an empty string on the message subject and it breaks the frontend style.

willemarcel:
@ramyaragupathy I added the message team members frontend component to `/manage/teams/:id/`

ramyaragupathy:
> I noticed the backend is accepting an empty string on the message subject and it breaks the frontend style.

Added a required check to schema validation

backend/api/teams/actions.py
```diff
@@ -289,13 +289,21 @@ def post(self, team_id):
             authenticated_user_id = token_auth.current_user()
             team_id = request.view_args["team_id"]
             message_dto = MessageDTO(request.get_json())
+            # Validate if team is present
+            try:
+                TeamService.get_team_by_id(team_id)
+            except NotFound:
+                return {"Error": "Team not found"}, 400
```
willemarcel, line 296: code should be 404

backend/services/team_service.py
```diff
@@ -561,6 +561,7 @@ def send_message_to_all_team_members(
                     message.message_type = MessageType.TEAM_BROADCAST.value
                     message.save()
                     user = UserService.get_user_by_id(team_member.user_id)
-                    messages.append(dict(message=message, user=user))
+                    if user.teams_notifications:
```
willemarcel, line 564: @ramyaragupathy I saw your last commit. Should we remove that check?

ramyaragupathy: Test locally with a team and flags on and off for users. Similar to mentions, DB messages are always sent, only emails updates are customizable by users.

PR #3136, opened 2020-06-12
Validate repeated task in project creation

JorgeMartinezG:


willemarcel:
@JorgeMartinezG closing it, as it seems we don't need that check

PR #3158, opened 2020-06-16
Cleanup and simplify code

eternaltyro:
- Remove unnecessary comments from appspec.yml
- Add example values for cfn-config parameter prompts
- Add resource tags to target group
- Get target group to use modern TLS ciphers with Forward Secrecy

### DIFF
--- a/scripts/aws/cloudformation/tasking-manager.template.js
+++ b/scripts/aws/cloudformation/tasking-manager.template.js
@@ -20,7 +20,8 @@ const Parameters = {
   },
   DatabaseDump: {
     Type: 'String',
-    Description: 'Path to database dump on S3'
+    Description: 'Path to database dump on S3',
+    Default: 's3://example-bucket/tasking-manager.sql'
   },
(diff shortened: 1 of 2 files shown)


scripts/aws/cloudformation/tasking-manager.template.js
```diff
@@ -20,7 +20,8 @@ const Parameters = {
   },
   DatabaseDump: {
     Type: 'String',
-    Description: 'Path to database dump on S3'
+    Description: 'Path to database dump on S3',
+    Default: 's3://example-bucket/tasking-manager.sql'
```
dakotabenjamin, line 24: If we add this then every cloudformation update will attempt to restore a backup from "example-bucket/tasking-manager.sql". Better to leave it blank, and instead add the examples into the descriptions.

eternaltyro: Makes sense. Let me change it.

PR #3164, opened 2020-06-17
Fix Tags syntax for ALB TargetGroup

eternaltyro:


eternaltyro:
This is a minor change that fixes CloudFormation template syntax. So I'm going to merge it without review.

PR #3169, opened 2020-06-17
Fix username popup in textarea component

### LINKED ISSUE #3025: name popup obscures typing of name
image

In the Questions and Comments edit box, you can start typing a name `@xyz` and a popup filtering the names that match the currently typed letters is displayed.  That popup covers the letters that are being typed.

JorgeMartinezG:
closes #3025 
ItUf58bZNW


willemarcel:
It worked on the map / validation page, but not on the question and comments section of the project detail page.

JorgeMartinezG:
working fine @willemarcel

wireguy:
please test one more case with the new solution.  move the question and answer box to bottom of the browser window.  then type the users name.   the popup should still be visible (above the text you are typing?) and not disappear below the browser windows frame.

willemarcel:
@wireguy It's working that way already

wireguy:
image

Here's the test of that with the fix...

willemarcel:
@wireguy The component is not made by us, and it uses as reference the page size. On that case, you just need to scroll the page, so it's not a blocker. On the map/validation page where the form is very close to the bottom of the page, it loads to the top if needed.

PR #3185, opened 2020-06-22
make permission evaluation only after team request is finished

willemarcel:


ramyaragupathy:
Per chat:
* Logged in as non-admin
* member of validator trainees group

I was able to lock tasks for validation and there were no error prompts in between which I have noticed earlier. This is good to go 👍

Issue #3175, opened 2020-06-19
Task stats in User DTO 


ramyaragupathy:
`/users/queries/username` endpoint is missing details on task stats - mapped/validated/invalidated

willemarcel:
@ramyaragupathy those stats are available on `/api/v2/users/<username>/statistics/` and I prefer to concentrate the stats on that endpoint. We already get too much data on the login.

ramyaragupathy:
Agree with you @willemarcel - as long is this data point is not used on the frontend for any other page, I am ok to remove it

PR #3192, opened 2020-06-23
Remove stats from user dto

ramyaragupathy:
Per convo in https://github.com/hotosm/tasking-manager/issues/3175 - removed stats field from user DTO


willemarcel:
@ramyaragupathy the only problem I see on merging that PR is that we need a way to find if the user has already mapped a task. The other endpoints that exposes that information would trigger a heavy query on the database and cost time on the login.

I propose to add on the user DTO that field that we have on the UserStatsDTO: https://github.com/hotosm/tasking-manager/blob/fix/username-query/backend/services/users/user_service.py#L376

ramyaragupathy:
Per chat, we're sticking with `projectsMapped` field. Updated DTO and relevant frontend

Issue #2983, opened 2020-05-20
Missing user on project contributions endpoint


willemarcel:
![image](https://user-images.githubusercontent.com/666291/82474698-d27c6000-9aa1-11ea-811b-78276bfc1e1a.png)

@wireguy worked on this project, but the `/api/v2/projects/8470/contributions/` endpoint doesn't return his name

wireguy:
not just me, also  DEO NISHIMWE and Awania Moris,  if that helps with debugging.

PR #3201, opened 2020-06-24
Include all contributors for project_id/contributions

ramyaragupathy:
From https://github.com/hotosm/tasking-manager/issues/2983:
* replace left outer join with full outer join
* include both `mapped_by` and `validated_by` filter for user table join


JorgeMartinezG:
Hi @ramyaragupathy. I made a slight change to return always an array in the DTO, even when there are 0 taskIds found for a given user. I also included the frontend changes.

PR #3238, opened 2020-07-01
Handle member omission for private teams

ramyaragupathy:
Per chat w/ project managers for 8999 - we figured out that with the new query parameter `omitMemberList`, private team details are not appended for a user since permissions check is overridden because of a conditional placement check.

As a result if a user is a member of a private team and if `omitMemberList=True` then team data is not appended to the final list, thereby interrupting the user from contributing to the project. This fix updates the logic in team_service.

### DIFF
--- a/backend/services/team_service.py
+++ b/backend/services/team_service.py
@@ -239,6 +239,8 @@ class TeamService:
 
         for team in query.all():
             team_dto = TeamDTO()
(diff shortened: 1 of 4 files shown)


willemarcel:
@ramyaragupathy coud you include the omitMemberList query param we removed on #3241?

PR #3274, opened 2020-07-07
Average time and estimated time to finish components

### LINKED ISSUE #3107: [project-stats] Create average time and completion prediction components
On the project stats page, a project manager wants to see
- [ ] the average time users spend to map a task
- [ ] the average time users spend to validate a task
- [ ] a prediction of when a project would be totally mapped
- [ ] a prediction of when a project would be totally validated

About the last two items, I think we can include two information:
- how many hours of work are needed
- keeping the same work pace, when it would be totally mapped/validated

JorgeMartinezG:
closes #3107


backend/models/postgis/project.py
```diff
@@ -723,6 +723,14 @@ def get_project_stats(self) -> ProjectStatsDTO:
                     )
                     project_stats.average_validation_time = average_validation_time
 
+        time_to_finish_mapping = (
```
ramyaragupathy, line 726: @JorgeMartinezG - why do use this over `total_mapping_seconds`?

backend/models/postgis/project.py
```diff
@@ -723,6 +723,14 @@ def get_project_stats(self) -> ProjectStatsDTO:
...
+            self.total_tasks - (self.tasks_mapped + self.tasks_bad_imagery)
+        ) * project_stats.average_mapping_time
+        project_stats.time_to_finish_mapping = time_to_finish_mapping
+        project_stats.time_to_finish_validating = (
+            self.total_tasks - self.tasks_validated
```
ramyaragupathy, line 731: Similar to mapping time, should we use total_validation_seconds here?

backend/models/postgis/project.py
ramyaragupathy: Understand this is for time estimates on remaining tasks - in that case, how do we handle projects with 0 tasks mapped/validated

willemarcel:
@ramyaragupathy I added the tasks status section here

PR #3276, opened 2020-07-08
Validate not null geometry for project search and creation

### LINKED ISSUE #3269: Projects with null geometry
There are cases where we have projects with null centroid/geometry created and subsequently, no tasks are identified with the projects too. Such projects interfere with search results resulting in 500 error

JorgeMartinezG:
Closes #3269


JorgeMartinezG:
@ramyaragupathy I included the restriction to the centroid column in projects models and migration

Issue #951, opened 2017-11-13
Improve tests and coverage


ethan-nelson:

xamanu:
Hijacking this issue here. During the last week's tech team workshop we identified tests and test coverage to be one of our priorities to assure code and service quality around the Tasking Manager. With @ramyaragupathy I started a conversation about bringing up relevant test scenarios that urge most to us and get them implemented. Let's use this issue to progress on this side.

xamanu:
First two cents from my side:

* Check whether a user who hasn't been logged in at all can do.
* Check whether the blocks of a blocked user are working (like selecting a task for mapping and validation or comment on any task or project).
* Test deployments before actual deployments

JorgeMartinezG:
Code coverage was included in https://github.com/hotosm/tasking-manager/pull/799. Users can run tests and verify coverage in the `server` folder. However, during deployment and local tesing, there is no minimum coverage percentage value for Pull request to pass

PR #3280, opened 2020-07-08
Sort task history applying most recent first

### LINKED ISSUE #3044: TaskHistory order on map/validation page
Revert ordering of history items on map/validation page, in order to make the most recent items be on the top:
image

JorgeMartinezG:
closes #3044

### DIFF
--- a/backend/models/postgis/task.py
+++ b/backend/models/postgis/task.py
@@ -508,7 +508,9 @@ class Task(db.Model):
     )
 
     # Mapped objects
-    task_history = db.relationship(TaskHistory, cascade="all")
+    task_history = db.relationship(
+        TaskHistory, cascade="all", order_by=desc(TaskHistory.id)
+    )
     task_annotations = db.relationship(TaskAnnotation, cascade="all")
     lock_holder = db.relationship(User, foreign_keys=[locked_by])
     mapper = db.relationship(User, foreign_keys=[mapped_by])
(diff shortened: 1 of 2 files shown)


backend/models/postgis/task.py
```diff
@@ -1013,7 +1013,7 @@ def as_dto(
     def as_dto_with_instructions(self, preferred_locale: str = "en") -> TaskDTO:
         """Get dto with any task instructions"""
         task_history = []
-        for action in reversed(self.task_history):
+        for action in self.task_history:
```
ramyaragupathy, line 1016: Ran this against a task #951 on project 5660, comparing task history ID values from production and current branch:

Production results | Current branch results
-------------------|-----------------------
7683258<br>7683259<br>7683260<br>7683261<br>7717936<br>7717508<br>10858873<br>10858874<br>10858872<br>11187402|11187402<br>10858872<br>10858874<br>10858873<br>7717508<br>7717936<br>7683261<br>7683260<br>7683259<br>7683258

The current branch simply inverts the array list. If you see the production results you can observe there is already a problem in task history ordering which is carried to the array reversal function too. We should correct this ordering in [Task.get](https://github.com/hotosm/tasking-manager/blob/develop/backend/models/postgis/task.py#L573) - but if this affects other methods, then may be we should go for sort with python instead of reversed method.  I tired `sorted(self.task_history, key=lambda x:x.id, reverse=True)` and this returned me IDs in the following order and it aligned well with the `action_date` timestamps.

11187402<br>10858874<br>10858873<br>10858872<br>7717936<br>7717508<br>7683261<br>7683260<br>7683259<br>7683258

JorgeMartinezG:
@ramyaragupathy Instead of performing the sort in Python, I modified the relationship in order to perform the sort directly in Postgres. Please pull changes again since I rebased the branch

PR #3284, opened 2020-07-10
Improvement in statistics queries

### LINKED ISSUE #3282: optimize api system/statistics/
Currently its response time is 13 seconds:
https://tasking-manager-staging-api.hotosm.org/api/v2/system/statistics/

NoemiNahomy:
closes #3282 
In this case to improve response time it is more convenient to use sql queries.
Before:    13000 ms
After:       329 ms


JorgeMartinezG:
@NoemiNahomy could you please use sqlalchemy to the queries?

PR #3286, opened 2020-07-11
Make user details page components handle case where no data is present

willemarcel:
... and add tests to the components.


willemarcel:
@ramyaragupathy could you review the text messages?

Issue #3324, opened 2020-07-15
Asynchronous messages


NoemiNahomy:
list of methods for sending messages that should be in the background:
- [x]     send_message_after_validation
- [x]     send_welcome_message
- [x]     send_message_after_comment
- [x]     send_message_after_chat
- [x]     send_favorite_project_activities

PR #3325, opened 2020-07-15
asynchronous sending messages

NoemiNahomy:
Related #3324


NoemiNahomy:
it seems that there was a problem with the PR, I will do it again

PR #3364, opened 2020-07-21
improvement in project search

NoemiNahomy:
Improvement in project search
api : /api/v2/projects/?campaign=4+County+Digitization+Liberia

Also to improve pagination response times in project search, the [seq_page_cost](https://www.postgresql.org/docs/9.5/runtime-config-query.html#GUC-SEQ-PAGE-COST) parameter could be set in postgres.

### DIFF
--- a/backend/services/project_search_service.py
+++ b/backend/services/project_search_service.py
@@ -209,14 +209,18 @@ class ProjectSearchService:
     @staticmethod
     def _filter_projects(search_dto: ProjectSearchDTO, user):
         """ Filters all projects based on criteria provided by user"""
+
         query = ProjectSearchService.create_search_query(user)
+
         query = query.join(ProjectInfo).filter(
             ProjectInfo.locale.in_([search_dto.preferred_locale, "en"])
         )
         project_status_array = []
         if search_dto.project_statuses:
-            for project_status in search_dto.project_statuses:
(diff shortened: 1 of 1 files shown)


backend/services/project_search_service.py
```diff
@@ -325,7 +328,13 @@ def _filter_projects(search_dto: ProjectSearchDTO, user):
 
         all_results = []
         if not search_dto.omit_map_results:
-            all_results = query.all()
+            query_result = query
+            query_result.column_descriptions.clear()
```
ramyaragupathy, line 332: Can you please explain about these steps a bit?

ramyaragupathy:
@NoemiNahomy - Any changes that we have on the DB, let's do it via a migration.

backend/services/project_search_service.py
NoemiNahomy: in this line when we make this query `query.all()`, this query uses all the project attributes and the filter and we don't need all those attributes, it only needs 3. So I use the filter but I only do the query with 3 values.       
These values ​​are used for the geojson of the map
 https://github.com/hotosm/tasking-manager/blob/2e94a550a28b6d9bff2be8953d69ea693fa04b25/backend/services/project_search_service.py#L193

ramyaragupathy:
DB setting will be handled in a separate  PR migration

PR #3370, opened 2020-07-22
Message service to use notification settings

ramyaragupathy:
Update message service functions to use exiting user settings for notification.


ramyaragupathy:
Closing in favor of #3371

Issue #3356, opened 2020-07-20
Modify mention notifications behaviour


willemarcel:
Modify the "mention notification" settings on the user profile to enable/disable sending an email to the user, but create the notification to the user even if it's disabled.

PR #3371, opened 2020-07-22
Validate each user settings before sending message

JorgeMartinezG:


backend/services/messaging/message_service.py
```diff
@@ -124,20 +124,32 @@ def send_message_to_all_contributors(project_id: int, message_dto: MessageDTO):
...
             user = message.get("user")
+            obj = message.get("message")
+            if (
+                user.mentions_notifications is False
+                and obj.message_type == MessageType.MENTION_NOTIFICATION.value
+            ):
+                continue
+            if (
+                user.projects_notifications is False
+                and obj.message_type == MessageType.PROJECT_ACTIVITY_NOTIFICATION.value
```
willemarcel, line 145: The intention of https://github.com/hotosm/tasking-manager/issues/3356 was to modify the behaviour only to Mention notifications. This change extends to projects and then to task comments and project chat. We could discuss that possibility, but I think it can also generate a lot of noise to mappers.

ramyaragupathy: new changes looks good

PR #3396, opened 2020-07-28
handle plural case on "Validate selected tasks" button

### LINKED ISSUE #3392: Change 'Validate selected task' button text to show how many tasks are selected
image

It looks like there are 3 tasks that I have selected to validate.   The button just says 'Validate selected task'.
There are actually 4 tasks selected.  One is off screen, and I accidentally clicked on it.
It would be nice if the button text said 'Validated 4 selected tasks'.
Then I have a chance to notice my error.

willemarcel:
Resolves #3392

image

### DIFF
--- a/frontend/src/components/taskSelection/footer.js
+++ b/frontend/src/components/taskSelection/footer.js
@@ -214,7 +214,10 @@ const TaskSelectionFooter = (props) => {
             ) ? (
               <FormattedMessage {...messages.selectAnotherProject} />
             ) : (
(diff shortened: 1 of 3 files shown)


willemarcel:
@ramyaragupathy I didn't merge it yet, because I'm not sure about the convenience of the new message format for translators. I'm afraid it can cause errors of translation

ramyaragupathy:
Singular and plural tone in message works good

Issue #3357, opened 2020-07-20
Change "Project updates" settings text


willemarcel:
![image](https://user-images.githubusercontent.com/666291/87967473-3d8fe600-ca95-11ea-8b19-ad605d4c69ec.png)

Rewrite to "Receive a notification every time a task you mapped is marked as Finished or More mapping needed.", as these are the only actions that are triggering that type of notifications.

russdeffner:
Would also say this isn't "Project updates" but "Task Updates" or maybe "Task Status (Updates)"

ramyaragupathy:
Capturing the current behaviour on Project updates - this is for the overall progress of projects that a user has contributed to or favorited. This is different from Task-related notifications. I would suggest adding it as a different category - as Task updates notifications.

willemarcel:
![image](https://user-images.githubusercontent.com/666291/89298252-6ee2e700-d63b-11ea-9d57-2d7b851c2f63.png)
We added the tasks type and let the project updates type as it was previously

PR #3398, opened 2020-07-28
Rewrite notifications and improve notification types settings

ramyaragupathy:
From #3357 - regrouping notification types to project and task notifications


willemarcel:
@ramyaragupathy A summary of changes I've made
- Include message content on emails
- Rewrite some messages and subjects
- Add function to replace variables on the message templates
- Instead of a generic "You have a new message on Tasking Manager" use the message subject on the email
- Use ORG_CODE and ORG_NAME instead of hardcoded "HOT"

ramyaragupathy:
Tested all your changes @willemarcel with the following use cases:
* mention with settings true and false
* task validation/invalidation with setting true and false

We were missing to cover DB message save when task notification is false. Fixed that case. Good to merge

PR #3407, opened 2020-07-30
Add BaseTestCase and test environment

willemarcel:


willemarcel:
There are 32 tests that needs work.

Steps needed to finish the work:

- Use `BaseTestCase` in all tests that needs database
- Delete all skipTest, setUp and tearDown function from existing tests
- If you need to run something on the `setUp`, use `super().setUp()` and add your code
- The `create_canned_user` and `create_canned_project` are good to use
- execute the tests always with` python -m unittest discover tests/backend`

PR #3479, opened 2020-08-14
Filter and fetch team related

ramyaragupathy:
Per chat w/ @willemarcel - we've got a few issues on team filters and on further investigation fixed the table join type and filter conditions.


willemarcel:
@ramyaragupathy it would be good to make the team search by name case insensitive

PR #3516, opened 2020-08-21
Remove frontend views from backend server

willemarcel:
As it causes a lot of errors on the logs and also is confusing to maintain those views on the project without using it effectively.
Added a redirect to the heartbeat, instead.


ramyaragupathy (approved):
Dev setup works as expected

Issue #2566, opened 2020-03-24
Slim top bar while editing in ID


xamanu:
Might be nice to slim down the top bar while in iD mode to increase the amount of mapping area.

JorgeMartinezG:
@willemarcel we could make a chevron-up down option to the header and make a state of it, what do you think?

willemarcel:
@JorgeMartinezG we have a chevron icon at the side of the username, I don't have a good idea of icon that could fit that need... Initially I thought about hiding the top section of the header (where we have the hotosm.org link) on the map/validation page, but we would need to add some animation to avoid a page flicking.

ritvikjain99:
@willemarcel I can add transition which would reduce its height to zero when id mapping tab is opened, if I am allowed to !

Issue #3538, opened 2020-08-25
Option to open iD editor in it's own tab?


bmckinlay:
The previous version of HOT tasking manager would open the iD editor in a separate tab. This would allow me to run a full-screen iD editor session, switching back to the tasking manager in a separate tab to manage the task.

The new version of tasking manager opens an embedded iD editor. I can see how this is more user-friendly for some, but on my small 13" laptop screen, it leaves far less screen real-estate for iD - a poor user experience that makes editing much more difficult.

Is there an option to open iD editor separately (ie: not embedded in the tasking manager interface)? Could this option be added?

Thanks

willemarcel:
@bmckinlay you can press F11 to use iD on full screen mode

bmckinlay:
Huh? Pressing F11 does nothing for me. The tasking manager interface is still taking up a lot of screen space.

I guess a better way to phrase this is: How do I open a stand-alone instance of the iD editor, outside of the tasking manager?

bmckinlay:
@willemarcel OK, I figured it out - it's ctrl-command-F on mac.

But weirdly, only the actual keyboard command works, if you select "View/Enter full screen" from the menu it just gives you a full screen window with the task manager interface still wrapped around it.

It would be good if there as a more obvious way to do this!

willemarcel:
@bmckinlay What we will implement soon is to hide part of the header when iD is loaded, so it will make the screen space quite similar to what you have on osm.org.

It's not on the plan to support opening iD in another window, because we use a different API to load it integrated, and it would cost more to maintain both ways.

PR #3543, opened 2020-08-26
Hide organisation top bar when iD editor is running

willemarcel:


willemarcel:
Resolves #2566. Related to #3538

ramyaragupathy (approved):

Issue #1100, opened 2018-05-01
Pause project page updates if window/tab is not the focus


bgirardot:
If we are not doing this already, if a tab or page is not the focused page/tab in the browser, we should pause updates.

as soon as it returns to focus, we would then do an update and continue as usual.

ethan-nelson:
This relates to pausing javascript polling on the `/project/#` API endpoint when the tab loses focus.

PR #3545, opened 2020-08-26
disable network requests if the app is not visible to user

willemarcel:
Resolves https://github.com/hotosm/tasking-manager/issues/1100


ramyaragupathy (approved):
Tested on task selection page - no network requests were sent while away from window 👍 . This is useful for the backend, thank you

PR #3558, opened 2020-09-02
Handle country serialisation with inbuilt functions

ramyaragupathy:
Per discussion updated the country serialisation approach

### DIFF
--- a/backend/models/postgis/project.py
+++ b/backend/models/postgis/project.py
@@ -272,13 +272,10 @@ class Project(db.Model):
             lat, lng
         )
         country_info = requests.get(url).json()  # returns a dict
-        country_info_json = json.dumps(country_info)  # dict -> json
-        # Load the JSON to a Python list & dump it back out as formatted JSON
-        data = json.loads(country_info_json)
-        if data["address"].get("country") is not None:
-            self.country = [data["address"]["country"]]
+        if country_info["address"].get("country") is not None:
+            self.country = [country_info["address"]["country"]]
         else:
-            self.country = [data["address"]["county"]]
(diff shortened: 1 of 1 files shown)


backend/models/postgis/project.py
```diff
@@ -271,8 +271,8 @@ def set_country_info(self):
         url = "https://nominatim.openstreetmap.org/reverse?format=jsonv2&lat={0}&lon={1}".format(
             lat, lng
         )
-        country_info = requests.get(url)
-        country_info_json = country_info.content.decode("utf8").replace("'", '"')
+        country_info = requests.get(url).json()  # returns a dict
+        country_info_json = json.dumps(country_info)  # dict -> json
         # Load the JSON to a Python list & dump it back out as formatted JSON
         data = json.loads(country_info_json)
```
willemarcel, line 277: you can remove the `json.dumps` and the `json.loads`. `country_info["address"].get("country")` works.

JorgeMartinezG: Agree. Basically the change should be
```data = requests.get(url).json()```

and remove line 277

willemarcel (approved):
@ramyaragupathy @JorgeMartinezG I added a commit with a better error handling.

It won't fail if nominatim is offline or if the request doesn't return country information (geocoding failure).

willemarcel:
I made another PR with
- an improvement on the query
- if the project has not country information, fetch nominatim again when the project is updated.

ramyaragupathy:
Tested PR with a polygon that could not be geocoded to a country location and this works good.

willemarcel (approved):

PR #3060, opened 2020-05-30
WIP! - custom OSM server config for docker-compose

majdal:
**Note** This is work in progress - just opened the PR to make sure that I'm starting on the right foot. I'm not sure if this is good practice or not, so feel free to close and I'll re-open when I'm further along. 

Meant to address #2865 

A question though: Shall I also change the names in the user messages? e.g. the welcome message at: `backend/services/messaging/templates/welcome_message_en.txt`

PR #3564, opened 2020-09-04
Custom OSM server and editor configuration, based on @majdal's work

geohacker:
This PR is based on @majdal's work in https://github.com/hotosm/tasking-manager/pull/3060. I pulled his commit into a fork and addressed the comments from @willemarcel during his review. We can safely close #3060. I happened to be working on this right now, and can address any feedback.


geohacker:
This PR also changes Dockerfiles for backend and frontend to use code locally instead of pulling from Github, especially when using docker-compose.override.yml. I found that this is now the best way to have a local setup with docker and test changes. But this could probably be not the case for the use case here @willemarcel @JorgeMartinezG. I can take those changes out if you think it doesn't make sense.

willemarcel (approved):

willemarcel:
Thanks, @geohacker & @majdal !

PR #3573, opened 2020-09-08
v4.1.8 release

willemarcel:


dakotabenjamin (approved):
The new Envvars are not set in the cloudformation template so defaults will be used.

willemarcel:
@dakotabenjamin we don't need to set it. The default values in code makes TM work the same way it's now

JorgeMartinezG (approved):

PR #3598, opened 2020-09-15
add notification status filter and use that query on the notification bell

willemarcel:
Ordering notifications by `read` status to load it on the notification bell caused the date ordering to be inconsistent:

Screenshot from 2020-09-15 17-43-43

So I added a status filter on the endpoint and updated the frontend to use that query param, so now the notification bell is more consistent.

image

### DIFF
--- a/backend/api/notifications/resources.py
+++ b/backend/api/notifications/resources.py
@@ -124,7 +124,6 @@ class NotificationsAllAPI(Resource):
               name: messageType
               type: string
               description: Optional message-type filter; leave blank to retrieve all
-              default: 1,2
             - in: query
               name: from
               description: Optional from username filter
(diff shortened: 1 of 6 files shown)


willemarcel:
By purpose, the screenshots have different colours because the first one was made on Chrome and it has a [10 years old bug](https://bugs.chromium.org/p/chromium/issues/detail?id=44872) that render some colours wrongly

willemarcel:
The last commit resolves #3615

ramyaragupathy (commented):
A few observations:
* sortDirection and sortBy parameter has been reversed in place
* what's your idea for sortby=read asc and desc filters @willemarcel ?

ramyaragupathy:
Ignore the previous comments @willemarcel - it was just my local working directory

ramyaragupathy (approved):
Works as expected

PR #3600, opened 2020-09-15
Add sentry integration

willemarcel:


backend/__init__.py
```diff
@@ -9,10 +9,21 @@
...
-
 from backend.config import EnvironmentConfig
 
 
+def sentry_init():
+    """Initialize sentry.io event tracking"""
+    import sentry_sdk
+    from sentry_sdk.integrations.flask import FlaskIntegration
+
+    sentry_sdk.init(
```
willemarcel, line 20: @dakotabenjamin is [NetworkEnvironment](https://github.com/hotosm/tasking-manager/blob/feature/sentry-backend-integration/scripts/aws/cloudformation/tasking-manager.template.js#L7) available as a ENV VAR to the backend?
To set the environment on sentry, we just need to add `environment=foo` here

dakotabenjamin: can we use the stack naming convention for that, similar to how we do New Relic? https://github.com/hotosm/tasking-manager/blob/feature/sentry-backend-integration/scripts/aws/cloudformation/tasking-manager.template.js#L394

willemarcel: Yes, could we change the env var from `NEW_RELIC_ENVIRONMENT` to `TM_ENVIRONMENT`, so if we stop using NewRelic, we don't need to change the env var and make the variable name more expressive.

willemarcel:
@dakotabenjamin I also added sentry to the frontend and adjusted the backend sample rate

backend/__init__.py
dakotabenjamin: `NEW_RELIC_ENVIRONMENT` is something used internally by New Relic, so we can't change it.

dakotabenjamin:
@willemarcel I added the environment variable assignment, and made the generic "TM_ENVIRONMENT" envvar. For production it will show up as "tm4-production".

dakotabenjamin (commented):
tested on aws and was able to get the key to work in Sentry, but it says it's being rate limited. I'm not able to trigger any issues manually either. Everthing else seems to be working, so perhaps we merge into staging and test from there?

PR #3607, opened 2020-09-18
Apply subqueries to reduce response time in projects last activities …

JorgeMartinezG:
…endpoint

Response time reduced from 10s to ~500 ms

### DIFF
--- a/backend/models/dtos/mapping_dto.py
+++ b/backend/models/dtos/mapping_dto.py
@@ -69,7 +69,7 @@ class TaskStatusDTO(Model):
     task_id = IntType(serialized_name="taskId")
     task_status = StringType(serialized_name="taskStatus")
     action_date = UTCDateTimeType(serialized_name="actionDate")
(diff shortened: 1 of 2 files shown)


willemarcel (changes requested):
@JorgeMartinezG it's not returning data of all the tasks. Testing with a local project, the develop branch returns 170 tasks. This PR returns 101.

willemarcel:
@JorgeMartinezG I identified the issue. This new query is returning only data from tasks that had some activity. It would not be a problem, but we need to know when there are new tasks on the project (caused by task split) in order to update the map. The method we use on the frontend to decide when to fetch the tasks geojson endpoint is: do it when the number of tasks returned by the activities' endpoint changes in relation to the previous one.

JorgeMartinezG:
Thanks for reviewing wille. I'm checking and what is happening is that we do not capture the tasks that do not have changes since they do not appear in the task history table. The fix would be to get the ids first and then left outer join with task history

willemarcel (approved):
Perfect! Super improvement!

PR #3652, opened 2020-09-27
Add timezone setting

### LINKED ISSUE #3651: Backend raises a UserWarning: No timezone specified
```
backend       | /usr/local/lib/python3.7/site-packages/tzlocal/unix.py:158: UserWarning: Can not find any timezone configuration, defaulting to UTC.
```

To avoid this, the `TZ` environment variable should be set up. For example, `TZ=UTC` or `TZ=Europe/Rome`.

frafra:
Fix #3651.

### DIFF
--- a/scripts/docker/Dockerfile.backend
+++ b/scripts/docker/Dockerfile.backend
@@ -20,6 +20,7 @@ RUN pip install --no-cache-dir -r requirements.txt
 
 ## INITIALIZATION
 
+ENV TZ UTC # Fix timezone (do not change - see issue #3638)
 EXPOSE 5000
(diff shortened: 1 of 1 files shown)


willemarcel:
@frafra The backend is not prepared to deal with other time zones other than UTC, as it's not saving information on the database with time zone. I believe adding that setting will cause problems.

frafra:
Thanks for clarifying. Ok, so should be just set `TZ=UTC` as default environmental variable inside the Dockerfile just to mute the warning?

willemarcel:
Yes, adding that command to the Dockerfile seems good to me

PR #3654, opened 2020-09-27
Restart containers automatically

### LINKED ISSUE #3649: Use restart: always in docker-compose.yml
I suggest `restart: always` would be a good policy for production for all the services except migration (which has `restart: on-failure` at the moment and I would keep it like that).

frafra:
Fix #3649.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -16,6 +16,7 @@ services:
   backend:
     <<: *backend
     container_name: backend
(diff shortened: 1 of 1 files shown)


willemarcel:
Hi @frafra ! Thank you very much for your contributions. This is the last one missing to merge. Would you mind rebasing it and solving the conflict?

PR #3655, opened 2020-09-28
add missing MAX_AOI_AREA env to frontend expand file

geohacker:
The MAP_MAX_AREA env referenced in https://github.com/hotosm/tasking-manager/blob/develop/frontend/src/config/index.js#L22 is missing from the expand file.


willemarcel:
Thanks, @geohacker!

Issue #3554, opened 2020-08-28
Create TM embed widget


willemarcel:
Imagine someone wants to embed a list of Tasking Manager projects on another website, we could provide an embed code that would allow anyone to easily promote TM projects. It could be used in the [HOT](https://hotosm.org) and [Missing Maps](https://missingmaps.org/) websites and in other organizations and communities pages as well.

It should render a component like this, with a carousel to navigate through the list:
![image](https://user-images.githubusercontent.com/666291/91583040-067be280-e927-11ea-8d89-fab97fcc7a3b.png)

Some ideas:
- It should not be implemented as part of the TM frontend, but as an external project.
- It doesn't need necessarily to use React, it could be done with Svelte, Vue or vanilla JS. The only requirement is to implement exactly the same style.
- It should use the API endpoint: `https://tasking-manager-tm4-production-api.hotosm.org/api/v2/projects/` and allow to define some query params supported by the API, like organizations, campaigns and countries. Check the [API-Docs](https://tasks.hotosm.org/api-docs).

max-keviv:
@willemarcel can i handle this

willemarcel:
@max-keviv sure! Go for it!

willemarcel:
@max-keviv no, only the ones returned on the first page of the API pagination

max-keviv:
okay got it

max-keviv:
@willemarcel i am facing problem with styling of project card.  can you figure out steps to produce TM Embed widget

PR #3659, opened 2020-09-28
Create TM embed widget #3554

max-keviv:
It is responsive multi carousel of Tasking Manager Projects Cards
Screenshot from 2020-09-28 11-34-07

Screenshot from 2020-09-28 11-35-34

@willemarcel  please have review


willemarcel:
Thank you for your effort. As I mentioned on #3554: "It should **not** be implemented as part of the TM frontend, but as an external project.".

Issue #3344, opened 2020-07-17
Improve sign up error message


willemarcel:
![image](https://user-images.githubusercontent.com/666291/87780517-44082e80-c805-11ea-95f2-78ee5fb51b91.png)

This error is probably occurring because that email address is already registered, So we need to show a better message to the users.

k3KAW8Pnf7mkmdSMPHz27:
@willemarcel can you provide more information on this issue? I have a hard time reproducing it. Is it possible you have two accounts with the same e-mail address?
See,
https://github.com/hotosm/tasking-manager/blob/3d86c0dc967705cbbe9a26bffcc2a923e5698903/backend/services/users/user_service.py#L777-L779
where `one_or_none()` can raise `MultipleResultsFound` if multiple e-mails are found. It is handled in,
https://github.com/hotosm/tasking-manager/blob/3d86c0dc967705cbbe9a26bffcc2a923e5698903/backend/api/users/actions.py#L343-L349

Other than by having two identical e-mail addresses, I have a hard time not getting the correct error message, `f"Email address {user_dto.email} already exists"` shown.

willemarcel:
Sorry for the delay on answering you, @k3KAW8Pnf7mkmdSMPHz27. I confirmed it is working correctly. I just improved the frontend validation and error display with https://github.com/hotosm/tasking-manager/pull/3660

k3KAW8Pnf7mkmdSMPHz27:
Ok, no worries and great that it is working that it should!
I added some suggestions in #3662 for back-end changes related to detection of multiple e-mail addresses, feel free to add/change/discard the PR as it only indirectly relates to this issue.

PR #3662, opened 2020-09-28
Improve check for existing email addresses

k3KAW8Pnf7mkmdSMPHz27:
**Background**
When a user attempts to register a new account, the provided e-mail address is checked against the existing e-mails in the database to detect if there already exists an account with that e-mail address.

**The proposed changes** 
1. Perform the e-mail address comparison in lower-case to make it case-insensitive
2. If there are two or more users with the same e-mail address, return a 400 error (user already exists) instead of a 500 error (internal server error)

### DIFF
--- a/backend/services/users/user_service.py
+++ b/backend/services/users/user_service.py
@@ -777,7 +777,7 @@ class UserService:
     def register_user_with_email(user_dto: UserRegisterEmailDTO):
         # Validate that user is not within the general users table.
         user_email = user_dto.email.lower()
-        user = User.query.filter(func.lower(User.email_address) == user_email).one_or_none()
+        user = User.query.filter(func.lower(User.email_address) == user_email).first()
         if user is not None:
             details_msg = f"Email address {user_email} already exists"
             raise ValueError(details_msg)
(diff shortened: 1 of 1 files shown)


k3KAW8Pnf7mkmdSMPHz27:
Relates indirectly to #3344

Aadesh-Baral (approved):

k3KAW8Pnf7mkmdSMPHz27:
@Aadesh-Baral not sure what requirements you have, but if I'd redo this code today (i.e., three years later), I'd either use `casefold` or normalized Unicode (see https://docs.python.org/3/howto/unicode.html#comparing-strings for background).

On the other hand, since this has been fine for three years, it is unlikely it needs fixing.

Motivation for the original PR:
I was looking at mobile usability. Mobile phones often force an upper-case letter for the first character in an e-mail.
It is, fortunately, less of a problem these days when password managers are omnipresent.

Aadesh-Baral:
@k3KAW8Pnf7mkmdSMPHz27  Thank you for your comment! I understand your suggestion to use `casefold` or normalized Unicode for string comparison. While those approaches could potentially fine-tune the code, as you mentioned, it's not a high priority since the function in question isn't used frequently.

However, if you would like to create another pull request to implement the changes you suggested, I would be more than happy to review it. It's always good to consider improvements, especially if they can enhance the code's compatibility or robustness. Feel free to submit the PR.

PR #3669, opened 2020-09-29
Update existing country names for projects

ramyaragupathy:
Per discussion w/ @willemarcel - setting up a migration to move existing country names to English. Backend changes have already been made to fetch English names during project creation.

### DIFF
--- a/migrations/versions/7937dae319b5_.py
+++ b/migrations/versions/7937dae319b5_.py
@@ -1,0 +1,54 @@
+""" Ensure Project's country names are in English
+
+Revision ID: 7937dae319b5
+Revises: 14842761654b
+Create Date: 2020-09-21 17:17:10.542429
+
+"""
+from alembic import op
+import requests
+
+
+# revision identifiers, used by Alembic.
+revision = "7937dae319b5"
+down_revision = "14842761654b"
+branch_labels = None
+depends_on = None
+nominatim_base_url = "https://nominatim.openstreetmap.org"
+nominatim_data_format = "&format=jsonv2&accept-language=en&namedetails=1"
+
+
+def upgrade():
+    conn = op.get_bind()
+    # fetch existing country names
+    fetch_countries = (
+        "select distinct(unnest(country)) from projects where country is not null;"
+    )
+    countries = conn.execute(fetch_countries)
+    for country in countries:
+        country = country[0]
+        # search by name
+        url = nominatim_base_url + "/search?country=" + country + nominatim_data_format
+        country_search = requests.get(url).json()
+        if country_search != []:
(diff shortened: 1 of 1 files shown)


migrations/versions/7937dae319b5_.py
```diff
@@ -0,0 +1,65 @@
...
+    conn = op.get_bind()
+    # fetch existing country names
+    fetch_countries = (
+        "select distinct(unnest(country)) from projects where country is not null;"
+    )
+    countries = conn.execute(fetch_countries)
+    for country in countries:
+        country = country[0]
+        # search by name
+        url = nominatim_base_url + "/search?country=" + country + nominatim_data_format
```
willemarcel, line 31: If we add `namedetails=1` param to that request we don't need to call the reverse geocoding endpoint, as it will return all the names 😃

willemarcel (changes requested):
It's possible to do it with only one request. I commented on the code

migrations/versions/7937dae319b5_.py
```diff
@@ -0,0 +1,65 @@
...
+        )
+        country_info = requests.get(url).json()
+        if country_info != []:
+            updated_country_name = country_info.get("address").get("country")
+            if country != updated_country_name:
+                special_char = updated_country_name.find("'")
+                if special_char >= 0:
+                    updated_country_name = (
+                        updated_country_name[:special_char]
+                        + "'"
```
willemarcel, line 50: It's duplicating the single quote

willemarcel: On https://github.com/hotosm/iso-countries-languages/blob/master/res/countries/en.json, we have data exclusively from OSM and there isn't countries with `'` on the name. So I believe it's not necessary to handle this case.

migrations/versions/7937dae319b5_.py
ramyaragupathy: That was handy - I couldn't spot this initially while going through the nominatim API. Using it to avoid reverse geocoding

migrations/versions/7937dae319b5_.py
ramyaragupathy: Yes in order to handle special chars like single quote in sql queries - it needs to be escaped. Though the link you shared has `Ivory Coast` listed as English name for `Côte d'Ivoire` - recently updated data on nominatim has [the name listed for English with special chars](https://nominatim.openstreetmap.org/search?country=C%C3%B4te%20d%27Ivoire&format=jsonv2&accept-language=en&namedetails=1). `Ivory Coast` is listed under `old_name:en`. To handle this and escape special character in sql query we still need need this function.

willemarcel (approved):
@ramyaragupathy it worked well! Thanks.

Issue #2342, opened 2020-02-18
Review mobile version


pantierra:
* [ ] Order of navigation items in hamburger menu
* [ ] Notifications pop-up misplaced
* [ ] Check on project action bar
* [ ] Filter of "My Tasks" page
* [ ] Contributions timeline
* [ ] Main project page
* [x] Learn pages

-----

**1. Order of navigation items in hamburger menu**

![image](https://user-images.githubusercontent.com/97706/79884678-fb4b0000-83e4-11ea-8c35-f7451fbe2f6b.png)

It should separate the two different navigations and not mix them up :smile: 

-----

**2. Notifications pop-up misplaced**

![image](https://user-images.githubusercontent.com/97706/79884008-1ff2a800-83e4-11ea-9e0f-0a4daa062c06.png)

-----

**3. Check on project action bar**

![image](https://user-images.githubusercontent.com/97706/79884061-339e0e80-83e4-11ea-9fd1-4306b7287074.png)

-----

**4. Filters of "My Tasks" and "Explore projects pages**

![image](https://user-images.githubusercontent.com/97706/79883755-d2763b00-83e3-11ea-9240-35cdf5c3047d.png)

![image](https://user-images.githubusercontent.com/97706/79884318-8d063d80-83e4-11ea-8d39-b055ada7bd02.png)

-----

**5. Contributions timeline**

![image](https://user-images.githubusercontent.com/97706/79884154-53353700-83e4-11ea-8c30-0840572caaac.png)

-----

**6. Main project page**

![image](https://user-images.githubusercontent.com/97706/79884274-78c24080-83e4-11ea-872a-e83eb5c7c092.png)

-----

**7. Learn pages**

![image](https://user-images.githubusercontent.com/97706/79884495-c2129000-83e4-11ea-870a-8eca8fb07523.png)

willemarcel:
> when I clicked on the 'hamburger' menu (3 bars top right), the pop-up that appeared was very wide, and off my screen to the right 

I couldn't reproduce this... It would be good if we could have a screenshot.

The other problem reported (about the popup not going away when clicking on a link on it) was solved by https://github.com/hotosm/tasking-manager/pull/2365

xamanu:
An issue reported in #2555: Learn pages are not responsive.

PR #3675, opened 2020-10-01
Change responsive behaviour of notification pop-up

k3KAW8Pnf7mkmdSMPHz27:
Attempts to improve on "Notifications pop-up misplaced" in #2342. As always, add/remove/change/discard/comment on this PR, whichever suits best.

The PR sets the distance from the right side of the browser window to 0 for small displays, while keeping the current solution for all other browser sizes.

This is the "low hanging fruit" version of "solving" this issue.

**Alternatively**
I can take a look att anchoring the arrow to the bell icon. However, I don't think I can solve that problem without using `flexbox` and limiting the display width of the user name for medium screen sizes. I can go into more depth regarding why if that is interesting. The attached image shows two issues not addressed with the current approach
1. The header wraps for long names
2. The pop-up arrow isn't pointing towards the bell icon

Skärmavbild 2020-10-01 kl  17 05 53

### DIFF
--- a/frontend/src/assets/styles/_extra.scss
+++ b/frontend/src/assets/styles/_extra.scss
@@ -49,6 +49,9 @@ html {
   .user-popup-content {
     width: 13% !important;
   }
+  .right-4-ns {
+    right: 4rem;
+  }
 }
 
 @media screen and (min-width: 66rem) {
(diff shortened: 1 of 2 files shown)


k3KAW8Pnf7mkmdSMPHz27:
Expected appearance on iPhone 6/7/8/similar size
Skärmavbild 2020-10-01 kl  17 18 06

willemarcel:
@k3KAW8Pnf7mkmdSMPHz27 thanks a lot for this PR! I found a better solution: https://github.com/hotosm/tasking-manager/pull/3680
It's not working perfectly on Chrome though. Let me know if you have a suggestion.

PR #3680, opened 2020-10-02
Adjust position of notification popup tooltip

willemarcel:
Status: working well on Firefox, but not perfectly on Chrome.

### DIFF
--- a/frontend/src/components/header/notificationBell.js
+++ b/frontend/src/components/header/notificationBell.js
@@ -14,6 +14,7 @@ export const NotificationBell = (props) => {
   const token = useSelector((state) => state.auth.get('token'));
   const trigger = token !== null;
   const [forceUpdated, forceUpdate] = useForceUpdate();
+  const [bellPosition, setBellPosition] = useState(0);
   /* these below make the references stable so hooks doesn't re-request forever */
   const notificationBellRef = useRef(null);
   const params = useRef({ status: 'unread' });
@@ -51,6 +52,7 @@ export const NotificationBell = (props) => {
   const handleBellClick = (e) => {
     e.preventDefault();
     e.stopPropagation();
+    setBellPosition(window.innerWidth - e.target.getBoundingClientRect().right);
     setPopoutFocus(!isPopoutFocus);
     if (unreadNotifications) {
       forceUpdate(); // update the notifications when user clicks and there are unread messages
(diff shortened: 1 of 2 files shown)


k3KAW8Pnf7mkmdSMPHz27:
@willemarcel what issues are you having in Chrome? I get the same result in Firefox and Chrome (and both looks good to me)

There is some nitpicking which you can choose to ignore if you don't agree,

* Update the location of the pop-up as well. This would avoid the unlikely issue of people having inconveniently long display names causing a misplacement of the pop-up (or perhaps I should change my nickname generator 😞 ))
Skärmavbild 2020-10-02 kl  16 11 37

* would you consider making the location of the pop-up conditional? It would give an extra 40px for users with small screens, I made an example anchoring on the left,
https://github.com/k3KAW8Pnf7mkmdSMPHz27/tasking-manager/blob/e63f3acdf42b2dab1862adf8dbc23d9532951035/frontend/src/components/header/notificationBell.js#L55
and with the (conditional) popup position in,
https://github.com/k3KAW8Pnf7mkmdSMPHz27/tasking-manager/blob/e63f3acdf42b2dab1862adf8dbc23d9532951035/frontend/src/views/notifications.js#L19-L27
Unfortunately, the arrow would have to be repositioned as well. I took a rough guess at a good position in the code but it can be improved.
* Is updating the position on the resizing event an option? It is **REALLY NOT IMPORTANT** if performance is a concern but it would avoid the potential of,
Skärmavbild 2020-10-02 kl  14 58 57

willemarcel:
@k3KAW8Pnf7mkmdSMPHz27 I liked all your 3 first proposals. Resizing event would be good too, but not a priority on this first iteration. Would you like to make a PR on top of that?

The problem that I was having on Chrome is actually happening in Firefox too when I don't have the responsive design mode active. 
image

k3KAW8Pnf7mkmdSMPHz27:
> The problem that I was having on Chrome is actually happening in Firefox too when I don't have the responsive design mode active.

@willemarcel that is strange, I can't replicate it. (Chrome Version 85.0.4183.121 on Chrome Mac OS 10.15.6)
I can get it to look strange by adjusting the default font-size, but that creates other visual artifacts as well (the bell proportions)
Skärmavbild 2020-10-05 kl  14 47 59

> Would you like to make a PR on top of that?

Open a PR into this branch? (`fix/notif-popup-tooltip-position`)



If you are willing to test it afterwards, I can set the arrow's position within the DOM/CSS and see if fixes the arrow's position outside of the responsive mode?

willemarcel:
You can open a PR to develop including my commit and yours.

> If you are willing to test it afterwards, I can set the arrow's position within the DOM/CSS and see if fixes the arrow's position outside of the responsive mode?

yes, sure

willemarcel:
Merged it through #3886

PR #3685, opened 2020-10-05
Handle null campaigns in PATCH

ramyaragupathy:
To test, try saving a project with null campaigns and this should get through

### DIFF
--- a/backend/models/postgis/project.py
+++ b/backend/models/postgis/project.py
@@ -447,12 +447,16 @@ class Project(db.Model):
         else:
             if self.custom_editor:
                 self.custom_editor.delete()
-
-        self.campaign = []
-        if project_dto.campaigns:
-            self.campaign = Campaign.query.filter(
-                Campaign.id.in_([c.id for c in project_dto.campaigns])
-            ).all()
+        # handle campaign update
+        try:
+            new_ids = [c.id for c in project_dto.campaigns]
+            new_ids.sort()
+        except TypeError:
+            new_ids = []
+        current_ids = [c.id for c in self.campaign]
+        current_ids.sort()
+        if new_ids != current_ids:
+            self.campaign = Campaign.query.filter(Campaign.id.in_(new_ids)).all()
 
         if project_dto.mapping_permission:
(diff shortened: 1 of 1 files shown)


willemarcel:
@ramyaragupathy I made the query more efficient. A yet better solution would be:

```
        try:
            new_ids = [c.id for c in project_dto.campaigns]
        except TypeError:
            new_ids = []
        current_ids = [c.id for c in self.campaign]
        if new_ids != current_ids:
            self.campaign = Campaign.query.filter(
                Campaign.id.in_(new_ids)
            ).all()
```

@JorgeMartinezG some opinion on this?

willemarcel:
@ramyaragupathy We should apply the same solution to the Interests field

JorgeMartinezG:
What about this? If you want to avoid some unneeded query when model and dto have same ids

```
self.campaign = []
if project_dto.campaign is not None:
    model_ids = [c.id for c in self.campaign]
    dto_ids = [c.id for c in project_dto.campaigns]

    # Get ids that are in dto but not in model (New campaigns).
    new_ids = [i for i in dto_ids if i not in model_ids]

    # Get ids that are in model but not in dto (Remove campaigns).
    removed_ids = [i for i in model_ids if i not in dto_ids]

    # If there is at least one item within the two lists, update.
    if len(new_ids) > 0 or len(removed_ids) > 0:
        self.campaign = [Campaign.query.get(i) for i in dto_ids]
```

willemarcel:
@JorgeMartinezG your proposal has some problems:
- If you makes `self.campaign` empty on the beginning, `model_ids` will be empty
- If we don't make it empty and keep that `if` statement, we can not make it empty by desire of the user
- `dto.campaigns` can be none and raise `TypeError`

I added the `.sort()` to my proposal

JorgeMartinezG:
Corrected. BTW, everything is jumped when project_dto is None, not empty.

the dto has campaigns field with a default of [], so it should deserialize to an empty list if campaigns is not provided in the endpoint.

willemarcel (approved):

PR #3692, opened 2020-10-06
Simplify CircleCI config and fix sentry envvar setting

dakotabenjamin:
Despite the branch name, the primary feature here is the improvement on the circleci config. Using [contexts](https://circleci.com/docs/2.0/env-vars/#setting-an-environment-variable-in-a-context), we simplify the circleci script. The context will be set on CircleCI for each deployment (so far only for staging frontend, hence WIP) removing the need for most of the parameters, except the stack name. CircleCI does not allow us to use the reserved "name" key for the job parameters. 

Incidentally this PR will also fix the missing frontend environment variable for Sentry integration by simplifying the envvar setting. We can use `${STACK_NAME}` for the frontend and backend sentry `environment` parameter.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -263,111 +263,16 @@ jobs:
             aws rds delete-db-snapshot --db-snapshot-identifier tm3-<< parameters.stack_name >>-${RDS_ID}-latest
           fi
   frontend_deploy:
+    working_directory: /home/circleci/tasking-manager
+    docker:
+    - image: circleci/python:3.7-buster-node
     parameters:
       stack_name:
         description: "the name of the stack for cfn-config"
         type: string
-      environment_name:
-        description: "The environment (staging or production) on AWS"
-        type: enum
-        enum: ["staging", "production"]
-      autoscaling_policy:
-        description: "Setting autoscaling parameters"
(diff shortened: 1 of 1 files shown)


dakotabenjamin:
Ready for testing at https://tasks-test.hotosm.org/

Major changes include: 
- Frontend variables are saved in a CircleCI Context
- Backend does not need any CI variables anymore, those are pulled from the cfn-config configuration file on AWS S3
- consequently, that long JSON_CONFIG variable previously used for backend deployment is gone now :grinning: 
- If we want to change the Cloudformation infra parameters, we can do so from the command line using cfn-config. those changes will be saved to s3 so CircleCI will not affect them. We do have to be aware of the GitSha parameter when doing so. 
- I have left in the config a workflow called "test" which will deploy to a CloudFormation environment called tasking-manager-test. In the future we can use that for testing deployments but will need to setup that environment manually first, then change the branch on the workflow.

dakotabenjamin:
Before merging we should make sure we have new contexts created for production, assisted, and teachosm

eternaltyro (approved):
LGTM

willemarcel (approved):

PR #3722, opened 2020-10-14
Improve user tasks card and add tests

yunjuc:
- Improve user tasks card display
- Add tests for ElementsMapped & TaskStats components

New layout
Screen Shot 2020-10-13 at 10 08 54 AM


willemarcel (approved):
Thanks a lot, @yunjuc !

PR #3731, opened 2020-10-15
get env file into frontend for docker

### LINKED ISSUE #3721: can't save in iD (with docker setup)
https://www.openstreetmap.org/oauth/request_token is giving me an HTTP 500 error.  it looks like the only parameter is the `Autorization` header, which had an empty string for the `oauth_consumer_key` value.

i discovered that env vars from tasking-manager.env aren't passed to the frontend during the docker build phase (specifically `TM_CONSUMER_KEY`).  and i couldn't figure out how to tell docker compose to use a environment file during the build phase.  i added all of the environment variables separately except the secrets, so now `oauth_consumer_key` has a value but i'm still getting an HTTP 500 error.

i also noticed the frontend uses `REACT_APP_OSM_CONSUMER_SECRET` (which i didn't set during the build) but i'm wondering if that is correct/safe?  it could also be why i'm still getting the 500 error.

also i'm pretty sure `TM_APP_API_URL` should have `/api` in `example env`, but it's overridden in `docker-compose.override.yml` so i guess it makes no difference.

jayenashar:
fixes #3721


willemarcel:
@jayenashar sorry for the delay on giving you a feedback. We have improved the example.env with a better explanation and a better initial setup to those who are setting a local development instance. Could you take a look on how it is now and propose the better solution to the docker setup?

Maybe setting `TM_APP_BASE_URL` and `TM_APP_API_URL` by another way on the docker setup? or appending it to the end of the `tasking-manager.env`?

jayenashar:
i think copying the .env file in the dockerfile is a fine solution.

or are you saying `TM_APP_BASE_URL` and `TM_APP_API_URL` are different for local dev vs docker?

willemarcel:
The issue is that what I recommend for a local dev setup is to run the backend on localhost:5000 and the frontend on localhost:3000. On our production setup we deploy backend and frontend on different domain names, but on the docker setup it is sharing the same domain with the backend on `/api`

jayenashar:
thanks for the explanation.  i'll update this PR when i have a chance, with what i think would work.

willemarcel:
Merged through #4633

PR #3744, opened 2020-10-20
Style emails, set ORG_LOGO var, remove non used image assets

willemarcel:


ramyaragupathy (commented):
Tested out messages for the following scenarios:
* welcome message 
* Validation - yes and no
* mention in project chat

image

image

image

image

PR #3772, opened 2020-10-23
Replace logger.error by .debug on Exceptions

### LINKED ISSUE #3541: Analyse reason why some errors are so frequent on logs
- Was the task unlocked automatically and the user tried to update after the auto unlock?

`ERROR: Status must be LOCKED_FOR_MAPPING to unlock [in /tasking-manager/backend/services/mapping_service.py:28]`

`ERROR: Task 2398 is not LOCKED_FOR_VALIDATION [in /tasking-manager/backend/services/validator_service.py:35]`

Possible solution is https://github.com/hotosm/tasking-manager/issues/3539

- The only possibility I see to those errors happening is if the user tried to lock a task, but someone else has already locked it (we refresh tasks states each 60s).

`ERROR: Task 733 is not MAPPED, BADIMAGERY or VALIDATED [in /tasking-manager/backend/services/validator_service.py:35]`

`ERROR: Task in invalid state for mapping [in /tasking-manager/backend/services/mapping_service.py:28]`

Possible solution is #3540

When fixed those issues, we should consider stop logging those exceptions as errors.

willemarcel:
Resolves #3541

### DIFF
--- a/backend/models/postgis/utils.py
+++ b/backend/models/postgis/utils.py
@@ -17,7 +17,7 @@ class InvalidGeoJson(Exception):
 
     def __init__(self, message):
         if current_app:
-            current_app.logger.error(message)
+            current_app.logger.debug(message)
 
 
 class UserLicenseError(Exception):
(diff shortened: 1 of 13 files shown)


willemarcel:
Reference on when to use logging: https://docs.python.org/3/howto/logging.html#when-to-use-logging

Issue #3802, opened 2020-11-02
Project save fails when there is no country information


ramyaragupathy:
While attempting to save a project, if the project is missing information - then the code should set the country info using the centroid information. However this part is failing with a 500 error message.

PR #3812, opened 2020-11-03
Handle NoneType condition check during project update

ramyaragupathy:
Fixes https://github.com/hotosm/tasking-manager/issues/3802


willemarcel (approved):
Thanks, @ramyaragupathy!

PR #3838, opened 2020-11-06
Replace injectIntl by useIntl + replace props by explicit object

willemarcel:


d-rita (commented):
The branch needs to be updated with the develop branch, otherwise LGTM 🎉

PR #3842, opened 2020-11-07
Improve TM CFN template format

eternaltyro:
- Parameterize Backend Instance AMI ID
- Use better parameter types to manage them easier
- Add Systemd service unit for Tasking Manager inline
- Remove redundancies in apt installs
- Correct IAM role for DatabaseDumpAccess and simplify

TODO:
- Shorten security group names
- Test parameter types
- Test new Ubuntu 16.04 AMI ID
- Test NoEcho for password strings
- Test systemd service for tasking manager gunicorn process
- Move install location for tasking manager from /


willemarcel:
> Test systemd service for tasking manager gunicorn process
We are right now starting gunicorn by a newrelic command, so it may require a special setup

PR #3853, opened 2020-11-09
Refactor name and URL CustomEditorForm inputs

d-rita:
**What does this PR do?**
Refactor text inputs in the CustomEditorForm

**Steps Taken**
- Create a CustomEditorTextInput component
- Add conditional rendering of the input label
- Refactor name and URL inputs using the new component

### DIFF
--- a/frontend/src/components/projectEdit/customEditorForm.js
+++ b/frontend/src/components/projectEdit/customEditorForm.js
@@ -1,87 +1,95 @@
-import React, {useContext} from 'react';
+import React, { useContext } from 'react';
 import { FormattedMessage } from 'react-intl';
 
 import messages from './messages';
-import {StateContext, styleClasses} from '../../views/projectEdit';
+import { StateContext, styleClasses } from '../../views/projectEdit';
 import { Button } from '../button';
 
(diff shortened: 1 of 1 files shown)


frontend/src/components/customEditorInput.js
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { styleClasses } from '../views/projectEdit';
+import { FormattedMessage } from 'react-intl';
+import messages from './projectEdit/messages';
+
+export const CustomEditorTextInput = ({ name, value, handleChange }) => {
```
willemarcel, line 6: As it doesn't have the possibility to be reused on other components, I prefer to let it on the `projectEdit/customEditorForm.js` file.

willemarcel (changes requested):
@d-rita a small suggestion of change. Also make sure to have the [git pre-commit hook](https://github.com/hotosm/tasking-manager/blob/develop/docs/contributing-code.md#committing) setup so, the indentation of customEditor file will be fixed (it's the only file with 4 spaces indentation on the frontend)

willemarcel (approved):

PR #3856, opened 2020-11-10
Remove nodejs from cfn template

willemarcel:
We don't need nodejs on the backend, since we are not building or serving the frontend from there.


dakotabenjamin (approved):
tested on AWS and got a healthy backend response.

PR #3859, opened 2020-11-11
Add action filters for list projects endpoint

willemarcel:
Filter projects the user can map. It considers:
- the level and teams of the user
- the permissions and teams of the project
- if there are tasks that needs to be mapped

You can test it with https://tasks-stage.hotosm.org/explore?action=map
compare the results with https://tasks-stage.hotosm.org/explore


willemarcel:
I added also the validate filter: https://tasks-stage.hotosm.org/explore?action=validate

d-rita (commented):
This works well, and the separation of the validation + mapping action filters logic into separate static methods is good. 👍🏾

ramyaragupathy (approved):
Tested with frontend changes - works good

PR #3861, opened 2020-11-11
Remove enabled field from CustomEditor model

willemarcel:
- Handle CustomEditor activation directly on project's mappingEditors and validationEditors
- Use switch toggle instead of checkbox


migrations/versions/14842761654b_.py
```diff
@@ -0,0 +1,31 @@
...
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.drop_column("project_custom_editors", "enabled")
+    # ### end Alembic commands ###
+
+
+def downgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column(
+        "project_custom_editors",
+        sa.Column("enabled", sa.BOOLEAN(), autoincrement=False, nullable=False),
```
willemarcel, line 29: @ramyaragupathy I had problems to downgrade with the `nullable=False`. Do you recommend a better solution?

d-rita (approved):

PR #3893, opened 2020-11-19
Refactor Campaign Service and Campaign static methods

d-rita:
**What does this PR do?**
This PR refactors the Campaign and Campaign Service static methods.

**Steps Taken:** 
- Add a new static method in the Organisation Service class to get all the organisations under a campaign
- Add a new static method to convert a collection of campaigns to a List DTO
- Move replicated static methods from the Campaign model to the Campaign Service class
- Refactor methods to make use of newly added ones.

**Some of the URLs to test:**
1) Via Postman. Method: `GET`. 
- `http://127.0.0.1:5000/api/v2/campaigns/`
- `http://127.0.0.1:5000/api/v2/campaigns/<int:campaign_id>/`
- `http://127.0.0.1:5000/api/v2/organisations/<int:organisation_id>/campaigns/`
- `http://127.0.0.1:5000/api/v2/projects/<int:project_id>/campaigns/`
2) Frontend
- `http://localhost:3000/manage/campaigns/`
- `http://localhost:3000/manage/campaigns/<int:campaign_id>`

### DIFF
--- a/backend/services/organisation_service.py
+++ b/backend/services/organisation_service.py
@@ -8,6 +8,7 @@ from backend.models.dtos.organisation_dto import (
     ListOrganisationsDTO,
     UpdateOrganisationDTO,
 )
+from backend.models.postgis.campaign import campaign_organisations
 from backend.models.postgis.organisation import Organisation
 from backend.models.postgis.project import Project, ProjectInfo
 from backend.models.postgis.utils import NotFound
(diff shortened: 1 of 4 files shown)


backend/services/organisation_service.py
```diff
@@ -207,3 +208,34 @@ def is_user_an_org_manager(organisation_id: int, user_id: int):
...
+        """
+        organisation_list_dto = ListOrganisationsDTO()
+        orgs = (
+            Organisation.query.join(campaign_organisations)
+            .filter(campaign_organisations.c.campaign_id == campaign_id)
+            .all()
+        )
+
+        for org in orgs:
+            if user_id != 0:
```
willemarcel, line 225: I don't know if we need that restriction. Some opinion, @ramyaragupathy ?

willemarcel:
I like the new structure. We are not supporting to assign organizations with campaigns on the frontend and we are not using that information of organizations related to a campaign.

I think the best would be to remove that organization field from the campaign detail endpoint. When we have that need we can build a new endpoint: `/api/v2/campaigns/<int:campaign_id>/organisations/`

willemarcel (changes requested):
comment above

d-rita:
> comment above

willemarcel (approved):
Awesome! Thanks!

Issue #3852, opened 2020-11-09
Add new search param to refine mapping type filter


willemarcel:
The search param `/api/v2/projects/?mappingTypes=BUILDINGS` right now returns projects whose mapping type field contains buildings, but includes projects set with buildings and other types.

Add a new search param that would filter projects strictly, so in the case mentioned above it would return projects whose mapping type is only BUILDINGS

d-rita:
@willemarcel, To add to this functionality, what should happen when a person selects multiple mapping types? Should the result include projects with all mapping types or any of them? For example, should the search query `/api/v2/projects/?mappingTypes=BUILDINGS, ROADS` return projects strictly with both types, or can the search results include projects with any and both types?

willemarcel:
@d-rita On the backend, I propose to add a new search param, it could be named `mappingTypesExact` to receive a true / false value. The default would be `false` and would make `mappingTypes` work as it is now. If it is `true`, would switch to the exact match. 

Then on the frontend we can add a switch.

@ramyaragupathy what do you think?

ramyaragupathy:
@willemarcel - Yes that sounds right - so it will be,
`https://tasking-manager-tm4-production-api.hotosm.org/api/v2/projects/?mappingTypes=ROADS&mappingTypesExact=true`

It would be nice to have these two as interdependent parameters, so that it `mappingTypesExact` becomes required only if `mappingTypes` is specified. I am checking out options in flask framework for this implementation, @willemarcel - if you're already aware of this, please add.

willemarcel:
@ramyaragupathy I think it's not required to do that check, only assure that `mappingTypesExact` would have any effect if it's alone. If it's not present, the default value is `false`.

PR #3904, opened 2020-11-20
Search for projects by exact mapping type

d-rita:
**What does this PR do?**
Give the user the option of searching for projects matching the exact mapping type or get all projects that contain the mapping type. 

**Steps Taken:**
- Add mappingTypesExact search param
- Return only projects of the given mapping type if mappingTypesExact is True. 
-  Return all projects containing the mapping type(s) if mappingTypesExact is False.

**Related issues:**
#3852 

**Test:**
- Using Postman with the `GET` method, test for projects of different mapping types using: `http://127.0.0.1:5000/api/v2/projects/?mappingTypes=<string:mapping_type>&mappingTypesExact=true`
- For example: ` http://127.0.0.1:5000/api/v2/projects/?mappingTypes=ROADS&mappingTypesExact=true`


backend/api/projects/resources.py
```diff
@@ -520,6 +520,9 @@ def setup_search_dto(self):
             search_dto.mapping_types = map(
                 str, mapping_types_str.split(",")
             )  # Extract list from string
+        search_dto.mapping_types_exact = strtobool(
```
willemarcel, line 523: On the line 572, you need to add the new field on the view's docstrings, so it will generate the documentation on Swagger

backend/services/project_search_service.py
```diff
@@ -268,7 +268,10 @@ def _filter_projects(search_dto: ProjectSearchDTO, user):
                 for mapping_type in search_dto.mapping_types
             ]
 
-            query = query.filter(Project.mapping_types.contains(mapping_type_array))
+            if search_dto.mapping_types_exact and len(mapping_type_array) == 1:
```
willemarcel, line 271: I don't like the previous implementation with contains and that limitation of only one `mappingType` on the exact matching, because it is not intuitive to the users (we allow users to search for more than one type) and it's quite restrictive.

For example, I would like to get all projects to map buildings or roads, I was expecting to receive projects:
- with only buildings
- with only roads
- with only these two types
- with these two types and others

or projects with only the two types on the exact match search.

If the implementation is too difficult or expensive to the database, we can keep that way. Do you have an opinion, @ramyaragupathy?

willemarcel (changes requested):

backend/services/project_search_service.py
d-rita: Let me see the changes I can make to get this working in that format.

willemarcel (approved):
Fantastic solution!!!

willemarcel:
@d-rita I added the field on the frontend. Could you test it?

PR #3919, opened 2020-11-23
[WIP] Encode URI for images in markdown comments

### LINKED ISSUE #3917: TM comment image uploads fail with spaces in file name
Windows (and i believe Mac OS also) allow for spaces in file names.

When a user uploads an image with spaces in the filename it will fail.

As an example see the comments in this task:
https://tasks.hotosm.org/projects/9190/tasks/?search=4166

A non tech savvy user will not be able to solve this by her/himself, resulting possibly that (s)he will renounce the upload.

Could it be possible to allow spaces in image filenames?

Regards @Martien (Slack)

k3KAW8Pnf7mkmdSMPHz27:
Fixes #3917. 

I have had some issues with my setup, so I can't test/verify the changes right now. It should be solved tomorrow.
I'll add test cases as soon as my local setup works. As always, add/change/delete anything.

- [x] Encodes the image URL in the task comments markdown (`alt`)
- [x] Use file name instead of 'image' as alternative text (this gives the user some control over the alternative text, albeit it is unlikely to be useful)
- [ ] Check if any other Dropzone has the same issue.
- [ ] Add test cases

### DIFF
--- a/frontend/src/components/comments/commentInput.js
+++ b/frontend/src/components/comments/commentInput.js
@@ -12,7 +12,8 @@ import { DROPZONE_SETTINGS } from '../../config';
 
 export const CommentInputField = ({ comment, setComment, enableHashtagPaste = false }: Object) => {
   const token = useSelector((state) => state.auth.get('token'));
-  const appendImgToComment = (url) => setComment(`${comment}\n![image](${url})\n`);
+  const appendImgToComment = (url) =>
+    setComment(`${comment}\n![${url.substring(url.lastIndexOf('/') + 1)}](${encodeURI(url)})\n`);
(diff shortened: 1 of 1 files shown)


willemarcel:
@k3KAW8Pnf7mkmdSMPHz27 I decided to solve the problem on the upload (#3920), but for sure we can do that improvement as well

k3KAW8Pnf7mkmdSMPHz27:
> @k3KAW8Pnf7mkmdSMPHz27 I decided to solve the problem on the upload (#3920),

@willemarcel great! Then I can take it easy getting my local setup going again 😃 

> but for sure we can do that improvement as well

I don't see the point in implementing this feature by itself. It will force someone to review code that might contain a "nice-to-have" feature but with low utility.

PR #3920, opened 2020-11-24
Slugify image filenames to upload

### LINKED ISSUE #3917: TM comment image uploads fail with spaces in file name

willemarcel:
Resolves #3917

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -60,6 +60,7 @@
     "redux-thunk": "^2.3.0",
     "short-number": "^1.0.7",
     "shpjs": "^3.6.2",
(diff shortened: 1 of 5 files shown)


d-rita (approved):
LGTM

PR #3926, opened 2020-11-25
Improve learn page

willemarcel:
- Submenu with the same style as the management section
image

- Link to Explore Projects page:
image

- Add missing key attr on react elements
- Translate some strings
- Remove typo 'bac' attr
- Remove unneeded return statements


d-rita (approved):
LGTM

PR #3940, opened 2020-11-30
Add Intl.Locale polyfill

willemarcel:


willemarcel:
It didn't solve the problem.

PR #3941, opened 2020-12-01
Simplify leading, trailing slash strip

eternaltyro:
Simplifies the operation and makes it slightly faster and more readable


willemarcel (approved):

willemarcel:
Thanks, @eternaltyro !

PR #3950, opened 2020-12-02
Use .contains instead of .ilike on user filtering

### LINKED ISSUE #842: [admin] user name search works with "starts with"
In the users list, the search engine looks for usernames starting with the given pattern.
It would probably be better to be able to use a "contains" search.

willemarcel:
Resolves #842

### DIFF
--- a/backend/models/postgis/user.py
+++ b/backend/models/postgis/user.py
@@ -151,7 +151,7 @@ class User(db.Model):
             ]
             base = base.filter(User.mapping_level.in_(mapping_level_array))
         if query.username:
-            base = base.filter(User.username.ilike(query.username.lower() + "%"))
+            base = base.filter(User.username.contains(query.username.lower()))
 
         if query.role:
             roles = query.role.split(",")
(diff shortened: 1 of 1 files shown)


backend/models/postgis/user.py
```diff
@@ -151,7 +151,7 @@ def get_all_users(query: UserSearchQuery) -> UserSearchDTO:
             ]
             base = base.filter(User.mapping_level.in_(mapping_level_array))
         if query.username:
-            base = base.filter(User.username.ilike(query.username.lower() + "%"))
+            base = base.filter(User.username.contains(query.username.lower()))
```
d-rita, line 154: When I make a search for a character for example the letter 'a', this works well and returns usernames that contain it. 
Screenshot 2020-12-03 at 11 49 49
However if I try searching for a complete name, it does not return it. 
Screenshot 2020-12-03 at 11 49 58
I wonder if we could just build on the previous implementation. Something like` base.filter(User.username.ilike(("%" + query.username + "%"))` so that we can get names that contain a character or the name itself.

d-rita (changes requested):

PR #3954, opened 2020-12-04
Request users to view task history comments

### LINKED ISSUE #3507: Add message asking user to read comments
If the task was invalidated previously, ask users to read the comments when mapping it again

d-rita:
What does this PR do?
Resolves #3507 

Steps taken:
- Create a custom hook for conditional render of request to read comments if the task was previously invalidated and has comments to guide the mapper.

Screenshot:
Screenshot 2020-12-13 at 15 54 50

### DIFF
--- a/frontend/src/components/svgIcons/index.js
+++ b/frontend/src/components/svgIcons/index.js
@@ -72,3 +72,4 @@ export { SidebarIcon } from './sidebar';
 export { QuestionCircleIcon } from './questionCircle';
 export { ChartLineIcon } from './chart';
 export { EditIcon } from './edit';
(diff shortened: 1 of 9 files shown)


willemarcel (changes requested):
Thanks, @d-rita!
We need some adjustments:
1. Only show the message if:
- The task has been invalidated previously (check the history for a STATE_CHANGE to INVALIDATED
- There is at least one comment on the history (the history includes both comments and state changes)
2. Let's try to use a tooltip instead of a banner? When I click on the history tab, it's flicking to the top. I think a tooltip will call more attention from the user too.

willemarcel (approved):

PR #3963, opened 2020-12-08
Polyfill Intl.Locale when needed

willemarcel:
Tasking Manager was failing on some browsers like Safari 13 and below. With the `Intl.Locale` polyfill we expand the browser support.


d-rita (approved):
LGTM

PR #3968, opened 2020-12-10
Give more visibility to our code of conduct

willemarcel:


bopercival-hot (commented):
I don't know if it's too late, sorry it took a couple of days to get to this. But a good idea and maybe we can do a review of our tech contributors code of conduct in the future as we start to build our tech collective.

PR #3972, opened 2020-12-11
Changing Task model include relationship to new TaskDependency model, using Adjacency List pattern

JonLim:
# Changelist

* Added `TaskDependency` class for Task model, describes relationship between two tasks (with indices on both columns!)
* Added `task_dependencies` relationship in Task model


JonLim:
Apologies, I was trying to push a PR to an internal test!

PR #3999, opened 2020-12-17
Allow users to filter task history comments and state changes

### LINKED ISSUE #3977: Allow users to filter comments and state changes on History tab
Add checkboxes to allow users to filter comments and state changes on the history tab:

image

d-rita:
**What does this PR do?** 
Resolves #3977 

**Screenshots:**

- Comments only
Screenshot 2020-12-22 at 03 11 57

- State Changes only
Screenshot 2020-12-22 at 03 11 48

### DIFF
--- a/frontend/src/components/taskSelection/taskActivity.js
+++ b/frontend/src/components/taskSelection/taskActivity.js
@@ -63,9 +63,10 @@ const PostComment = ({ projectId, taskId, setCommentPayload }) => {
 export const TaskHistory = ({ projectId, taskId, commentPayload, mapperLevel }) => {
   const token = useSelector((state) => state.auth.get('token'));
   const [history, setHistory] = useState([]);
-  const [showTaskComments, setShowTaskComments] = useState(false);
+  const taskChangeSetting = mapperLevel === 'BEGINNER' ? false : true;
+  const [showTaskComments, setShowTaskComments] = useState(true);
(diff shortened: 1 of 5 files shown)


frontend/src/components/taskSelection/taskActivity.js
```diff
@@ -59,21 +60,30 @@ const PostComment = ({ projectId, taskId, setCommentPayload }) => {
   );
 };
 
-export const TaskHistory = ({ projectId, taskId, commentPayload }) => {
+export const TaskHistory = ({ projectId, taskId, commentPayload, mapperLevel }) => {
   const token = useSelector((state) => state.auth.get('token'));
   const [history, setHistory] = useState([]);
+  const [showTaskComments, setShowTaskComments] = useState(false);
```
willemarcel, line 66: It would be better to start both `showTaskComments` and `showTaskChanges` as true and make `showTaskChanges` as `false` if the user is begginer, so the user understands what is activated.

willemarcel (changes requested):
@d-rita it's working pretty well, some suggestions on style:

On the task history modal, we need to align it better:
image

Probably setting it with the same alignment as the history items would be good
image

Another suggestion is to replace "Task changes" by "Actions" or "Activities".

willemarcel (approved):

PR #4002, opened 2020-12-18
Fix imagery loading in internal iD and className attrib on mapping sidebar

### LINKED ISSUE #3916: Imagery does not load
Scenario:

_A first time mapper launches a task in a project which has been tweaked to automatically load **Maxar** imagery. This works seamlessly._ 

_This new mapper moves onto a different project. This new project specifies that **Bing** must be used. Unknown to this new mapper the same imagery from the previous project (Maxar) loads. The new mapper does not know how to check that the right imagery has loaded, or how to manually switch imagery. Unwittingly the mapper maps the entire square using incorrect imagery._

**Note:** This only seems to occur in the first instance. If the user selects another task within the new project it will load the correct imagery.

willemarcel:
Resolves #3916 

How to test: lock a task for mapping or validation in a project that has a custom imagery set, then lock in another project, test if both tasks are loaded on iD with the correct imagery. Best if tested in projects with different imagery.

### DIFF
--- a/frontend/src/components/editor.js
+++ b/frontend/src/components/editor.js
@@ -12,23 +12,17 @@ export default function Editor({ setDisable, comment, presets, imageryUrl, gpxUr
   const locale = useSelector((state) => state.preferences.locale);
   const [customImageryIsSet, setCustomImageryIsSet] = useState(false);
   const windowInit = typeof window !== undefined;
+  const customSource =
+    iDContext && iDContext.background() && iDContext.background().findSource('custom');
 
   useEffect(() => {
-    if (
-      !customImageryIsSet &&
-      imageryUrl &&
-      iDContext &&
-      iDContext.background() &&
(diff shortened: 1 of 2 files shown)


ramyaragupathy (commented):
@willemarcel - I have tested this with two projects that has different custom imagery  8886 and 8883. Steps followed:
* From an incognito window signed into the application as admin
* Locked a task from project
* In both the cases, by default Bing loaded in the background
* After unlocking the task from the first time and in the second task lock, custom imagery (each project's custom imagery url loaded perfectly) showed up in the iD editor
I expect the custom imagery to load in the first task lock event as well.

willemarcel:
Thanks, @ramyaragupathy ! I pushed an update.

ramyaragupathy (approved):
Tested with projects having different custom imagery urls - this PR loads the respective project imagery while locking tasks with iD editor.

PR #4010, opened 2020-12-23
Add Docker support for additional frontend parameters

### LINKED ISSUE #4009: Docker build does not support TM_IMPORT_MAX_FILESIZE and TM_MAX_AOI_AREA
`TM_IMPORT_MAX_FILESIZE` and `TM_MAX_AOI_AREA` are two variables which should be set at build time. They are not set in the frontend Dockerfile, nor they are mentioned in `docker-compose.override.yml` (only `TM_APP_API_URL` is mentioned).

frafra:
Fix #4009.

Docker does not support the `env_file` parameter at build time, so the variables have been copied like it has been done with `TM_APP_API_URL` and the defaults too. It is a bit redundant and setting those frontend arguments in `tasking-manager.env` has no effect. This PR does not change how those variables are handled in the Docker frontend, it just adds two more to allow the user to set them.

To avoid such kind of redundancy, it would be enough to rename `tasking-manager.env` as `.env`, but that would be a major change which is out of the scope of this PR.

### DIFF
--- a/docker-compose.override.yml
+++ b/docker-compose.override.yml
@@ -25,6 +25,8 @@ services:
       dockerfile: "./scripts/docker/Dockerfile.frontend"
       args:
         - TM_APP_API_URL=http://localhost/api
+        - TM_IMPORT_MAX_FILESIZE=1000000
+        - TM_MAX_AOI_AREA=5000
     volumes:
       - ".:/usr/src/app"
     labels:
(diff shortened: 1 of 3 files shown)


willemarcel (approved):

willemarcel:
Thanks and sorry for the delay, @frafra

PR #4018, opened 2020-12-28
Avoid error on project list + set action=any on My / Manage projects pages

### LINKED ISSUE #4015: Project GET - unhandled error: {"action": ["Value must be one of ('map', 'validate', 'any')."]}
Sentry Issue: [TASKINGMANAGER-BACKEND-CY](https://sentry.io/organizations/humanitarian-openstreetmap-tea/issues/2094853043/?referrer=github_integration)

```
Project GET - unhandled error: {"action": ["Value must be one of ('map', 'validate', 'any')."]}
```

willemarcel:
Resolves #4015

### DIFF
--- a/backend/models/dtos/project_dto.py
+++ b/backend/models/dtos/project_dto.py
@@ -26,15 +26,6 @@ from backend.models.postgis.statuses import (
 )
 from backend.models.dtos.campaign_dto import CampaignDTO
 
-ORDER_BY_OPTIONS = (
-    "id",
-    "mapper_level",
-    "priority",
-    "status",
-    "last_updated",
-    "due_date",
-)
-
 
 def is_known_project_status(value):
     """ Validates that Project Status is known value """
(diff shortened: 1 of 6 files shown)


ramyaragupathy (commented):
@willemarcel - while the `action` functionality works now as expected while loading `contributed` projects for a mapper - can you please elaborate why would you remove some of the options for query parameters?

willemarcel:
@ramyaragupathy the choices makes the API return a 500 error when an unexpected option is used. A 500 error is not the correct way of handling it.

Let's say I want to filter projects by status if I use `?statuses=abc` and there aren't projects with status `abc`, it should return an empty result.

If I use the `orderBy=abc`, but the options of ordering are only `ASC` and `DESC`, it should just ignore it instead of returning a critical 500 error.

ramyaragupathy (approved):

Issue #3949, opened 2020-12-02
Filter users by username on Project's contributors list


willemarcel:
![image](https://user-images.githubusercontent.com/666291/100815959-e0814500-3423-11eb-9676-424c346b93a1.png)

Add a text input field to allow filter users by username.

wireguy:
How is this enhancement better than just doing a find Ctrl+F on the page?

max-keviv:
@willemarcel you want to sort username lexicographically when input is given
if I'm wrong pls correct me 
somewhat like this may be
![image](https://user-images.githubusercontent.com/59287619/102620273-c5247280-4163-11eb-940e-cde0e87a1977.png)

willemarcel:
> How is this enhancement better than just doing a find Ctrl+F on the page?
@wireguy a CTRL + F is not so user-friendly and if we use it frequently it's better to have on the interface

@max-keviv we want to list only the users whose username contains the text you type on the input.

max-keviv:
@willemarcel  hello
i had implemented the code could you please check if its all right if not please suggest changes

- A another input input is added to choice the username in the given level 
- out of valid contribution username, the username which matches the input text to select from

https://github.com/max-keviv/tasking-manager/blob/18f32ce57a13ad8d378ced6a047db6b45970275e/frontend/src/components/taskSelection/contributions.js#L157-L166

ref-
1)![image](https://user-images.githubusercontent.com/59287619/103153668-37c2cd00-47b8-11eb-9db1-cef5f1efcbab.png)

2)![image](https://user-images.githubusercontent.com/59287619/103153696-62148a80-47b8-11eb-89fa-778e8bd36043.png)

max-keviv:
@willemarcel  it's right i can create a pull request

willemarcel:
That looks good, @max-keviv. Could you open a PR? Please move the level filter dropdown to the right.

PR #4022, opened 2020-12-29
 #3949 Filter users by username on Project's contributors list

max-keviv:
#3949 Filter users by username on Project's contributors  

- A another input input is added to choice the username in the given level

- out of valid contribution username, the username which matches the input text to select from


willemarcel:
@max-keviv it would be better if you open another PR to the user stats endpoint

max-keviv:
@willemarcel  i haven't yet opened PR to the user stats endpoint, i want you to verify it before please

PR #4050, opened 2021-01-05
Add condition for custom URL in cfn template

dakotabenjamin:
In order to allow for non-hotosm urls (such as tasks.teachosm.org) in the cfn template, we add a condition for the url parameter to check if its a hotosm.org domain. We will not create the Route53 record in the case that it is an external url. 

In this case currently it only works when we also control the ssl certificate in ACM, so it is not fully generalized. that should come with future development. 

@willemarcel I would like to test this with `staging-test` before merging.


dakotabenjamin:
Everything works on staging using `staging-test` branch so this is ready for merge

PR #4055, opened 2021-01-06
Forbid to split tasks with zoom level >=18 or area < 25k m²

### LINKED ISSUE #3428: Limit split levels
We allow users to split tasks without any limit. We received suggestions to limit it at some level to avoid unnecessary split actions.

willemarcel:
Resolves #3428


ramyaragupathy (approved):
Per conversation - updated the code:
* to fetch geometry in WKB format
* convert WKB to Geography
* use area function on the new geography

Rest of the changes looks good

PR #4057, opened 2021-01-06
Makes customization of images on header and homepage easier

willemarcel:
Adds two new Environment Variables, which defines the images used on the homepage:
`HOMEPAGE_IMG_HIGH`
`HOMEPAGE_IMG_LOW`

Uses the `TM_ORG_LOGO` to define the organization logo on the header and keep the current SVG as fallback.


d-rita (approved):
LGTM

willemarcel:
@frafra @geohacker I believe this change will interest you

Issue #4011, opened 2020-12-23
Create tasks stats endpoint


willemarcel:
Add tasks stats endpoint to the API. Return the number of tasks with activities on the period.

URL: `/api/v2/tasks/statistics/`

Query params (in priority order to be implemented):
- start date
- end date
- organization
- campaign
- project ids (comma separated values)
- country

Result format:
```
{
  mapped: 0,
  validated: 0,
  invalidated: 0,
  unavailable: 0,
}
```

PR #4061, opened 2021-01-07
Task statistics feature

ramyaragupathy:
From https://github.com/hotosm/tasking-manager/issues/4011 -

How to test the PR?
- Using `http://127.0.0.1:5000/api/v2/tasks/statistics`, get task stats satisfying different query parameters, i.e. start_date, end_date, organisation_id, organisation_name, campaign, project_id, country

### DIFF
--- a/backend/api/tasks/statistics.py
+++ b/backend/api/tasks/statistics.py
@@ -1,5 +1,16 @@
+import datetime
 from flask_restful import Resource, current_app, request
 from backend.services.stats_service import NotFound, StatsService
+
+
+def validate_date_input(date):
+    if date:
+        date_format = "%Y-%m-%d"
+        current_date = datetime.datetime.now()
+        date = datetime.datetime.strptime(date, date_format)
+        if date <= current_date:
+            return date
+        raise Exception("Error: Date out of range for task activity")
 
 
 class TasksStatisticsAPI(Resource):
@@ -25,17 +36,17 @@ class TasksStatisticsAPI(Resource):
               type: string
               default: null
             - in: query
-              name: organisationName
+              name: organisation_name
               description: Organisation name to filter by
               required: false
               default: null
             - in: query
-              name: organisationID
+              name: organisation_id
               description: Organisation ID to filter by
               required: false
               default: null
             - in: query
-              name: campaigns
+              name: campaign
               description: Campaign name to filter by
               required: false
               default: null
@@ -58,16 +69,23 @@ class TasksStatisticsAPI(Resource):
                 description: Internal Server Error
         """
         try:
-            print(request.args)
-            start_date = request.args.get("start_date")
-            end_date = request.args.get("end_date")
-            org_id = request.args.get("org_id")
-            org_name = request.args.get("org_name")
-            campaign = request.args.get("campaigns")
-            project_ids = request.args.get("project_ids")
-            country = request.args.get("country")
+            start_date = validate_date_input(request.args.get("start_date"))
+            end_date = validate_date_input(request.args.get("end_date"))
+            organisation_id = request.args.get("organisation_id", None, int)
+            organisation_name = request.args.get("organisation_name", None, str)
+            campaign = request.args.get("campaign", None, int)
+            project_id = request.args.get("project_id")
+            if project_id:
+                project_id = map(str, project_id.split(","))
+            country = request.args.get("country", None, str)
             task_stats = StatsService.get_task_stats(
-                start_date, end_date, org_id, org_name, campaign, project_ids, country
+                start_date,
+                end_date,
+                organisation_id,
+                organisation_name,
+                campaign,
+                project_id,
+                country,
             )
             return task_stats.to_primitive(), 200
         except NotFound:
--- a/backend/api/tasks/statistics.py
+++ b/backend/api/tasks/statistics.py
@@ -1,0 +1,73 @@
+from flask_restful import Resource, current_app
+from backend.services.stats_service import NotFound, StatsService
+
+
+class TasksStatisticsAPI(Resource):
+    def get(self, project_id):
+        """
+        Get Task Stats
+        ---
+        tags:
+          - tasks
+        produces:
+          - application/json
+        parameters:
+            - in: query
+              name: start_date
+              description: Date to filter as minimum
+              required: false
+              type: string
+              default: null
+            - in: query
+              name: end_date
+              description: Date to filter as maximum
+              required: false
+              type: string
+              default: null
+            - in: query
+              name: organisationName
+              description: Organisation name to filter by
+              required: false
+              default: null
+            - in: query
+              name: organisationID
+              description: Organisation ID to filter by
+              required: false
+              default: null
+            - in: query
+              name: campaigns
+              description: Campaign name to filter by
+              required: false
+              default: null
+            - in: query
+              name: country
+              description: Country name to filter by
+              required: false
+              default: null
+            - in: query
+              name: country
+              description: Country name to filter by
+              required: false
+              default: null
+            - in: query
+              name: project_id
+              description: Project IDs to filter by
+              required: false
+              default: null
+        responses:
+            200:
+                description: Task statistics
+            404:
+                description: Not found
+            500:
+                description: Internal Server Error
+        """
+        try:
+            task_stats = StatsService.get_task_stats()
+            return task_stats.to_primitive(), 200
+        except NotFound:
+            return {"Error": "Not found"}, 404
+        except Exception as e:
+            error_msg = f"Task Statistics GET - unhandled error: {str(e)}"
+            current_app.logger.critical(error_msg)
+            return {"Error": "Unable to fetch task statistics"}, 500
--- a/backend/api/tasks/statistics.py
+++ b/backend/api/tasks/statistics.py
@@ -1,9 +1,9 @@
-from flask_restful import Resource, current_app
+from flask_restful import Resource, current_app, request
 from backend.services.stats_service import NotFound, StatsService
 
 
 class TasksStatisticsAPI(Resource):
-    def get(self, project_id):
+    def get(self):
         """
         Get Task Stats
         ---
@@ -40,18 +40,13 @@ class TasksStatisticsAPI(Resource):
               required: false
               default: null
             - in: query
-              name: country
-              description: Country name to filter by
-              required: false
-              default: null
-            - in: query
-              name: country
-              description: Country name to filter by
-              required: false
-              default: null
-            - in: query
               name: project_id
               description: Project IDs to filter by
+              required: false
+              default: null
+            - in: query
+              name: country
+              description: Country name to filter by
               required: false
               default: null
         responses:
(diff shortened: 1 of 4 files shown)


backend/api/tasks/statistics.py
```diff
@@ -0,0 +1,90 @@
...
+            500:
+                description: Internal Server Error
+        """
+        try:
+            start_date = validate_date_input(request.args.get("start_date"))
+            end_date = validate_date_input(request.args.get("end_date"))
+            org_id = request.args.get("org_id", None, int)
+            org_name = request.args.get("org_name", None, str)
+            campaign = request.args.get("campaign", None, int)
+            project_ids = request.args.get("project_ids")
```
ramyaragupathy, line 77: Please use the write query parameter for project ID. `project_ids` -> `project_id`

ramyaragupathy (commented):
From  initial review - Can you please clean up the variable naming for query parameters in YAML specs and the subsequent `request.args` within the controller function?

Still reviewing the function in stats service for better ordering and query optimisation

ramyaragupathy (commented):
While renaming please make sure the YAML names obey the Python naming convention - `snake_case`

backend/api/tasks/statistics.py
ramyaragupathy: All the argument names should match the YAML query parameter names, without which values would be `None`

backend/services/stats_service.py
```diff
@@ -477,3 +479,92 @@ def update_project_stats(project_id: int):
...
+            # Unnest country column array.
+            sq = Project.query.with_entities(
+                Project.id, func.unnest(Project.country).label("country")
+            ).subquery()
+
+            query = query.filter(sq.c.country.ilike("%{}%".format(country))).filter(
+                TaskHistory.project_id == sq.c.id
+            )
+
+        for i in query:
```
willemarcel, line 553: Instead of iterating over the query, you can do `query.filter(Task.task_status == TaskStatus.MAPPED.value).count()`

backend/services/stats_service.py
```diff
@@ -477,3 +479,87 @@ def update_project_stats(project_id: int):
...
+                Project.id, func.unnest(Project.country).label("country")
+            ).subquery()
+
+            query = query.filter(sq.c.country.ilike("%{}%".format(country))).filter(
+                TaskHistory.project_id == sq.c.id
+            )
+
+        stats_dto = TaskStatsDTO()
+        stats_dto.mapped = query.filter(
+            Task.task_status.in_((TaskStatus.MAPPED.value, TaskStatus.VALIDATED.value))
```
willemarcel, line 553: That line should be `Task.task_status == TaskStatus.MAPPED.value`. It's better to return the raw values and sum it on the frontend if needed.

backend/services/stats_service.py
```diff
@@ -477,3 +479,87 @@ def update_project_stats(project_id: int):
...
+        ).count()
+        stats_dto.validated = query.filter(
+            Task.task_status == TaskStatus.VALIDATED.value
+        ).count()
+        stats_dto.invalidated = query.filter(
+            Task.task_status == TaskStatus.INVALIDATED.value
+        ).count()
+        stats_dto.unavailable = query.filter(
```
willemarcel, line 561: I used `badimagery` as key name on #4097, as it's the term we use on backend

willemarcel (changes requested):
some small changes needed

backend/api/tasks/statistics.py
```diff
@@ -0,0 +1,110 @@
...
+            - in: query
+              name: start_date
+              description: Date to filter as minimum
+              required: true
+              type: string
+              default: null
+            - in: query
+              name: end_date
+              description: Date to filter as maximum
+              required: true
```
willemarcel, line 35: Are you planning to restrict the maximum date range we can query? Like 1 or 2 years? I think that param doesn't need to be required in case the current date is inside that range. It would make the frontend interface behaviour better to the users.

backend/services/stats_service.py
```diff
@@ -479,3 +482,122 @@ def update_project_stats(project_id: int):
...
+
+        dates_stats = []
+        task_tracker = {
+            "MAPPED": [],
+            "VALIDATED": [],
+            "INVALIDATED": [],
+            "BAD IMAGERY": [],
+        }
+
+        # r -> (58, 3, 'MAPPED', datetime.date(2021, 1, 5)):
```
ramyaragupathy, line 557: Good though to include sample data in the comment

backend/models/dtos/stats_dto.py
```diff
@@ -141,3 +141,19 @@ def __init__(self):
...
+    mapped = IntType(serialized_name="mapped")
+    validated = IntType(serialized_name="validated")
+    invalidated = IntType(serialized_name="invalidated")
+    bad_imagery = IntType(serialized_name="badImagery")
+
+
+class AllTaskStatsDTO(Model):
+    """ Contains all tasks stats broken down by day"""
+
+    stats = ListType(ModelType(TaskStatsDTO))
```
ramyaragupathy, line 159: An improvement we could do - initializing `stats` variable in constructor, similar to other DTO classes. That way we ensure objects are initialized properly. From the execution perspective, either ways will work, and it's a matter of style, but to maintain uniformity in the code, I suggest we user constructors for initialization.

backend/models/dtos/stats_dto.py
```diff
@@ -141,3 +141,19 @@ def __init__(self):
...
+    """ DTO for tasks stats for a single day """
+
+    date = DateType(required=True)
+    mapped = IntType(serialized_name="mapped")
+    validated = IntType(serialized_name="validated")
+    invalidated = IntType(serialized_name="invalidated")
+    bad_imagery = IntType(serialized_name="badImagery")
+
+
+class AllTaskStatsDTO(Model):
```
ramyaragupathy, line 156: Let's have this class as TaskStatsDTO - and the individual date DTO named as `TaskStats' - some way to know that these two are related and one of this is the building block for another.

For example we have: 
* ProjectComment, ProjectCommentsDTO
* ProjectContribDTO, ProjectContribsDTO
* UserContribution, ProjectContributionsDTO

backend/services/stats_service.py
```diff
@@ -479,3 +482,122 @@ def update_project_stats(project_id: int):
...
+
+            values = [(r[0], r[1], r[2]) for r in query if d == r[3]]
+            values.sort(reverse=True)
+
+            for value in values:
+                task_detail = (value[0], value[1])  # tuple for (task_id, project_id)
+                task_status = value[2]
+
+                if task_status == "MAPPED":
+                    if task_detail not in task_tracker["MAPPED"]:
```
ramyaragupathy, line 581: A mapped task could be validated or invalidated. Also a validated task could be reset to `mapped` state through `Request revalidation` action. If a task has been validated and then it was reset to `mapped` state, then it should be counted only for validation. So we should include an additional check here. Considering the count scenarios, I feel we should have timestamp as a separate field and go through the list in order of the timestamp

backend/services/stats_service.py
```diff
@@ -479,3 +482,122 @@ def update_project_stats(project_id: int):
...
+                        task_tracker["MAPPED"].append(task_detail)
+                        day_stats_dto.mapped += 1
+                elif task_status == "VALIDATED":
+                    if task_detail not in task_tracker["VALIDATED"]:
+                        task_tracker["VALIDATED"].append(task_detail)
+                        day_stats_dto.validated += 1
+                        if task_detail in task_tracker["INVALIDATED"]:
+                            task_tracker["INVALIDATED"].remove(task_detail)
```
ramyaragupathy, line 589: Again here - if we don't consider the timestamp, then the task is counted for both validation and invalidation.

backend/services/stats_service.py
```diff
@@ -479,3 +482,114 @@ def update_project_stats(project_id: int):
...
+            query = query.filter(sq.c.country.ilike("%{}%".format(country))).filter(
+                TaskHistory.project_id == sq.c.id
+            )
+
+        dates_stats = []
+        task_tracker = {
+            "MAPPED": [],
+            "VALIDATED": [],
+            # "INVALIDATED": [],
+            "BAD IMAGERY": [],
```
willemarcel, line 553: It's better to avoid spaces in the key

backend/services/stats_service.py
```diff
@@ -479,3 +482,114 @@ def update_project_stats(project_id: int):
...
+                if task_status == "MAPPED":
+                    if task_detail not in task_tracker["MAPPED"]:
+                        task_tracker["MAPPED"].append(task_detail)
+                        day_stats_dto.mapped += 1
+                elif task_status == "VALIDATED":
+                    if task_detail not in task_tracker["VALIDATED"]:
+                        task_tracker["VALIDATED"].append(task_detail)
+                        day_stats_dto.validated += 1
+                else:
+                    if task_detail not in task_tracker["BAD IMAGERY"]:
```
willemarcel, line 587: the bad imagery results are not correct for me

backend/services/stats_service.py
willemarcel: I tested requesting revalidation in one task, and it worked well. Only the first mapped date was considered.

backend/services/stats_service.py
```diff
@@ -479,3 +482,114 @@ def update_project_stats(project_id: int):
...
+        )
+
+        if org_id:
+            query = query.join(Project, Project.id == TaskHistory.project_id).filter(
+                Project.organisation_id == org_id
+            )
+        if org_name:
+            organisation = OrganisationService.get_organisation_by_name(org_name)
+            query = query.join(Project, Project.id == TaskHistory.project_id).filter(
+                Project.organisation_id == organisation.id
```
willemarcel, line 529: We need to handle better the case when there is not an organisation with that name. I suggest:

```
            try:
                organisation_id = OrganisationService.get_organisation_by_name(org_name).id
            except NotFound:
                organisation_id = None
            query = query.join(Project, Project.id == TaskHistory.project_id).filter(
                Project.organisation_id == organisation_id
            )
```

willemarcel (changes requested):

backend/services/stats_service.py
```diff
@@ -479,3 +482,114 @@ def update_project_stats(project_id: int):
...
+        query = (
+            db.session.query(
+                TaskHistory.task_id,
+                TaskHistory.project_id,
+                TaskHistory.action_text,
+                func.DATE(TaskHistory.action_date).label("day"),
+            )
+            .filter(
+                TaskHistory.action == "STATE_CHANGE",
+                or_(
```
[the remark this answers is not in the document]
ramyaragupathy: With `and_`: the query is as follows:
```
SELECT 
task_history.task_id AS task_history_task_id, 
task_history.project_id AS task_history_project_id,
task_history.action_text AS task_history_action_text, 
DATE(task_history.action_date) AS day
FROM task_history
WHERE task_history.action = 'STATE_CHANGE' AND 
task_history.action_text = 'MAPPED' AND 
task_history.action_text = 'VALIDATED' AND 
task_history.action_text = 'BAD_IMAGERY' AND 
DATE(task_history.action_date) >= '2021-01-01' AND
DATE(task_history.action_date) <= '2021-02-01' 
GROUP BY task_history.action_text, day, task_history.project_id, task_history.task_id ORDER BY day
```

With `or_` query is framed as:
```
SELECT task_history.task_id AS task_history_task_id,
task_history.project_id AS task_history_project_id, 
task_history.action_text AS task_history_action_text,
DATE(task_history.action_date) AS day
FROM task_history
WHERE task_history.action = 'STATE_CHANGE' AND (
	task_history.action_text = 'MAPPED' OR 
	task_history.action_text = 'VALIDATED' OR 
	task_history.action_text = 'BAD_IMAGERY') AND 

DATE(task_history.action_date) >= '2021-01-01' AND
DATE(task_history.action_date) <= '2021-02-01' 
GROUP BY task_history.action_text, day, task_history.project_id, task_history.task_id ORDER BY day
```

So a `OR` filter picks rows with one of those values whereas `AND` filter looks for entries with all the specified values.

backend/services/stats_service.py
ramyaragupathy: Can you share more details? what was missing?

backend/services/stats_service.py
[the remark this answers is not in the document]
willemarcel: @ramyaragupathy that is right! I read the code wrongly.

willemarcel (approved):

PR #4067, opened 2021-01-07
Teachosm circleci

dakotabenjamin:
merging to `teachosm-style` because i'm using that branch as an upstream for deployment to `deployment/teachosm-tasking-manager`.


willemarcel (approved):
looks good!

Issue #4012, opened 2020-12-23
Create users stats endpoint


willemarcel:
Add user stats endpoint to the API. Return the number of users registered in a period of time.

URL: `/api/v2/users/statistics/`

Query params:
- start date
- end date

`Note: query by the user registration date`

Result format:
```
{
  total: 0, 
  beginner: 0,
  intermediate: 0,
  advanced: 0,
  tasks_mapped: 0, 
  tasks_validated: 0,
}
```

**Notes:**
- `beginner, intermediate and advanced` correspond to the current level of the users registered on the date query period.
- `tasks_mapped and tasks_validated` correspond to the total number of tasks mapped or validated by users registered on the date query period.

max-keviv:
@willemarcel can you tell me reasons behind Unauthorized Access response in api

max-keviv:
@willemarcel  i have implemented users statistics Api 
:-Return the number of users registered in a period of time.
![Screenshot from 2021-01-07 20-33-44](https://user-images.githubusercontent.com/59287619/103913065-22419180-512e-11eb-83c4-140fe2b0e146.png)

![Screenshot from 2021-01-07 20-33-30](https://user-images.githubusercontent.com/59287619/103913037-1524a280-512e-11eb-90f6-89e1912c72a8.png)


![Screenshot from 2021-01-07 20-31-56](https://user-images.githubusercontent.com/59287619/103921045-03e09380-5138-11eb-8e4d-c7403eca5902.png)


can you please check and verify it,

#https://github.com/max-keviv/tasking-manager/commit/bce663c031b8d682d467dc65c02b9cbbbb587742

PR #4080, opened 2021-01-10
Create users stats endpoint #4012

max-keviv:
Create users stats endpoint #4012

Add user stats endpoint to the API. Return the number of users registered in a period of time.


max-keviv:
@willemarcel  i have opened another pull request for this issue. please give a review #4012

PR #4081, opened 2021-01-10
Pointing out that email won't send without a from address

iandees:
Noticed that without setting this variable the system won't send email, so I made its status a bit more clear.


willemarcel (approved):

willemarcel:
Thanks, @iandees!

Issue #4128, opened 2021-01-20
Frontend should not suggest to validate tasks already validated


willemarcel:
When selecting tasks from the contributions tab, the frontend is suggesting to validate multiple tasks. It should lock for validation only the ones that really need it.

PR #4133, opened 2021-01-22
Exclude tasks that are not on mapped status, when locking tasks of a user

willemarcel:
Resolves https://github.com/hotosm/tasking-manager/issues/4128

**How to test:**

- **Case 1:** Select the tasks mapped by a user:
Screen Shot 2021-01-22 at 10 05 22

Although it shows "Validate 16 selected tasks" on the button, with that PR we are excluding the tasks that are not on the mapped status and lock only the mapped ones.

- **Case 2:** If the selected user has not tasks on mapped status, it will show an error message
Screen Shot 2021-01-22 at 10 07 27


ramyaragupathy (approved):
image
 Works as expected - though I could see the option to validate, later I see the message that no task is available for validation

PR #4141, opened 2021-01-24
JIRA issue: [HTR-70]; Github Issue: #27 in the tech repository

eternaltyro:
This patch adds security headers to fix the Mozilla Observatory reports
for missing security headers. I have added the following headers in their
recommended settings using HTTP meta tags.

X-Content-Type-Options
X-XSS-Protection
X-Frame-Options
Strict-Transport-Security
Content-Security-Policy

For Content-Security-Policy, I have used Mozilla Labs - CSP generator
to scan the site and generate a base-policy that I then tweaked with
instructions from Mapbox, Google and Matomo documentations.

Also used csper.io Chrome extension to generate the CSP policies. Need
to verify and make the policy exhaustive.

SECURITY NOTE:

HSTS header may not work when passed in a HTTP file. Might need to tweak
the CDN to pass a HSTS header.

### DIFF
--- a/frontend/public/index.html
+++ b/frontend/public/index.html
@@ -5,6 +5,26 @@
     <link rel="shortcut icon" href="%PUBLIC_URL%/static/favicon.ico" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
 
+    <!-- for security -->
+    <meta http-equiv="Content-Security-Policy" 
+          content="default-src 'self';
+                   connect-src 'self' https://api.mapbox.com https://cdn.jsdelivr.net https://dev.virtualearth.net https://events.mapbox.com https://osm-stats-production-api.azurewebsites.net https://osmstats-api.hotosm.org https://taginfo.openstreetmap.org https://tasking-manager-tm4-production-api.hotosm.org https://www.openstreetmap.org; 
+                   font-src https://fonts.gstatic.com; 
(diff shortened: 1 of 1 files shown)


eternaltyro:
Reported issue: hotosm/tech#127

frontend/public/index.html
```diff
@@ -5,6 +5,26 @@
     <link rel="shortcut icon" href="%PUBLIC_URL%/static/favicon.ico" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
 
+    <!-- for security -->
+    <meta http-equiv="Content-Security-Policy" 
+          content="default-src 'self';
+                   connect-src 'self' https://api.mapbox.com https://cdn.jsdelivr.net https://dev.virtualearth.net https://events.mapbox.com https://osm-stats-production-api.azurewebsites.net https://osmstats-api.hotosm.org https://taginfo.openstreetmap.org https://tasking-manager-tm4-production-api.hotosm.org https://www.openstreetmap.org;
```
willemarcel, line 11: I think it's quite difficult to maintain that list of urls, as we can use new APIs, and other instances will use others too. I observed that even other Mozilla websites don't have it, like getpocket.com, getfirefox.com, developer.mozilla.org

eternaltyro: These URL are not arbitrary, we simply list the connections that are made from within tasking-manager. Our code must have this list. Apart from this, the various maps that OSM uses make a small list.

PR #4166, opened 2021-02-01
Avoid ZeroDivisionError in the percent mapped / validated / bad imagery

### LINKED ISSUE #4165: Project GET - unhandled error: division by zero
Sentry Issue: [TASKINGMANAGER-BACKEND-E](https://sentry.io/organizations/humanitarian-openstreetmap-tea/issues/1941054890/?referrer=github_integration)

```
Project GET - unhandled error: division by zero
```

The error is happening on `https://tasking-manager-tm4-production-api.hotosm.org/api/v2/projects/?page=262&omitMapResults=true&projectStatuses=ARCHIVED%2CDRAFT&mapperLevel=ALL&orderBy=last_updated&orderByType=ASC`

willemarcel:
Resolves #4165

### DIFF
--- a/backend/models/postgis/project.py
+++ b/backend/models/postgis/project.py
@@ -1139,16 +1139,19 @@ class Project(db.Model):
         target, total_tasks, tasks_mapped, tasks_validated, tasks_bad_imagery
     ):
         """ Calculates percentages of contributions """
-        if target == "mapped":
-            return int(
-                (tasks_mapped + tasks_validated)
-                / (total_tasks - tasks_bad_imagery)
-                * 100
-            )
-        elif target == "validated":
-            return int(tasks_validated / (total_tasks - tasks_bad_imagery) * 100)
-        elif target == "bad_imagery":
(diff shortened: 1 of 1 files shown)


d-rita (approved):

backend/models/postgis/project.py
```diff
@@ -1139,16 +1139,19 @@ def calculate_tasks_percent(
...
-            return int(
-                (tasks_mapped + tasks_validated)
-                / (total_tasks - tasks_bad_imagery)
-                * 100
-            )
-        elif target == "validated":
-            return int(tasks_validated / (total_tasks - tasks_bad_imagery) * 100)
-        elif target == "bad_imagery":
-            return int((tasks_bad_imagery / total_tasks) * 100)
+        try:
```
maxgrossman, line 1142: could you just check if total_tasks == tasks_bad_imagery instead of using an exception for control flow?
I also think that we need to catch this case in the [project explore result](https://github.com/DigitalGlobe/tasking-manager/blob/master/backend/services/project_search_service.py#L101-L102)

PR #4188, opened 2021-02-03
Create users stats endpoint

max-keviv:
Create users stats endpoint #4012

Add user stats endpoint to the API. Return the number of users registered in a period of time.

@willemarcel  please have a review of it


willemarcel:
@max-keviv thanks a lot. I merged it through https://github.com/hotosm/tasking-manager/pull/4202

PR #4193, opened 2021-02-04
show error modal if webgl not supported.

### LINKED ISSUE #4119: Check if browser has WebGL support
If a browser doesn't have permissions to use WebGL, the pages with maps will fail to load. We could check it and show a message to the users in case WebGL is not active.

Abhirup-99:
Fixes #4119

### DIFF
--- a/frontend/src/App.js
+++ b/frontend/src/App.js
@@ -8,6 +8,7 @@ import * as Sentry from '@sentry/react';
 import './assets/styles/index.scss';
 import { ORG_NAME, MATOMO_ID } from './config';
 import { Header } from './components/header';
+import { WebglErrorModal } from './components/webglErrorModal';
 import { Footer } from './components/footer';
 import { Preloader } from './components/preloader';
 import { Home } from './views/home';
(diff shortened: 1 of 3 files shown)


Abhirup-99:
@willemarcel please review.

willemarcel (changes requested):
It's not rendering the header for me (in a supported browser). I discovered a lib that makes a more robust check https://github.com/mapbox/mapbox-gl-supported

I think we could use it and try to render an error message instead of the map in the pages with maps.

Abhirup-99:
> It's not rendering the header for me (in a supported browser). I discovered a lib that makes a more robust check https://github.com/mapbox/mapbox-gl-supported
> 
> I think we could use it and try to render an error message instead of the map in the pages with maps.

The modal box would show only in an unsupported browser. Would you explain the header part, I didn't get it.

willemarcel:
@Abhirup-99  The tasking manager header, with the login button, the navigation menu, etc is not being rendered

HelNershingThapa:
Closed with https://github.com/hotosm/tasking-manager/pull/5200 being merged in

Issue #4200, opened 2021-02-04
Add "exempt" field to the GET /organisation/{id} endpoint


willemarcel:
Depends on #4194

Issue #4194, opened 2021-02-04
Add "exempt' field to the organisation model


willemarcel:
Add "exempt' field to the organisation model to express if an organisation is under the TM sustainability model or not. This field should not be available to update through the API.

Field should be `boolean` and default value = `True`

PR #4218, opened 2021-02-09
Add type field to organisation model

d-rita:
**What does this PR do?** 
- Add a type field to reflect an organisation's subscription type, and return it in the response. 

To test the PR:
- run `python manage.py db upgrade`
- go to `GET /organisations/{id}` to view the response. 

Related issues:
- #4194
- #4200


backend/models/dtos/organisation_dto.py
```diff
@@ -40,6 +40,7 @@ class OrganisationDTO(Model):
     projects = ListType(StringType, serialize_when_none=False)
     teams = ListType(ModelType(OrganisationTeamsDTO))
     campaigns = ListType(ListType(StringType))
+    exempt = BooleanType()
```
willemarcel, line 43: let's use type here too

backend/models/postgis/organisation.py
```diff
@@ -165,6 +167,7 @@ def as_dto(self, omit_managers=False):
         organisation_dto.description = self.description
         organisation_dto.url = self.url
         organisation_dto.managers = []
+        organisation_dto.exempt = self.exempt
 
         if omit_managers:
```
willemarcel, line 172: here too

migrations/versions/86c0f6b6a176_.py
```diff
@@ -0,0 +1,52 @@
...
+branch_labels = None
+depends_on = None
+
+
+def upgrade():
+    # ### commands auto generated by Alembic - please adjust! ###
+    op.add_column("organisations", sa.Column("type", sa.Integer(), nullable=True))
+    op.execute("UPDATE organisations SET type = {}".format(OrganisationType.FREE.value))
+    op.alter_column("organisations", "type", nullable=False)
+    op.drop_index(
```
willemarcel, line 26: Why do we have those drop / create index commands?

willemarcel (changes requested):

migrations/versions/86c0f6b6a176_.py
d-rita: They were auto-generated when the migration file was created. I figured the index was being dropped and recreated because of the change in the model/table. Should they be removed?

willemarcel: I don't know. I don't see that behaviour on Django Migrations.

willemarcel (approved):

PR #4226, opened 2021-02-12
Fix contributor counting.

### LINKED ISSUE #3882: Wrong filter logic on get_total_contributors of a project
We have two different functions to get the number of total contributors of a project.

That one queries the information for an individual project and it's correct: https://github.com/hotosm/tasking-manager/blob/develop/backend/models/postgis/project.py#L929

That one is counting the number of task history entries without excluding the comments and is wrong: https://github.com/hotosm/tasking-manager/blob/develop/backend/services/project_search_service.py#L153

teymour-aldridge:
Fixes #3882


willemarcel (approved):
Thanks, @teymour-aldridge !

PR #4231, opened 2021-02-14
Make `calculate_tasks_percent` into a classmethod.

### LINKED ISSUE #3841: Make Project.calculate_tasks_percent a class method
It would be easier to deal with that function if it was a class method instead of a static method:
https://github.com/hotosm/tasking-manager/blob/develop/backend/models/postgis/project.py#L1137

teymour-aldridge:
Fixes #3841

### DIFF
--- a/backend/models/postgis/project.py
+++ b/backend/models/postgis/project.py
@@ -1134,9 +1134,9 @@ class Project(db.Model):
         tags_dto.tags = [r[0] for r in query]
         return tags_dto
 
-    @staticmethod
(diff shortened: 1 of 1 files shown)


willemarcel (changes requested):
The idea of turning it on a class method was to avoid having to pass too many arguments to the function. So we could do something like
```
p = Project.get(id=1)
p.calculate_tasks_percent()
```

It will require changes in other parts of the code that uses that method.

PR #4237, opened 2021-02-16
Modify query to return counts

ramyaragupathy:


willemarcel (changes requested):
Some dates appear more than one time and it makes the frontend chart to fail:

image

willemarcel:
It's not solving the issue of considering only the first mapped and validated date.

willemarcel (approved):

PR #4252, opened 2021-02-18
Reset arbitrary tasks option when clicking the Reset button

willemarcel:
When we imported a file and enabled the "Arbitrary tasks" option, it was keeping that option active even if we reset the AoI and draw a new one. That PR fixes it by resetting the "Arbitrary tasks" option when clicking on reset


d-rita (approved):
LGTM!

PR #4254, opened 2021-02-18
Discard tiny tasks when creating projects

### LINKED ISSUE #4118: Discard tiny tasks when generating task grid
On the project creation process, it's common to have some super tiny tasks, due to the grid generation process. We should discard tasks that are smaller than a threshold. I believe something around 1000 m² would be good, but we can do a small research to find the better number.

willemarcel:
Resolves #4118

Missing:
- [ ] update the AoI geometry to reflect the new area of the project without the discarded tasks


willemarcel:
Replaced by https://github.com/hotosm/tasking-manager/pull/4426

PR #4255, opened 2021-02-19
Show discounted cost for organisations under that type

### LINKED ISSUE #4244: Show discounted cost if organization has that right
- [x] Show the discounted cost to organizations that have the right to a discounted fee, instead of the full fee.
- [ ] Add a "discounted" label, so they know the value has a reduction.

d-rita:
Resolves #4244

### DIFF
--- a/frontend/src/components/teamsAndOrgs/messages.js
+++ b/frontend/src/components/teamsAndOrgs/messages.js
@@ -237,10 +237,6 @@ export default defineMessages({
     id: 'management.organisations.stats.cost.estimation',
     defaultMessage: 'Estimated cost by the end of {year}',
   },
-  discountedLabel: {
-    id: 'management.organisations.stats.cost.estimation.discounted',
-    defaultMessage: 'discounted',
-  },
   actionsToNextLevel: {
     id: 'management.organisations.stats.next_level.actions',
     defaultMessage: 'Actions to reach the level {n}',
--- a/frontend/src/components/teamsAndOrgs/messages.js
+++ b/frontend/src/components/teamsAndOrgs/messages.js
@@ -237,6 +237,10 @@ export default defineMessages({
     id: 'management.organisations.stats.cost.estimation',
     defaultMessage: 'Estimated cost by the end of {year}',
   },
+  discountedLabel: {
+    id: 'management.organisations.stats.cost.estimation.discounted',
+    defaultMessage: 'discounted',
+  },
   actionsToNextLevel: {
     id: 'management.organisations.stats.next_level.actions',
     defaultMessage: 'Actions to reach the level {n}',
--- a/frontend/src/components/teamsAndOrgs/orgUsageLevel.js
+++ b/frontend/src/components/teamsAndOrgs/orgUsageLevel.js
@@ -6,15 +6,16 @@ import messages from './messages';
 import { StatsCardContent } from '../statsCardContent';
 import { ProgressBar } from '../progressBar';
 import { usePredictYearlyTasks } from '../../hooks/UsePredictYearlyTasks';
-import { useOrganisationLevel, useGetLevel } from '../../hooks/UseOrganisationLevel';
+import { useOrganisationLevel, usePredictLevel } from '../../hooks/UseOrganisationLevel';
 
 export function OrganisationUsageLevel({ completedActions, orgName, type, userIsOrgManager }) {
   const [currentLevel, nextLevelThreshold] = useOrganisationLevel(completedActions);
   const percent = parseInt((completedActions / nextLevelThreshold) * 100);
   const yearPrediction = usePredictYearlyTasks(completedActions, new Date());
-  const levelPrediction = useGetLevel(yearPrediction);
+  const levelPrediction = usePredictLevel(yearPrediction, type);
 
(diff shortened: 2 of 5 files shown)


frontend/src/components/teamsAndOrgs/messages.js
```diff
@@ -237,6 +237,14 @@ export default defineMessages({
     id: 'management.organisations.stats.cost.estimation',
     defaultMessage: 'Estimated cost by the end of {year}',
   },
+  estimatedTierCostLabel: {
+    id: 'management.organisations.stats.cost.estimation.label',
+    defaultMessage: '{cost} {discountLabel}',
```
willemarcel, line 242: It's better to handle it on the code instead of generating a string for translation that doesn't need translation.

frontend/src/components/teamsAndOrgs/orgUsageLevel.js
```diff
@@ -6,15 +6,17 @@ import messages from './messages';
...
 export function OrganisationUsageLevel({ completedActions, orgName, type, userIsOrgManager }) {
   const [currentLevel, nextLevelThreshold] = useOrganisationLevel(completedActions);
   const percent = parseInt((completedActions / nextLevelThreshold) * 100);
   const yearPrediction = usePredictYearlyTasks(completedActions, new Date());
-  const levelPrediction = useGetLevel(yearPrediction);
+  const levelPrediction = usePredictLevel(yearPrediction, type);
 
   const showTierInfo = ['DISCOUNTED', 'FULL_FEE'].includes(type) && userIsOrgManager;
+  const showDiscountLabel =
+    JSON.stringify(currentLevel) !== JSON.stringify(levelPrediction) && type === 'DISCOUNTED';
```
willemarcel, line 19: What is the goal of that comparison between `currentLevel` and `levelPrediction`?

frontend/src/components/teamsAndOrgs/orgUsageLevel.js
```diff
@@ -94,8 +96,24 @@ export function OrganisationUsageLevel({ completedActions, orgName, type, userIs
...
+                        cost: (
+                          <FormattedNumber
+                            value={levelPrediction.fee}
+                            // eslint-disable-next-line
+                            style="currency"
+                            currency="USD"
+                          />
+                        ),
+                        discountLabel: showDiscountLabel ? (
+                          <FormattedMessage {...messages.discountedLabel} />
```
willemarcel, line 111: I prefer to add `<FormattedMessage {...messages.discountedLabel} />` under a `<span className="f4">`

willemarcel (changes requested):

frontend/src/components/teamsAndOrgs/orgUsageLevel.js
d-rita: If an organisation's tier did not change from its current level. Although, I should have tested it for different use cases. 🤔My only consideration was a free organisation whose predicted actions(<1000) would still maintain it at the same tier and so its cost would not really be discounted just maintained at 0.

willemarcel: I think it would be good if we don't show the discounted tag if the predicted tier is FREE, so that should be the check.

d-rita: Alright

frontend/src/components/teamsAndOrgs/messages.js
```diff
@@ -237,6 +237,10 @@ export default defineMessages({
     id: 'management.organisations.stats.cost.estimation',
     defaultMessage: 'Estimated cost by the end of {year}',
   },
+  discountedLabel: {
+    id: 'management.organisations.stats.cost.estimation.discounted',
+    defaultMessage: ' (discounted)',
```
willemarcel, line 242: I don't like to include trailing spaces in the strings, because if the translator doesn't include it, we will have problems, maybe we should remove the parenthesis too.
On the component you can use `{' '}` or a padding with CSS.

frontend/src/components/teamsAndOrgs/orgUsageLevel.js
```diff
@@ -6,15 +6,16 @@ import messages from './messages';
...
 
 export function OrganisationUsageLevel({ completedActions, orgName, type, userIsOrgManager }) {
   const [currentLevel, nextLevelThreshold] = useOrganisationLevel(completedActions);
   const percent = parseInt((completedActions / nextLevelThreshold) * 100);
   const yearPrediction = usePredictYearlyTasks(completedActions, new Date());
-  const levelPrediction = useGetLevel(yearPrediction);
+  const levelPrediction = usePredictLevel(yearPrediction, type);
 
   const showTierInfo = ['DISCOUNTED', 'FULL_FEE'].includes(type) && userIsOrgManager;
+  const showDiscountLabel = levelPrediction.tier !== 'free' && type === 'DISCOUNTED';
```
willemarcel, line 18: Write a test to `OrganisationUsageLevel` to confirm that condition and to confirm we are showing the discounted value.

willemarcel (changes requested):

willemarcel (approved):

PR #4256, opened 2021-02-19
Rename tasksNo to tasksNumber and update it when going back to Step 2

willemarcel:
After trimming the tasks grid on the Step 3 of project creation and going back to step 2, the number of tasks were not being updated. This PR fixes it.


d-rita (approved):
Works nicely. LGTM!

Issue #2168, opened 2020-01-10
Cursor icon update while saving information


ramyaragupathy:
While choosing to save information from the profile page, cursor icon should change to buys icon while the backend changes are made.

![save-option](https://user-images.githubusercontent.com/12103383/72184217-ee963380-3415-11ea-8b32-ae887411944d.gif)

willemarcel:
I would prefer to disable the button and then show a notification that the change was done successfully. I think the mouse cursor doesn't communicate too much.

willemarcel:
Planned solution: disable the button and show a loading spinner, like that https://ant.design/components/button/#components-button-demo-loading

PR #4295, opened 2021-02-26
Add loading icon and use it on Preloader component

willemarcel:
An improvement that will contribute to https://github.com/hotosm/tasking-manager/issues/2168, but doesn't solve it yet.


d-rita (commented):
Works nicely. Curious as to whether we can modify the Preloader's height and width properties?

d-rita (approved):

PR #4301, opened 2021-02-28
Test projectDetail components

d-rita:
This PR tests the following components that make up the `projectDetail` view:
- download buttons: AOI and TaskGrid
- footer
- favorites
- bigProjectTeaser
- shareButton
- QuestionsAndComments
- infoPanel
- statusBox
- header


frontend/src/components/projectDetail/tests/favorites.test.js
```diff
@@ -0,0 +1,24 @@
...
+    const { container } = render(
+      <ReduxIntlProviders>
+        <AddToFavorites {...props} />
+      </ReduxIntlProviders>,
+    );
+    const button = screen.getByRole('button');
+    expect(button.className).toBe(' input-reset base-font bg-white blue-dark f6 bn pointer');
+    expect(button.className).not.toBe('dn input-reset base-font bg-white blue-dark f6 bn pointer');
+    expect(container.querySelector('svg').classList.value).toBe('pt3 pr2 v-btm o-50 ');
+    expect(button.textContent).toBe('Add to Favorites');
```
willemarcel, line 22: We could mock the API request and test the click on the button.

willemarcel: @d-rita in case you have not started working on that change, I think we can dedicate time to other components instead of this one, as this functionality doesn't seem to be very widely used.

frontend/src/components/projectDetail/questionsAndComments.js
```diff
@@ -66,19 +70,23 @@ export const QuestionsAndComments = ({ projectId }) => {
...
+      fetchLocalJSONAPI(`projects/${projectId}/comments/?perPage=5&page=${page}`, token).then(
+        (res) => {
+          if (isMountedRef.current) {
+            setComments(res);
+          }
+        },
+      );
     }
-  }, [page, projectId, token]);
+  }, [page, projectId, token, isMountedRef]);
```
willemarcel, line 89: on that case, wouldn't be better to trigger the request only when `isMountedRef` is `true`? I didn't test on my browser, but maybe it would avoid duplicated requests

d-rita: I'll test it out and see.

willemarcel (approved):

PR #4337, opened 2021-03-04
Display "no members" message if team is empty

### LINKED ISSUE #4330: Improve UserAvatarList component
Show message when the list of users is empty on `UserAvatarList`: https://github.com/hotosm/tasking-manager/blob/develop/frontend/src/components/user/avatar.js#L97

d-rita:
Fixes #4330


willemarcel (changes requested):
image
I think showing the message inside the team cards makes it a bit dirty and I remembered that it's used on other parts of TM. So I think you could undo the change on the UserAvatarList and add another message to the Managers section of the team membership page with "There are not managers yet"

PR #4348, opened 2021-03-08
List all orgs to ADMIN users + add msg when orgs list is empty

willemarcel:
- Show message when the organisation list is empty
- Enable the "All" orgs option by default to admin users


d-rita (approved):
LGTM!

PR #4351, opened 2021-03-08
Fix privacy issue on user detail API + add privacy notice

willemarcel:
Expose gender information only to the own user + add privacy note about it


ramyaragupathy (commented):
image
 Made some backend edits

PR #4355, opened 2021-03-09
Filter teams by organisation

### LINKED ISSUE #3028: Support filtering teams by organisations
After #3027 is fixed, add support to organisation filters on `ListTeams` component: https://github.com/hotosm/tasking-manager/blob/fix/view-all-projects-link/frontend/src/views/teams.js#L47

d-rita:
Fixes #3028


willemarcel (commented):
Let's plan some additional actions to that issue to the next sprint

d-rita:
Since the endpoint/functionality to filter teams by an organisation already exists in the backend, in this PR this frontend feature is built to help those with **manage access** to be able to do so.
**Before filtering:** 
before filter

**After filtering:**
after filter

Maybe this is a feature that would be useful to the community and/or can be added to/modified. cc: @dakotabenjamin @petya-kangalova @HelNershingThapa @Ichchhie @Aadesh-Baral

PR #4357, opened 2021-03-09
Display number of comments per task on "My tasks" list

### LINKED ISSUE #4098: Add number of comments information on "My tasks" list
It would be helpful for users if they could see the number of comments each task received on "My tasks" list

image

d-rita:
Fixes #4098 

Steps:

- [x] Return task history/number of comments in the UserTask endpoint
- [x] Display the number of comments for each task. 

Screenshot:

task number

### DIFF
--- a/backend/services/users/user_service.py
+++ b/backend/services/users/user_service.py
@@ -269,7 +269,7 @@ class UserService:
             TaskHistory.query.with_entities(
                 TaskHistory.project_id,
                 TaskHistory.task_id,
-                func.count(TaskHistory.action).label("cnt"),
+                func.count(TaskHistory.action).label("count"),
             )
             .filter(TaskHistory.action == "COMMENT")
             .group_by(TaskHistory.task_id, TaskHistory.project_id)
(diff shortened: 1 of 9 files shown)


backend/services/users/user_service.py
```diff
@@ -264,14 +264,37 @@ def get_tasks_dto(
         user_task_dtos = UserTaskDTOs()
         task_id_list = base_query.subquery()
 
+        comments_query = (
+            TaskHistory.query.with_entities(
+                TaskHistory.project_id,
+                TaskHistory.task_id,
+                func.count(TaskHistory.action).label("cnt"),
```
willemarcel, line 271: Instead of `cnt`, I suggest to use `count`. "Explicit is better than implicit"

willemarcel (approved):
I let a comment about a small detail on the code and pushed a commit with some improvements to the frontend.

It worked well. The increase on the response time is not relevant.

ramyaragupathy (approved):

PR #4407, opened 2021-03-22
Add basemap options to Priority Areas form

### LINKED ISSUE #2401: Add basemap options to priority area form
The project priority area form map should have the same basemap options as the project creation map.

d-rita:
Fixes #2401

### DIFF
--- a/frontend/src/assets/styles/_extra.scss
+++ b/frontend/src/assets/styles/_extra.scss
@@ -284,7 +284,7 @@ div.messageBodyLinks {
   }
 }
 
-#project-creation-map {
+#project-creation-map, #priority-area-map {
   .mapboxgl-ctrl-top-right {
     top: 2.5rem;
   }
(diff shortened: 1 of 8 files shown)


willemarcel (changes requested):
When I click to switch the layers, the map data disappears from the screen

willemarcel (changes requested):
The clear all button is not working. It's only resetting the data to the last saved one.

Let's style the aoi layer in orange the same way as in https://github.com/hotosm/tasking-manager/blob/develop/frontend/src/components/projectCreate/projectCreationMap.js#L110

And let's also style the priority area the same way as https://github.com/hotosm/tasking-manager/blob/develop/frontend/src/components/taskSelection/map.js#L240 (we use two layers to make that style possible)

willemarcel:
@d-rita I pushed some improvements to the style. A critical problem I found is that it's generating duplicated geometries. I didn't have time to investigate it deeply.

willemarcel (approved):

d-rita (commented):
The changes look good, @willemarcel

PR #4426, opened 2021-03-25
Add option to discard tasks smaller than 2000m²

### LINKED ISSUE #4118: Discard tiny tasks when generating task grid

willemarcel:
Resolves #4118

This PR also contains a refactoring of the project creation sources and layers management and change the colours of the AoI and of the task grid.


d-rita (commented):
I can distinguish the tiny tasks (less than 2000m²) and discard them, as well as create a project with the correct number of tasks. Nicely done 🎉

d-rita (approved):
LGTM!

PR #4447, opened 2021-03-29
Show loading icon on buttons during mapping/validation API requests

willemarcel:
Resolves https://github.com/hotosm/tasking-manager/issues/2168


d-rita (approved):
Tested submitting a task after mapping and validation, splitting a task, stop validation, and the loading icon is well displayed in the button during the loading state. Overall, LGTM1

PR #4451, opened 2021-03-29
Improve Login page

### LINKED ISSUE #4334: Improve login page
image

That page we show to users when they try to access a page that requires authentication could be improved to give more incentives to users to join.

willemarcel:
Resolves #4334


d-rita (approved):
Works well. The only change is the description of the test in `login.test.js` to name the Login component being tested.

PR #4452, opened 2021-03-29
Refactor duplicated components - server and error messages

### LINKED ISSUE #4376: Refactor duplicated components
Refactor `ServerMessage` and `ErrorMessage` and also declare it outside the components.
https://github.com/hotosm/tasking-manager/blob/develop/frontend/src/views/campaigns.js#L63

d-rita:
Fixes #4376


frontend/src/components/responseMessages.js
```diff
@@ -0,0 +1,21 @@
...
+const ServerMessage = ({ children }) => {
+  return (
+    <div className="red ba b--red pa2 br1 dib pa2">
+      <CloseIcon className="h1 w1 v-mid pb1 red mr2" />
+      {children}
+    </div>
+  );
+};
+
+export const ErrorMessage = ({ error, success, children }) => {
```
willemarcel, line 14: The `success` argument is not being used

willemarcel (changes requested):
I think we can go a bit deeper on this refactoring. Some additional things we can do:
- handle the API requests on the Create / Edit campaign components with the new useAsync
- Remove the error message after the input field change
- In both cases, we are calling Error Message with the same children. It's not clear how the return should change because of the `error` argument. Try to make it easier to understand.

If you want we can schedule a session to pair on it.

willemarcel (approved):
@d-rita I could disable the error alert when the form gets dirty, and I also improved the style of the Error Alert.

d-rita (commented):
Looks good @willemarcel

PR #4481, opened 2021-04-06
Modify language switch on project edit page

### LINKED ISSUE #4358: Modify language switch on project edit page
We received feedback that the language switch is not intuitive. Furthermore, when translating a project field, it requires switching from one language to the other multiple times.

It would be better to have an "Add" button to include a new language and show the input field of all translations on the page:
image

willemarcel:
Resolves #4358


d-rita (commented):
The added input field for a different language works quite well while maintaining the default field. Nicely done!

d-rita (approved):

Issue #4537, opened 2021-04-22
Improve user onboarding


willemarcel:
When a user registers on Tasking Manager, we need to provide some basic instructions in order to improve the quality of their first edits. Some ideas:
- Suggesting them to watch some videos before locking a task for mapping (Use the videos from https://tasks.hotosm.org/learn/)
- Load iD with the walkthrough tutorial
- Build a quiz or some type of interactive training

PR #4549, opened 2021-04-25
Implementation of a tutorial page helping user onboarding

lyndenW:
Issue: https://github.com/hotosm/tasking-manager/issues/4537
Feature: A tutorial page that runs an interactive quiz together with the videos from https://tasks.hotosm.org/learn/.

### DIFF
--- a/frontend/src/App.js
+++ b/frontend/src/App.js
@@ -15,6 +15,7 @@ import { Home } from './views/home';
 import { FallbackComponent } from './views/fallback';
 import { AboutPage } from './views/about';
 import { LearnPage } from './views/learn';
+import { TutorialPage} from "./views/tutorial";
 import { QuickstartPage } from './views/quickstart';
 import { UserDetail } from './views/userDetail';
 import {
@@ -97,6 +98,7 @@ let App = (props) => {
                   <OrganisationDetail path="organisations/:slug/" />
                   <LearnPage path="learn" />
                   <QuickstartPage path="learn/quickstart" />
+                  <TutorialPage path="learn/tutorial" />
                   <AboutPage path="about" />
                   <ContactPage path="contact/" />
                   <ContributionsPageIndex path="contributions">
--- a/frontend/src/views/messages.js
+++ b/frontend/src/views/messages.js
@@ -685,4 +685,8 @@ export default defineMessages({
     id: 'pages.create_campaign.error',
     defaultMessage: 'There was an error saving this campaign.',
   },
+  tutorial: {
+    id: 'pages.learn.tutorial',
+    defaultMessage: 'Tutorial',
+  }
 });
--- a/frontend/src/views/tutorial.js
+++ b/frontend/src/views/tutorial.js
@@ -1,0 +1,297 @@
+import React, { Component } from "react";
+import {TopBar} from "../components/header/topBar";
+import {FormattedMessage} from "react-intl";
+import messages from "./messages";
+
+const Videos = [
+    {
+        message: 'learnSignUp',
+        youTubeId: 'wqQdDgjBOvY',
+    },
+    {
+        message: 'learnMapBuildings',
+        youTubeId: 'nswUcgMfKTM',
+    },
+    {
+        message: 'learnMapRoads',
+        youTubeId: 'NzZWur1YG1k',
+    },
+];
+
+const QuizData = [{
+    question: "Which of the followings is not required for signing up on Task Manager?",
+    options: [{id: 1, text: "Name"}, {id: 2, text: "Email Address"}, {id: 3, text: "Password"},
+        {id: 4, text: "Driver License Number"}],
+    answer: 4,
+    message: "You need a valid name, email address, and password in order to sign up on Task Manager."
+}, {
+    question: "After signing up, which icon to click if you want to start contributing?",
+    options: [{id: 1, text: "Explore Projects"}, {id: 2, text: "My Contributions"},
+        {id: 3, text: "Learn"}, {id: 4, text: "About"}],
+    answer: 1,
+    message: "Clicking on 'Explore Projects' will direct you to the projects awaiting contribution."
+},{
+    question: "What kind of information is not provided for each project?",
+    options: [{id: 1, text: "Priority level"}, {id: 2, text: "Project number"},
+        {id: 3, text: "Difficulty level"}, {id: 4, text: "Completion level"}, {id: 5, text: "None of the above"}],
+    answer: 5,
+    message: "Priority level, Project number, Difficulty level, and Completion level are all provided and visable" +
+        "for all projects"
+},{
+    question: "True or False: You should always work inside the bright pink boundary when contributing.",
+    options: [{id: 1, text: "True"}, {id: 2, text: "False"}],
+    answer: 1,
+    message: "You should always work inside the bright pink boundary when contributing."
+},{
+    question: "True or False: You cannot choose if someone else can review your edits.",
+    options: [{id: 1, text: "True"}, {id: 2, text: "False"}],
+    answer: 2,
+    message: "You decide if someone else can review your edits when you commit your changes."
+},{
+    question: "If a road crosses another, how should you make sure they intersect?",
+    options: [{id: 1, text: "Click on the road it crosses"}, {id: 2, text: "Double click the map"},
+        {id: 3, text: "Close the webpage"}, {id: 4, text: "Nothing needs to be done"}],
+    answer: 1,
+    message: "If a road crosses another, you should click on the road it crosses wo ensure they intersect."
+},{
+    question: "If you don't know how to tag a road, which website should you check?",
+    options: [{id: 1, text: "tasks.hotosm.org"}, {id: 2, text: "wiki.openstreetmap.org"},
+        {id: 3, text: "petfinder.com"}, {id: 4, text: "store.steampowered.com"}],
+    answer: 2,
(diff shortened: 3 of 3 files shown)


SColchester:
What's the status with this one? If it is going ahead is there any chance of seeing a preview of it?

Ichchhie:
Hi @SColchester, here is a quick preview of what it looks like. Maybe we can have a discussion here or at the community meetup next week more on what others in the community think about this particular addition:
tutorial_page_tm

There are 7 questions having a youtube video associated with that question so that users can watch the video if the answer gets wrong or want to know more

**There will be things needed to be done before we merge this:**

1. Finalization of questions and options to keep
2. Make it supportable for other languages, right now it's only English supported
3. Design decisions to take on where would be the best place to keep this feature

What I feel here regarding point no 3 is that since the quiz can be primarily helpful to the users who have just started using Tasking Manager, so adding the quiz link in the welcome email can be the probable option to start with, or popping up a dialog to play the quiz in the first login and then keeping it somewhere inside learn page on a permanent basis.

Please feel to suggest some inputs once you see the preview. Thanks for bringing this into the attention!

Ichchhie:
Deploying into the staging isn't that practical right now in this point, as things need to be confirmed before doing that. However, if you would be interested I can list down the questions with associated Youtube videos here.

SColchester:
Sounds good thanks!

Ichchhie:
Below are the list of questions with related info for the quiz:
 
    Question 1:"Which of the followings is not required for signing up on Task Manager?",
    Options:"Name","Email Address","Password","Driver License Number"
    Message shown after the submission:"You need a valid name, email address, and password in order to sign up on Task Manager."
    Video: [here](https://www.youtube.com/watch?v=wqQdDgjBOvY)

    Question 2:"After signing up, which icon to click if you want to start contributing?",
    Options:"Explore Projects","My Contributions","Learn","About"
    Message shown after the submission:"Clicking on 'Explore Projects' will direct you to the projects awaiting contribution."
    Video: [here](https://www.youtube.com/watch?v=nswUcgMfKTM)

    Question 3:"What kind of information is not provided for each project?",
    Options:"Priority level","Project number","Difficulty level","Completion level","None of the above"
    Message shown after the submission:"Priority level, Project number, Difficulty level, and Completion level are all provided and visible for all projects"
    Video: [here](https://www.youtube.com/watch?v=nswUcgMfKTM)

    Question 4:"True or False: You should always work inside the bright pink boundary when contributing.",
    Options:"True","False"
    Message shown after the submission:"You should always work inside the bright pink boundary when contributing."
    Video: [here](https://www.youtube.com/watch?v=nswUcgMfKTM)

    Question 5:"True or False: You cannot choose if someone else can review your edits.",
    Options:"True","False"
    Message shown after the submission:"You decide if someone else can review your edits when you commit your changes."
    Video: [here](https://www.youtube.com/watch?v=nswUcgMfKTM)

    Question 6:"If a road crosses another, how should you make sure they intersect?",
    Options:"Click on the road it crosses","Double click the map","Close the webpage","Nothing needs to be done",
    Message shown after the submission:"If a road crosses another, you should click on the road it crosses wo ensure they intersect."
    Video: [here](https://www.youtube.com/watch?v=NzZWur1YG1k)

    Question 7:"If you don't know how to tag a road, which website should you check?",
    Options:"tasks.hotosm.org","wiki.openstreetmap.org", "petfinder.com","store.steampowered.com",
    Message shown after the submission:"wiki.openstreetmap.org provides useful information about how to tag different roads."
    Video: [here](https://www.youtube.com/watch?v=NzZWur1YG1k)

petya-kangalova:
Discussion on this issue took place in the Tasking Manager Meet Up session. There was an agreement that a lot more discussion and consultation needs to happen before implementation, given the impact on users. Everyone agreed that  Maybe we consult with the training-working-group and see if they can support on the content generation. 

Some consideration also to note:
- Understand whether this tutorial is optional for the users or not; this needs to be fleshed out.
- Once we get to the implementation, graphical representation will be helpful. 
- Once content for the tutorial is fully developed, it need to be tested thoroughly; user experience needs to be considered.

An alternative to the tutorial is using a sandbox, which is an alternative option and can be looked at separately and not as a dependency to creating the tutorial.

martien-176:
I like very much the idea of some sort of an introduction for new mappers that teaches them the basic skills necessary for proper mapping.

As a member of the HOT Data Quality Group i took part in the discussion that lead to this feature request. Imo the main aim of it was to solve the problem that many new mappers create bad data. And thus cause a lot of extra work for validators.

Sorry, imo the questionnaire presented here does not lead to a solution for this. 

How then i think should the questionnaire/quiz look like?

From my experience as a mapper and validator the basic skills boil down to:
1. use of the Tasking Manager. Eg how to choose a project and task. How to finish a task. How to write/read a task comment.
2. use of the iD Editor. Eg how to write a changeset comment, how to save mappings 
3. how to map buildings. Eg how to start mapping buildings, how to square/move and/or disentangle shared nodes

I have still some more thoughts on this, will follow up on later comments here.

RAytoun:
I personally do not see the relevance of these questions in relation to getting a mapper to map buildings correctly.

russdeffner:
A 'quiz' could look like a Captcha a maybe modeled a bit on MapSwipe - i.e. user selects a buildings project and is presented a pop-up that says "Hey, it looks like this is your first Buildings project. Here are some examples of what you might see while contributing to the project..." And given 6 images with a request to "Mark all the images correctly mapped for buildings". If they don't get it right, then point them to a training module (in LearnOSM).

The biggest problems we will have here is that many project managers add multiple features to map (though we are trying to address that via on-boarding training, it will still happen as sometimes buildings and residential polygons make more sense to do together; as with waterways and roads, etc. etc.)

My biggest suggestion here is whatever is done, it's going to need some A/B testing (i.e. are those who are required to do the quiz/tutorial actually producing a higher quality of mapping or is it really just slowing down most people from doing a decent job without any quiz/training).

martien-176:
I like very much the suggestion from Russ about A/B testing.

I imagine something like comparing 2 different crafted tutorials in 2 comparable projects. And then see what is the result of each.

johnwoollett:
Hi,

Can we ask if mappers understand how to use the Q button to square rectangular buildings

Also do they understand that when they submit task up as mapped the whole task needs to be checked for buildings etc.

aawiseman:
I agree with Ralph, Martien and John. The issues we are trying to get at are about bad mapping, not the sign up process or how to request feedback on changesets, most of which don't get reviewed anyway. I think the questions should focus on those issues. I think a Captcha would be best but might be too hard to implement, so we could use images or other related questions to get at them instead.

Some questions could be an image with multiple examples of buildings, some mapped well and some mapped poorly (such as not squared, untagged, way too big, overlapping) and the person has to pick which ones are mapped correctly - such as 1, 3, and 4, and then the answer explains why they are good or bad. 

Then the next questions could be something "Is the goal to map as many things as possible, or to map things as accurately as possible" (accurately) "How do you make the corners of a building square or right angle" (right click and square button, or Q), "If there are many buildings in an area should you map them as one large building or multiple buildings" (multiple) and so on.

johnwoollett:
Mapswipe has some simple tutorials that are useful

aawiseman:
Also to @Ichchhie and @petya-kangalova 's questions, I think the quiz should be required the first time someone maps. In my experience very few project managers use the optional TM features, such as preset restrictions or things like that. I think just a quiz would be sufficient, and would require less work to build than a sandbox or Captcha style thing.  I think the issue is that people often don't get sufficient training, even in events, so this would make sure they get the bare minimum.

I also think training WG or data quality WG could do a few trial runs with new mappers and the questions and see how the feedback is.

Patrik-Br:
As a validator I agree that the questions should be more focused on how to correctly map buildings. I really like aawiseman's idea about picking images how the buildings should and should not be mapped. 
I have saved some print screen [examples from my validation](https://bit.ly/3P3HIeK) (when the mapping quality is too poor I am sending them printscreens like this with before/after stage in the feedback).

martien-176:
Repetition is the key to learning (so they say), so a way to offer the quiz again (preferably on a somewhat more advanced level) if mappers are more experienced would be great.

How about a reward for those that succeed?

aawiseman:
For example, something like this. A few pictures with why they are correct or wrong, and then afterwards with what it should look like.

Does this building look correct?
No, it is mapped too large. Make sure to trace the building perimeter as accurately as possible.

too_large


Does this building look correct?
No, the corners are not squared. Click Q on your keyboard to address that.

not_square


Does this building look correct?
No, multiple separate buildings are mapped as one large building. Make sure to map them as individual buildings.

multiple_asone


Does this building look correct?
Yes, the size is accurate and it probably extends under the tree.

looks_good

ramyaragupathy:
@royallsilwallz - since the issue has been lying around for a while, for you to review and share details on what we can use from here?

ramyaragupathy:
cc @varun2948

Patrik-Br:
Hello, I just got reminded about this possible feature request. Reading the comments again after long time sounds like some of the related points are being currently worked on? Like badges might be implement this year (I think?) and there is also sandbox being worked on that could be usefull for this case? In my head of ideall dream implementation it could work in a way that mapper map a building (in the sandbox OSM version) and then get to see how it is supposed to looks like. I basically have in mind something like the MapSwipe tutorial that was mentioned before but in ideall option the mapper would actually map the building. But that's just ideal scifi world, just the "captcha" solution would be efficient as well.

Issue #4377, opened 2021-03-15
Filter buttons responsivity on My Tasks page


willemarcel:
![image](https://user-images.githubusercontent.com/666291/111222677-e9ed8200-85ba-11eb-838c-6131c3d50d64.png)

PR #4550, opened 2021-04-25
Fix filter buttons responsivity on My Tasks page

lyndenW:
Issue: https://github.com/hotosm/tasking-manager/issues/4377

### DIFF
--- a/frontend/src/components/contributions/myTasksNav.js
+++ b/frontend/src/components/contributions/myTasksNav.js
@@ -57,7 +57,7 @@ export const MyTasksNav = (props) => {
           </div>
         </div>
       </div>
-      <div className="mv2">
+      <div className="mv2" style={{ lineHeight: "200%" }}>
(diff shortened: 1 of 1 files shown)


willemarcel:
Hello, @lyndenW !

Thank you very much for your pull request! I reviewed it and saw that it works well with a window width superior to 600px, but it is not good on smaller widths, like on that screenshot:
image

I also suggest that you use the [tachyons classes](http://tachyons.io/docs/) instead of using styles, whenever possible.

willemarcel (changes requested):
Hello, @lyndenW !

Thank you very much for your pull request! I reviewed it and saw that it works well with a window width superior to 600px, but it is not good on smaller widths, like on that screenshot:
image

I also suggest that you use the [tachyons classes](http://tachyons.io/docs/) instead of using styles, whenever possible.

willemarcel:
Replaced by https://github.com/hotosm/tasking-manager/pull/4722

PR #4612, opened 2021-05-04
Parse new database connection parameters

eternaltyro:
Problem:

There's no straightforward way to prevent PostgreSQL password 
from being displayed in AWS console (our current application host).
This is a shortcoming of AWS Cloudformation and available options
for AWS ECS `ValueFrom` field.

Solution:

Instead of trying to parse the secrets manager entry and pick specific 
parameters in the CloudFormation template, we do the parsing in 
config.py and pass the whole JSON stored in secrets manager. 

This is agnostic of AWS and can be extended to any cloud provider or
server host in the future.

### DIFF
--- a/backend/config.py
+++ b/backend/config.py
@@ -56,6 +56,22 @@ class EnvironmentConfig:
     # Assamble the database uri
     if os.getenv("TM_DB", False):
         SQLALCHEMY_DATABASE_URI = os.getenv("TM_DB", None)
+    elif os.getenv("TM_DB_CONNECT_PARAM_JSON", False):
+        """
+        This section reads JSON formatted Database connection parameters passed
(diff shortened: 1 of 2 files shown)


willemarcel (approved):

willemarcel:
@eternaltyro before merging, could you document the new variable on https://github.com/hotosm/tasking-manager/blob/develop/example.env and on https://github.com/hotosm/tasking-manager/blob/develop/docs/deploy-aws.md

eternaltyro:
> @eternaltyro before merging, could you document the new variable on https://github.com/hotosm/tasking-manager/blob/develop/example.env and on https://github.com/hotosm/tasking-manager/blob/develop/docs/deploy-aws.md

Wille, I've updated example.env. The AWS deploy documentation will need more tuning and I think it's better if I merge it along with the rest of the CloudFormation code for it to make sense.

PR #4616, opened 2021-05-05
Add new CSS rule to hide "Improve this map" link

willemarcel:
I have noticed that in the production instance, the link to the "Improve this map" has been shown, so I improved the CSS rule to hide it.
image


d-rita (approved):
The link does not show. LGTM!

PR #4632, opened 2021-05-11
Fix git commands in PR review docs

d-rita:
Fixes typo in the command to checkout an external branch


willemarcel (approved):
@d-rita I have added some additional types

PR #4646, opened 2021-05-13
Update docker-compose and other docker files

tsmock:
* Update `Makefile` so it works (`app` -> `backend`/`frontend`, depending upon context)
* Use an official postgis image (that is actually updated, `mdillon/postgis:11` -> `postgis/postgis:11-3.1` (3.1 is the postgis version)
* Use official node images (`tiangolo/node-frontend:10` -> `node:14` (node 10 is no longer supported, and node 12 EOL is less than a year away).

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -37,7 +37,7 @@ services:
       - traefik.http.services.frontend.loadbalancer.server.port=80
 
   postgresql:
-    image: mdillon/postgis:11
+    image: postgis/postgis:14-3.3
     container_name: postgresql
     restart: always
     env_file: ${ENV_FILE:-tasking-manager.env}
@@ -45,7 +45,7 @@ services:
       - tm-web
 
   traefik:
-    image: traefik:v2.3
+    image: traefik:v2.10
     restart: always
     ports:
       - "80:80"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -18,7 +18,7 @@ services:
     container_name: backend
     restart: always
     labels:
-      - traefik.http.routers.backend.rule=Host(`localhost`) && PathPrefix(`/api/`)
+      - traefik.http.routers.backend.rule=(Host(`127.0.0.1`) || Host(`localhost`)) && PathPrefix(`/api/`)
       - traefik.http.services.backend.loadbalancer.server.port=5000
 
   migration:
@@ -33,8 +33,8 @@ services:
     networks:
       - tm-web
     labels:
-      - traefik.http.routers.frontend.rule=Host(`localhost`)
-      - traefik.http.services.frontend.loadbalancer.server.port=80
+      - traefik.http.routers.frontend.rule=Host(`127.0.0.1`) || Host(`localhost`)
+      - traefik.http.services.frontend.loadbalancer.server.port=3000
 
   postgresql:
     image: postgis/postgis:14-3.3
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -59,4 +59,4 @@ services:
 
 networks:
   tm-web:
-    external: true
+    external: false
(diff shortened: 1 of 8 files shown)


willemarcel:
Thanks, @tsmock!! @d-rita Could you review it when you have the possibility?

docker-compose.yml
```diff
@@ -37,7 +37,7 @@ services:
       - traefik.http.services.frontend.loadbalancer.server.port=80
 
   postgresql:
-    image: mdillon/postgis:11
+    image: postgis/postgis:11-3.1
```
eternaltyro, line 40: We can now safely use [postgis/postgis:14-3.3](https://registry.hub.docker.com/r/postgis/postgis/tags?page=1&name=14-3.3)

scripts/docker/Dockerfile.frontend
```diff
@@ -1,4 +1,4 @@
-FROM tiangolo/node-frontend:10 as build
+FROM node:14 as build
```
eternaltyro, line 1: We can upgrade node version to 16 as well.

scripts/docker/Dockerfile.frontend_development
```diff
@@ -4,25 +4,22 @@ RUN mkdir -p /usr/src/app
...
 # Get the Tasking Manager
-ARG branch=master
+ARG branch=develop
 RUN git clone --depth=1 git://github.com/hotosm/tasking-manager.git \
 	--branch $branch /usr/src/app
 
 RUN rm -rf backend/ migrations/
 
-FROM tiangolo/node-frontend:10 as build
+FROM node:14 as build
```
eternaltyro, line 13: node 16 here too.

eternaltyro (changes requested):
Now that we have tested later versions of Node and Postgres (and Postgis), we can safely upgrade version tags for both of them. 

When that's done, this PR is good to go.

tsmock:
I need to do some fiddling. It has been awhile since I made this PR, so I need to reverify everything.

tsmock:
I don't remember if I decided it was ready or not. I'll rebase onto current dev and see if it all works.

I got distracted by some other stuff (mostly going through dependencies on the frontend or working on JOSM).

tsmock:
OK. I've been going over some of the code:
* I _think_ `HEALTHCHECK --interval=60s --start-period=15s CMD ["curl", "-f", "http://localhost:8000/api/v2/system/heartbeat/", "||", "exit", "1"]` will work, but I haven't verified it.

`podman-compose up` is working properly insofar as starting everything. I probably messed up the network configuration (I wasn't able to connect).

Aadesh-Baral:
> podman-compose up is working properly insofar as starting everything. I probably messed up the network configuration (I wasn't able to connect).

Could you please clarify where you were unable to connect?

Aadesh-Baral:
> podman-compose up is working properly insofar as starting everything. I probably messed up the network configuration (I wasn't able to connect).

Could you please clarify where you were unable to connect?

dakotabenjamin:
@tsmock on the HEALTHCHECK, `docker compose` runs the frontend/backend on port 80. In addition, I needed to change the following to get it to work: 
```
diff --git a/docker-compose.yml b/docker-compose.yml
index 31f1db1f..bfcfb9cb 100644
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -59,4 +59,4 @@ services:
 
 networks:
   tm-web:
-    external: true
+    external: false
```

This doesn't have to do with your PR, I'm not even really sure why it was set as external in the first place. I think the docs for running/developing with Docker sorely need updating, but I don't think we need to require that to merge this PR. With the two changes suggested above, I think this is good to go, and we can follow up with a cleanup/clarity PR.

dakotabenjamin (approved):
LGTM

PR #4756, opened 2021-06-17
Use radio buttons on Task History instead of checkboxes

### LINKED ISSUE #4038: Change checkbox to radio button on Task History
A user reported that it would be better to have radio buttons (comments / activities / all) instead of checkboxes on the task history component:

"Also, would radio buttons be better than checkboxes?
With checkbox, you can have the unchecked/unchecked state which shows all items (same as checked/checked).
It's also strange when both are checked, and I want to just see comments that I have to click the activities box.
With radios, you could have   comments, activities, comments and activities (or All).   You only get 1 of them.   If I want comments, I click comments.   So, I'm clicking the thing I want (not what I don't want)."

image

d-rita:
Fixes #4038


willemarcel:
@d-rita it's missing to execute `yarn build-locales` 😃

willemarcel (approved):

PR #4805, opened 2021-07-14
Retrieve default comment to get correct project stats

### LINKED ISSUE #4804: Fix functionality to retrieve project stats
Currently, the functionality takes a project's changeset comment, splits it, and strips the `#` sign before making an API call to return a project's stats using the default comment which follows the format `organisationName-project-projectId`. Ideally, the default comment is the first entry. But in projects where it does not come first, this does not return any stats.   

Capture d’écran 2021-07-09 235033

d-rita:
Fixes #4804


frontend/src/utils/defaultChangesetComment.js
```diff
@@ -0,0 +1,8 @@
+export const retrieveDefaultChangesetComment = (changesetComment, projectId) => {
+  let pattern = `\\w*(-project-${projectId})$`;
```
ramyaragupathy, line 2: Here we could use the `TM_DEFAULT_CHANGESET_COMMENT` [variable](https://github.com/hotosm/tasking-manager/blob/develop/example.env#L105) set in config file

ramyaragupathy: Just realized that this  variable is accessible only in the backend. We should let this information accessible through the frontend config settings, so that the default comment update in one place reflects across the application

ramyaragupathy:
Tested the functionality - it works as expected and retrieves the stats based on the default comment.
Changes suggested:
*  Accessing the default changeset comment in the frontend through an env variable
* Adding an information in the statistics page that the numbers are retrieved based on the default changeset comment. If not users might flag this as an issue. So it is important we highlight this detail.

d-rita:
@ramyaragupathy I have made the changes: 
- Using the default changeset comment from the environment variables to check and retrieve the project stats. 
- And, adding a tooltip on the Edit Stats title indicating how these stats are got.

ramyaragupathy:
Thanks @d-rita - this looks good!

PR #4860, opened 2021-08-30
Point to the right success message while messaging contributors

ramyaragupathy:
image


d-rita (approved):
It works as expected. LGTM

PR #4919, opened 2021-11-05
Add second banner for summit

ethan-nelson:
Per Russ, I've created a banner for the HOT Summit to add to the tasking manager. I'm not sure if this should be merged upstream or if there is a separate deployment for the HOT-specific TM. 

First time a user visits the site:
Screen Shot 2021-11-04 at 8 54 35 PM

Hiding behind that first banner is information about the summit:
Screen Shot 2021-11-04 at 8 54 43 PM

In the past, we've had issues with temporary banners not retaining user's close preference and getting annoying; based on brief testing, the browser seems to hold the preferences since I mimicked the tracking banner's browser storage.

I originally wanted to do a broader solution based on a database table for announcements, but wasn't able to get that finished up in time. That may come down the road.

### DIFF
--- a/frontend/src/App.js
+++ b/frontend/src/App.js
@@ -57,7 +57,7 @@ import {
   NotificationPageIndex,
   NotificationDetail,
 } from './views/notifications';
-import { Banner } from './components/banner/index';
+import { Banner, DonationBanner } from './components/banner/index';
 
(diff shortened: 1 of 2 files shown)


dakotabenjamin:
This looks great! Do you think an we get a button for the registration link directly?

PR #4982, opened 2022-01-31
Fix for failing test/Unblock CircleCI

zlavergne:
CircleCi was failing due to a failed JS test. I'm not sure exactly why this particular test started failing all of a sudden but it seems the reason is because the specific check was using `getAllByText()` which gets all elements whose **entire** content is "Low". As the second element with "Low" in it had full content of "Actions remaining on the Low tier" it wasn't being caught by the function. 

Assuming the purpose of this check is to find the "Low"'s circled in this image, this fix should work:
Screen Shot 2022-01-31 at 4 49 38 PM


d-rita (approved):
This fixes the failing test. LGTM

Issue #5017, opened 2022-02-28
Add support page link


dakotabenjamin:
In order to improve the support service HOTOSM can provide to our hosted TM (tasks.hotosm.org) we should have a link to our helpdesk portal (https://hotosm.atlassian.net/servicedesk/customer/portal/4). 

This link should be configurable in the environment variables for other hosts to apply their own link.

PR #5024, opened 2022-03-02
Service Desk link addition to Navigation 

ramyaragupathy:
From #5017 - made frontend changes to let access Jira Service desk through Tasking Manager. This link can be used to report any issues that the users encounter with the application.

@d-rita - here is a snapshot from my local testing

image

The  external link always is appended to the base link. Can you please help figure out what more changes are needed to make this work?


d-rita:
The Link component is mostly used to route within the application. So, for external links, we go with the anchor tag, `<a />`. That way, the service desk link will not be appended to the site URL before redirecting. 

Additionally, the changes added so far only affect the popup menu. 
popup

To affect the navbar displayed on larger screens, we'd have to apply the same changes to this [part](https://github.com/hotosm/tasking-manager/blob/fb77daece5e0741f6eab12c8f54fb3cae8685cc0/frontend/src/components/header/index.js#L163).

zlavergne:
Just a nitpick, since this is an external link in a navbar, could we add an icon to denote it's external. Like this: https://fontawesome.com/v5/icons/external-link-alt?s=solid

HelNershingThapa:
An anchor tag has been added for the support link. Like @zlavergne suggested, an external link icon has also been added.
support-big

Popup:
support-small

dakotabenjamin:
Is this still WIP?

HelNershingThapa:
@dakotabenjamin it's done

Issue #5027, opened 2022-03-04
Implement OAuth 2.0 for OSM Login


Aadesh-Baral:
OSM now supporting OAuth 2.0 for login and OAuth 1.0 will go away at some point, according to TomH: "we will hopefully be retiring OAuth 1 in the relatively near future" ([openstreetmap/openstreetmap-website#3299](https://github.com/openstreetmap/openstreetmap-website/issues/3299)). Now its time to upgrade TM to use OAuth 2.0.

Aadesh-Baral:
The idea here would be to not only upgrade TM to use OAuth 2.0 but also make changes to the TM embedded iD editor and its underlying OAuth library to support OAuth 2.0.

zlavergne:
@Aadesh-Baral regarding updating the iD editor; it may be best to make that PR to the iD repo instead so that the changes will more easily propagate. I'm curious on your thoughts on this.

Aadesh-Baral:
@zlavergne we are planning a PR to osm-auth library. If approved it would be easy to propagate changes to other projects as well.

PR #5029, opened 2022-03-09
Implement OAuth 2.0 for OSM Login

Aadesh-Baral:
The main purpose of this PR is to upgrade the authorization mechanism from OSM OAuth1.0 to OSM OAuth2.0. More info: https://github.com/hotosm/tasking-manager/issues/5027


backend/services/users/authentication_service.py
```diff
@@ -141,15 +146,30 @@ def generate_session_token_for_user(osm_id: int):
...
-        session.pop("osm_oauthtok")
+    def generate_random_state(length=30, chars=UNICODE_ASCII_CHARACTER_SET):
+        """Generates a non-guessable OAuth token
+        OAuth (1 and 2) does not specify the format of tokens except that they
+        should be strings of random characters. Tokens should not be guessable
+        and entropy when generating the random characters is important. Which is
+        why SystemRandom is used instead of the default random.choice method.
+        """
+        rand = SystemRandom()
+        return "".join(rand.choice(chars) for x in range(length))
```
eternaltyro, line 159: Since this must be non-guessable and cryptographically secure, can we use a slightly extended character set using `secrets.token_urlsafe(bytes)` instead of `SystemRandom` ? It is equally secure and we get to specify the strength of the token by bytes of data rather than the length of the string. 

```
def generate_random_state(bytes=32):
    import secrets
    return secrets.token_urlsafe(bytes)
```

Some sample strings:

```
>>> secrets.token_urlsafe(32)
'Xe6Tzp565oVznQjb3h-RrTqV9nXMNzypyNBTQaH4UEA'
>>> secrets.token_urlsafe(32)
'I9D4XI8z8D7QuzAFF0EMtQlN7vmpkhOMj_3eBkAW2E0'
>>> secrets.token_urlsafe(32)
'47p-cFn_HFwGnRRai3ymDDIK7Nk5WWvFzln47NHWe1E'
>>> secrets.token_urlsafe(32)
'BjQXLqR9_3U76Wu3xkc9pdZ9xs6uYtSXotsva2t2RU0'
>>> secrets.token_urlsafe(32)
'pYvDwptqbOXULihumsqUPIzjpdT3E_hB_0DysAvr-qk'
>>> secrets.token_urlsafe(32)
'78E74KJ1lYppbwgfW91rQ_HvyuwaR8V51eZzDxLT8rg'
>>> secrets.token_urlsafe(32)
'vsETdIUcFjNI5XrdTnB57Ge2Nz0jHTRo5Ef7-GIayAo'
```

frontend/package.json
```diff
@@ -8,7 +8,7 @@
     "@formatjs/intl-pluralrules": "^4.0.27",
     "@formatjs/intl-relativetimeformat": "^9.1.6",
     "@formatjs/macro": "^0.2.8",
-    "@hotosm/id": "^2.20.2",
+    "@hotosm/id": "HelNershingThapa/iD.git#osm-auth2.0",
```
eternaltyro, line 11: Can you please describe this change?

frontend/yarn.lock
```diff
@@ -2399,7 +2398,7 @@
     lodash-es "~4.17.15"
     marked "~2.0.0"
     node-diff3 "2.1.0"
-    osm-auth "1.1.0"
+    osm-auth "git+https://github.com/HelNershingThapa/osm-auth.git"
```
eternaltyro, line 2401: Please describe this change as well.

eternaltyro (commented):
@dakotabenjamin I took a quick look (with the exception of the CFN template) and added some comments. LGTM otherwise. We will need to test the changes for good measure.

The change to Cloudformation template seems to be just formatting . But I haven't looked into it in detail.

frontend/package.json
HelNershingThapa: This uses my [forked iD repo](https://github.com/HelNershingThapa/iD/tree/osm-auth2.0) as the _@hotosm/id_ dependency. I'll send a PR to once we've verified that these changes work as expected.

frontend/yarn.lock
HelNershingThapa: This uses my [forked osm-auth repo](https://github.com/HelNershingThapa/osm-auth) as the _osm-auth_ dependency. This is only for the time being and we'll be planning to send a PR to the dependencies upstream once we've verified that these changes work as expected. Or, since these changes in dependencies have been made with a focus on TM, we could maybe publish these libraries(_osm-auth_ and _ohauth_) from the HOT's npmjs account.

backend/services/users/authentication_service.py
Aadesh-Baral: That sounds great.

dakotabenjamin:
This does look good to me as well. I think we should try to make a PR for the two dependencies to their upstreams, because I do not want us to have to manage updates to them later down the line. If those changes are rejected, then we can consider using our own forks.

eternaltyro:
@Aadesh-Baral Since the `secrets` module might be troublesome to install, we can simply increase the length of the string to maybe `48` and add `_` and `-` to the character set. It's a band-aid bit it should be sufficiently secure.

Aadesh-Baral:
Yes, secrets have python-ldap as one of its dependency which needs other system dependencies as mentioned [here]( https://stackoverflow.com/questions/4768446/i-cant-install-python-ldap ). Will update PR increasing string length and adding `-` and `_` to character set.

eternaltyro:
@dakotabenjamin I've reviewed it and apart from the changes to the cloudformation template, it looks good. 

Can the changes to CloudFormation template be reverted please @Aadesh-Baral ?

Aadesh-Baral:
Sure @eternaltyro...

dakotabenjamin:
@Aadesh-Baral I made a commit here, but I can't push it to your repository: https://github.com/hotosm/tasking-manager/commit/823f2b4dc03cdae068a78e2745dfd718a3aac631

I noticed tests are failing on your side because the testing environment isn't setup- but the branch I made was tested here: https://app.circleci.com/pipelines/github/hotosm/tasking-manager/9002/workflows/42a12d40-173f-47fb-b6bb-ba58965824da/jobs/13908

Do you know how to fix this failure? 

```
#!/bin/bash -eo pipefail
cd ${CIRCLE_WORKING_DIRECTORY}
mkdir ${CIRCLE_WORKING_DIRECTORY}/tests/backend/lint
black --check manage.py backend tests migrations
flake8 manage.py backend tests migrations

would reformat /home/circleci/app/backend/api/system/authentication.py
Oh no! 💥 💔 💥
1 file would be reformatted, 274 files would be left unchanged.

Exited with code exit status 1

CircleCI received exit code 1
```

Aadesh-Baral:
@dakotabenjamin You can commit it to oauth2.0 branch of this repo and then i will sync my branch with it. 

This can be fixed with `black manage.py backend tests`. Let me know once you made a commit on if you want me to work on this fix.

PR #5034, opened 2022-03-09
Update READ.ME

petya-kangalova:
Simplified and updated the text. Removed the page referencing Working Group meetings

@dakotabenjamin kept the developer section as is and we can update later on. On the instances, we can also update later on if there are other instances missing.


Aadesh-Baral (approved):

Aadesh-Baral:
@petya-kangalova  I think images that are in the current Readme file are important. What are your thoughts about it?

petya-kangalova:
@Aadesh-Baral  happy for you to keep the images and then approve. Thanks so much!

Aadesh-Baral (approved):

Issue #3540, opened 2020-08-25
Improve task lock error handling on frontend


willemarcel:
If the user tries to lock a task, but someone else has already locked it (we refresh tasks states each 60s), it will return those errors:

`ERROR: Task 733 is not MAPPED, BADIMAGERY or VALIDATED [in /tasking-manager/backend/services/validator_service.py:35]`

`ERROR: Task in invalid state for mapping [in /tasking-manager/backend/services/mapping_service.py:28]`

We need to handle it better on the frontend and show a more informative message.

HelNershingThapa:
Closed in 5376cc2a76c9d16bf5f94a6ff9037d3024098b15

Issue #3069, opened 2020-06-02
Handle other types of lock errors


willemarcel:
If a user try to validate a task mapped by him/herself, we have a lock error, but we are not showing the correct message to the user.

PR #5039, opened 2022-03-14
Improve task lock error messages

HelNershingThapa:
This pull request addresses issue #3069. It does so by checking if the status code from the response is 403.

I've also updated the LockError component so that it can act somewhat like a generic error component. 

TODO: When trying to address #3540, I noticed that the status code is also 403 which limits the above implementation. I think we'll need to introduce some other variables in the response so that we can generate even better error messages.

cc @Aadesh-Baral


Aadesh-Baral:
@HelNershingThapa I have added `SubCode` along with an error message on API error response.

dakotabenjamin:
I am getting some errors on staging related to this PR I think https://sentry.io/share/issue/636c7f21a10a4581a82e574ff729da87/

PR #5044, opened 2022-03-16
set-default-permission-for-validation-to-intermediate-and-advanced

### LINKED ISSUE #4954: Set Default Permission for Validation to 'Intermediate and Advanced'
As with #1753 this isn't asking for any change outside of setting the default value to 'not beginner'. It has come up in the Quality Assurance and Control Working Group that possibly we would not allow beginner mappers to ever validate; but I think that's going too far. There may be a situation where you are training people or otherwise have identified new OSM profiles to be able to validate. 

So, a compromise might be that project creators would be 'forced' to consciously change the permission value of validators to any mapper by making the default restricted to intermediate and advanced. Thank you for considering, this should be a relatively easy change for a big increase in data quality.

defaultvalidation

Aadesh-Baral:
This PR closes #4954 
147188688-cc78024d-3a87-48f5-adef-40fe7366b61b

### DIFF
--- a/backend/models/postgis/project.py
+++ b/backend/models/postgis/project.py
@@ -132,7 +132,7 @@ class Project(db.Model):
     )  # Mapper level project is suitable for
     mapping_permission = db.Column(db.Integer, default=MappingPermission.ANY.value)
     validation_permission = db.Column(
-        db.Integer, default=ValidationPermission.ANY.value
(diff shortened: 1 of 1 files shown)


dakotabenjamin (approved):

aawiseman:
Awesome, thanks! I think this will help a lot with good data quality.

PR #5046, opened 2022-03-18
Fix/cfn init codedeploy

dakotabenjamin:
Fixes an issue with server initialization. We move the codedeploy init to the end of the script (Based on these suggestions https://docs.aws.amazon.com/codedeploy/latest/userguide/integrations-aws-auto-scaling.html#integrations-aws-auto-scaling-behaviors) and add a 60 second wait, to allow server init apt package lock to release.


scripts/aws/cloudformation/tasking-manager.template.js
```diff
@@ -340,6 +340,7 @@ const Resources = {
       UserData: cf.userData([
         '#!/bin/bash',
         'set -x',
+        'sleep 60',
```
eternaltyro, line 343: We could optionally complicate things by writing a check to see if apt process is locked. Something like this: https://gist.github.com/tedivm/e11ebfdc25dc1d7935a3d5640a1f1c90

eternaltyro (approved):
LGTM.

PR #5048, opened 2022-03-20
Fix user permission on Projects API

### LINKED ISSUE #5047: Only allow users with at least project manager role to send messages to all contributor of project.
**Describe the bug**
While going through `ProjectsActionsMessageContributorsAPI` i found out that any user can send a message to all contributors of a project. There is a function that checks if the user has permission to send messages to all users. But the result of the function is not used while sending messages.
https://github.com/hotosm/tasking-manager/blob/8cffe3420563cdb59170442e7ede14d08ebe993b/backend/api/projects/actions.py#L130-L132
This might not be a big issue considering sending message option is only available on the edit projects section which can be only accessed by managers and it can be executed through api-docs or directly calling API. 

**Expected behavior**
Only users with at least a project manager role must be able to send messages to all contributors of the project.

Aadesh-Baral:
fixes #5047


Aadesh-Baral:
Found the same issues on other Projects APIs. So this PR needs an update to address them as well. Drafting this for now

PR #5076, opened 2022-04-07
Display error message for self intersecting AOI

### LINKED ISSUE #4911: Trim Action for self-intersecting polygons during Project Creation fails with no error message
**Current Behaviour:**

When a user creates a project area that's self intersecting, `Trim` Action results in Exception and error is logged in the backend.
Example error log:

`TopologyException: Input geom 0 is invalid: Self-intersection at or near point -21.006146092804485 -20.134594635489307 at -21.006146092804485 -20.134594635489307
[2021-10-25 23:01:48,131] CRITICAL in actions: IntersectingTiles GET API - unhandled error: The operation 'GEOSIntersection_r' could not be performed. Likely cause is invalidity of the geometry <shapely.geometry.multipolygon.MultiPolygon object at 0x10736d5e0>`
self-intersection

**Expected Behaviour:**

If the input polygon is self-intersecting, indicate the issue to the user when the `Trim` Action is initiated.

Aadesh-Baral:
closes #4911

### DIFF
--- a/backend/api/projects/actions.py
+++ b/backend/api/projects/actions.py
@@ -12,6 +12,7 @@ from backend.services.messaging.message_service import MessageService
 from backend.services.users.authentication_service import token_auth, tm
 from backend.services.interests_service import InterestService
 from backend.models.postgis.utils import InvalidGeoJson
+from shapely.errors import TopologicalError
 
 
 class ProjectsActionsTransferAPI(Resource):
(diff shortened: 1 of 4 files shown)


HelNershingThapa:
Error message for self-intersecting AOI will be displayed as follows:
intersecting

And for errors that do not have a corresponding translation string:
generic-trim-error

PR #5071, opened 2022-04-05
Release 4.4.13

dakotabenjamin:
This is our first release in collaboration with Kathmandu Living Labs, and we are really excited to see all the new features and fixes coming from the talented people on that team! 

# Changelog 

**Added ~ New features:**
 
   - Add embedded RapiD editor by @zlavergne #4934
   - Add blocked option to dropdown by @HelNershingThapa  #5040
   - Add cheatsheet under manuals by @HelNershingThapa #5049
   - Add support link by @ramyaragupathy and @HelNershingThapa #5024
 
 **Changes ~ Bug fixes and enhancement:**
   - Fix reset all bad imagery tasks API by @zlavergne #5018
   - Update contributing guidelines by @petya-kangalova #5030
   - Update Readme by @petya-kangalova and @Aadesh-Baral  #5034
   - Update issue templates by @petya-kangalova #5026
   - Update bug report template by @petya-kangalova  #5028
   - Fix/cfn init codedeploy by @dakotabenjamin #5046
   - Change default validation permission by @Aadesh-Baral #5044
   - Update task lock error message by @Aadesh-Baral and @HelNershingThapa #5039
   - Update text in 'Learn to manage' section by @HelNershingThapa #5066
   
 **Known issues:**
   - The embedded editor does not work when switching between RapiD and iD. Please save your work before changing your editor. 
   - We are missing translations. See #5072

PR #5093, opened 2022-04-11
TM v4.4.14 (Hotfix)

dakotabenjamin:


dakotabenjamin:
It would appear that my name was added to the commits from the [last release PR](https://github.com/hotosm/tasking-manager/pull/5071) and now the branches have diverged. Normally we use the "Rebase and merge" option when merging these released to `master` and that shouldn't affect the commits. See [this doc](https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/about-merge-methods-on-github#rebasing-and-merging-your-commits) for more details. 

I want to propose that we actually change our process here and stop merging code back into `master`. It is a bit of an unnecessary step when we already tag commits for releases. What does it matter that the commits are on `develop` or `master`? We can still review pre-release code on tasks-staging as they are automatically deployed from `develop`. We don't perform feature freezes and are fairly diligent about deploying soon after a release is made, and in any case the deployment branches can be rebased against `develop` at a specific tag (such as "v4.4.13").

PR #5096, opened 2022-04-13
Handle switching editors with unsaved edits

### LINKED ISSUE #5025: Error while switching between RapiD and iD editor after RapiD integration.
There is an issue while switching between RapiD and iD after the integration of RapiD on the staging instance. There is no issue if there are no user changes. Also, there is no issue if the user clicks on reload button presented on the popup. But there is an issue if the user clicks on the cancel button on the popup.
To replicate this issue:

- Select any tasks to map in RapiD enabled project on [staging](https://tasks-stage.hotosm.org). [[Project with RapiD enabled ](https://tasks-stage.hotosm.org/projects/8616)]
- Make changes using RapiD or iD without uploading changes.
- Try reloading the editor. Switch to RapiD if you made changes using iD and vice versa.
- Click `cancel` on the popup.

Known Issue if clicked on cancel while reloading from iD to Rapid:
- Returns error page with `An error occurred ` that does resolve using `return` button until the page is reloaded.
Screenshot from 2022-03-02 16-56-22

Known Issues if clicked on cancel while reloading from RapiD to iD:
- All of the changes are lost.
- Icon of RapiD assist doesn't appear.
- RapiD assist layer color is changed.

Screenshot from 2022-03-03 10-33-36

zlavergne:
There was a problem reported with the RapiD integration when a user would attempt to switch editors from RapiD to iD when the user still had unsaved edits (#5025)

This PR does 3 main things to solve this issue:
1. Uses the `UnsavedMapChangesModalContent` component to show an alert if a user still has unsaved edits and tries to switch editors
tm-rapid_unsaved_edits
2. To make the actual component switch more performant, I changed how the editor switch is handled. Before, if you switched between RapiD and iD, it would reload the entire page; now it will only reload the single editor components.
3. To allow (2) to happen, I had to add a RAPIDCONTEXT to the redux store so the contexts don't get confused during the switch since they have slightly different values in their contexts.

Closes #5025

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -28,6 +28,7 @@
     "@turf/transform-scale": "^6.4.0",
     "@turf/truncate": "^6.4.0",
     "@webscopeio/react-textarea-autocomplete": "^4.7.3",
+    "RapiD": "facebookincubator/rapid#rapid-v1.1.8-tm.1",
     "axios": "^0.21.1",
     "chart.js": "^3.3.2",
     "date-fns": "^2.28.0",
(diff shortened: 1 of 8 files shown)


HelNershingThapa:
I've added an error popup that shows up when trying to switch to JOSM. Other than that, I tested the switching process locally, looks good to me @Aadesh-Baral

frontend/package.json
```diff
@@ -28,6 +28,7 @@
     "@turf/transform-scale": "^6.4.0",
     "@turf/truncate": "^6.4.0",
     "@webscopeio/react-textarea-autocomplete": "^4.7.3",
+    "RapiD": "facebookincubator/rapid#rapid-v1.1.8-tm.1",
```
HelNershingThapa, line 31: Will removing the previously installed RapiD dependency(rapid-v1.1.8_tm) be safe?

zlavergne: It should be totally safe. Making the editors be different components instead of reloading the page required a few changes to RapiD as well.

HelNershingThapa: Removed on 27ab132bbea2d108fbe51c17f09b84622b9e78e0

PR #5104, opened 2022-04-18
Enable adding YouTube videos with the ::youtube[] markdown tag

willemarcel:
Enable embedding YouTube videos in project's description and instructions. The sintax is `::youtube[video-id]`, it needs one line of spacing from other content (both up and down) to work.

**Project editing:**
image

**Project task's page:**
image

### DIFF
--- a/frontend/src/utils/htmlFromMarkdown.js
+++ b/frontend/src/utils/htmlFromMarkdown.js
@@ -1,9 +1,59 @@
 import { marked } from 'marked';
 import DOMPurify from 'dompurify';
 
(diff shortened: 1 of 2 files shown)


martien-176:
It would imo be more user friendly if the video-id part could be a full Youtube url. That way you can just copy that from the SHARE button on Youtube.

I think users can have difficulties when they have to strip out the video-id of the youtube url. 

Even more difficulties with a url like https://www.youtube.com/watch?v=Xs5wX592E1o&t=2s

PR #5105, opened 2022-04-18
Include extraIdParams field to projects

willemarcel:
The [iD API](https://github.com/openstreetmap/iD/blob/develop/API.md#url-parameters) supports some query parameters in the URL that makes it possible to load the editor with additional configuration.

This PR adds a new column to the project's table and loads both iD and Rapid with the extra parameters configured in each project.

**For example:**

image

Loads the iD editor with:

- A date range set in the photo overlays section (activate Mapillary to see it):

image

- the buildings layer disabled:

image

- and, finally, an offset to the background imagery:

image


frontend/src/components/projectEdit/extraIdParams.js
```diff
@@ -0,0 +1,17 @@
...
+    setProjectInfo((p) => ({ ...p, extraIdParams: e.target.value }));
+  };
+
+  return (
+    <input
+      className={styleClasses.inputClass}
+      onChange={onInputChange}
+      type="text"
+      name={`extraIdParams`}
+      value={value}
```
zlavergne, line 14: let's add a `placeholder` with the same example as the description example: `disabled_features=buildings&offset=-10,5`. Might even pass it in as a param so we can just change it in one spot if we ever need to.

zlavergne (approved):
Just one small, non-critical addition. I'll accept even without it. Looks great; thanks @willemarcel !

Aadesh-Baral (approved):
Looks good to me, need to change the revised migration version before merging

HelNershingThapa (approved):

PR #5120, opened 2022-05-09
Add checkbox to redirect to previous project after unlocking task in another project

HelNershingThapa:
If I'm trying to map a task in Project X, but I have a task locked in Project Y, I'd have to unlock the task in Project Y first to map the task in Project X. After unlocking the task in Project Y, there wasn't an option to be redirected to Project X.

This PR adds a checkbox input field that can be toggled to be redirected to Project X after unlocking the task in Project Y.

redirect-prev

### DIFF
--- a/frontend/src/components/taskSelection/action.js
+++ b/frontend/src/components/taskSelection/action.js
@@ -1,6 +1,6 @@
 import React, { useState, useEffect, useCallback, useMemo } from 'react';
 import { useSelector } from 'react-redux';
-import { navigate } from '@reach/router';
+import { navigate, useLocation } from '@reach/router';
 import ReactPlaceholder from 'react-placeholder';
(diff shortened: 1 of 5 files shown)


HelNershingThapa:
I tried implementing this with the help of states in @reach/router but couldn't find a workaround to capture the location states after the URL changes with the imagery background. So I switched to using local storage.

Patrik-Br:
Tested. It works perfectly, great idea 👍 It's something that I didn't knew I need till I tried 😄

martien-176:
https://docs.google.com/document/d/1iSp0k4Dpxh-0mYu-OV1uU9puuGzRdmn9YWl7-sHuxW8/edit

I tested acc. to the above sheet. It works as expected. 

This is a good emergency measure for when you mess things up while working in multiple projects at the same time.

PR #5172, opened 2022-06-06
Added switch field to enable/disable notifications

HelNershingThapa:


Aadesh-Baral:
Due to a change in implementation, this PR will be reworked on #5218.

PR #5184, opened 2022-06-15
build: update backend dependency management to pdm + dockerfile

### LINKED ISSUE #4642: Update backend dependencies management
Evaluate the use of [pipfile](https://github.com/pypa/pipfile)

spwoodcock:
Closes #4642

Looks like this issue has been open for a while.

Great suggestion by @frafra to use PDM. Thank you.

Notes:
- I updated the Dockerfile.backend to reflect the changes.
- I also added the --no-cache flag to the final stage apk install command to save some space.
- I pinned the Python version number for version solving to >=3.7,<=3.8.
- The values for pyproject.toml were taken from the repo. Let me know if they need to change.


swoodcoc:
This is only a partial solution, as there are multiple references to requirements.txt throughout the repo.

python-venv is also used, but is now redundant with using pdm.

I have a commit on this branch that I have not pushed yet, removing requirements.txt and modifying any docs / workflows with reference to it (14 files).
Let me know if you would like me to push.

Aadesh-Baral:
@swoodcoc I think it will be easier to review if you could push it.

swoodcoc:
No worries, I'll add it now.
I was worried it would increase the scope, as it also affects CI and deployment instructions.

I have updated all references to requirements.txt and using venv that I could find in the repo.

As for the benefit:
- PDM uses a dependency solver to check for compatibility between all dependencies.
- If a package is updated in the future that is incompatible with other package version, then it should be flagged during build.
- I also think PEP 582 dependency management is a lot simpler and reduces the need for virtual environments.

swoodcoc:
So it actually looks like you can't pip install dependencies directly from a pyproject.toml (I have never had to before).
There is a related issue https://github.com/pypa/pip/issues/8049

I wanted to keep the edits to CI etc simple by pip installing the already solved dependency list.

To achieve this we can
```
pip install toml && python -c 'import toml; c = toml.load("pyproject.toml")
print("\n".join(c["project"]["dependencies"]))' | pip install -r /dev/stdin
```

I will update.

swoodcoc:
Everything should be complete to the best of my knowledge.

There seem to be quite a few Dockerfiles for the backend to maintain - I updated them all, as I wasn't sure which ones are redundant or still used by external processes.

swoodcoc:
Also on a side note: the Dockerfile using `quay.io/hotosm/base-python-image` as a base uses Python **3.8.7**, while the Dockerfile.backend uses **3.7**. Perhaps a refactor is required to manage a consistent Python version.

The dependency install worked without issue (no conflicts) using Python **3.9**, so I feel like both could be updated to that anyway.

Aadesh-Baral (commented):
Some dependencies failed to install while setting up locally without docker but no such issues while using docker, other than that this looks good to me. Also, we may need to update the doc mentioning that the python version set in `pyproject.toml` should be preinstalled on the system before setting up locally without docker.

swoodcoc:
Thanks for the feedback, I'm more than happy to go through the docs and update as requested. 

As for the package installs, are you able to `pip install` the packages, but not `pdm install`? I would be surprised if this is the case.

I imagine the reason install failed is because of missing libraries on your Linux system, such as `proj-dev`.

A local install requires all the libs packaged into the docker image to be installed on your machine: https://github.com/hotosm/tasking-manager/blob/4757148e011f2c66f3452a588379e67daef0e3af/Dockerfile#L11-L22

The Dockerfile above uses Alpine, so the packages would be named differently for example on a Debian based distro (e.g. Ubuntu). `proj-dev` --> `libproj-dev` etc.

eternaltyro:
Related to #5705

spwoodcock:
I could update this PR and rebase with the current `develop`, if of interest.

eternaltyro:
Thank you, @spwoodcock . We are currently swamped and unable to review your upgrades immediately. Can you please give us a week or two to get `develop` branch cleaned-up and in order? I will keep you posted or you can follow the issues here on GH.

eternaltyro:
@ramyaragupathy @Aadesh-Baral let us discuss this in our next meeting and go over timelines and effort involved.

spwoodcock:
Of course @eternaltyro, no rush at all! Send me a message over any platform when you have time for this 👍

Dockerfile
```diff
@@ -1,12 +1,12 @@
 FROM quay.io/hotosm/base-python-image as base
-LABEL version=0.1
-LABEL maintainer="HOT Sysadmin <sysadmin@hotosm.org>"
-LABEL description="Builds backend docker image"
-
-WORKDIR /usr/src/app
+LABEL version=0.2
+LABEL hotosm.org.maintainer="HOT Sysadmin <sysadmin@hotosm.org>"
+LABEL hotosm.org.description="Builds docker image for TM Backend"
```
eternaltyro, line 4: I understand MAINTAINER label is deprecated and we can simply use LABELS for it. But what's the reason for using hotosm.org.maintainer as key? Docker suggests tld.domain.sub.maintainer but I'm not sure if it is standard. If it is, we can switch this to org.hotosm.maintainer.

Documentation about this is sparse though.

Dockerfile
```diff
@@ -21,22 +21,29 @@ RUN apk update && \
...
-
-RUN pip install \
-    --prefix=/install \
-    --no-cache-dir \
-    --no-warn-script-location \
-    -r requirements.txt
+# Setup backend Python dependencies
+COPY pyproject.toml pdm.lock README.md ./
+RUN pip install --no-cache-dir --upgrade pip \
+    && pip install --no-cache-dir pdm==2.0.2 \
```
eternaltyro, line 27: Latest `pdm==2.5.2`

Dockerfile
```diff
@@ -21,22 +21,29 @@ RUN apk update && \
...
+    && pdm config python.use_venv false
```
eternaltyro, line 28: Do we also need `--pep582` ?

Dockerfile
```diff
@@ -21,22 +21,29 @@ RUN apk update && \
...
+
+# Setup backend runtime dependencies
 RUN apk update && \
-    apk add \
+    apk add --no-cache \
         postgresql-libs geos proj-util
 
-COPY --from=builder /install /usr/local
+COPY --from=builder \
+    /install/__pypackages__/3.8 \
```
eternaltyro, line 45: In concept, this is okay. But when we upgrade out of Python 3.7, we might as well go to Python 3.10. In for a penny, in for a pound.

eternaltyro: I mention this because there could be further conflicts when this is merged, eventually after the current conflicts are resolved. This will depend on my work on the CI pipelines.

pyproject.toml
```diff
@@ -0,0 +1,83 @@
+[project]
+name = "HOTOSM Tasking Manager"
+version = "0.1.0"
```
eternaltyro, line 3: What is this the version of? Tasking Manager or the TOML file? If it is Tasking Manager, we could bump this before every release.

pyproject.toml
```diff
@@ -0,0 +1,83 @@
...
+    "sqlalchemy==1.3.20",
+    "text-unidecode==1.3",
+    "toml==0.10.1",
+    "unidecode==1.1.1",
+    "urllib3==1.25.10",
+    "webencodings==0.5.1",
+    "werkzeug==0.16.1",
+    "flask-mail==0.9.1",
+]
+requires-python = ">=3.7,<=3.8.13"
```
eternaltyro, line 74: This needs to be revisited. But okay for now.

scripts/docker/Dockerfile.backend
```diff
@@ -22,16 +22,26 @@ RUN apk update && \
...
+ENV PATH="/usr/src/python/bin:$PATH" \
+    PYTHONPATH="/usr/src/python/lib"
+
 # Setup backend runtime dependencies
-RUN apk update && apk add postgresql-libs geos proj-util
+RUN apk update && apk add --no-cache postgresql-libs geos proj-util
 
-COPY --from=builder /install /usr/local
+COPY --from=builder \
+    /install/__pypackages__/3.7 \
```
eternaltyro, line 43: [revisit] Same as the other Dockerfile.

scripts/docker/tasking-manager/Dockerfile
```diff
@@ -3,6 +3,9 @@ FROM python:3.7-buster
 RUN mkdir -p /usr/src/app
 WORKDIR /usr/src/app
 
+ENV PATH="/usr/src/app/__pypackages__/3.7/bin:$PATH" \
+    PYTHONPATH="/usr/src/app/__pypackages__/3.7/lib"
```
eternaltyro, line 7: [revisit] Same as the other Dockerfile.

eternaltyro (commented):
@spwoodcock I have left some comments here - mostly prompts and reminders to discuss these. Everything else looks good. This is a significant quality upgrade to the Backend infrastructure. Thank you very much for taking the time and effort to do this. 

@Aadesh-Baral Let's discuss this once we meet next week. Please hold merging even after current merge conflicts have been resolved. cc: @dakotabenjamin @ramyaragupathy

Dockerfile
spwoodcock: I don't think there are any standards around label keys as far as I know, we can definitely change it 👍

Dockerfile
spwoodcock: Running that will set the PATH variable for the current shell session, but we need it to persist throughout the docker layers. I do that here https://github.com/hotosm/tasking-manager/blob/385d7c77e8c0eeafc750dbcc4bb508ac08b60331/Dockerfile#L36

If running tasking manager without docker, the user may have to run --pep582 first to recognise the installed dependencies dir `__pypackages__`

Dockerfile
spwoodcock: Agreed 👍 The version should probably set as a build ARG

eternaltyro: That will be better.

pyproject.toml
eternaltyro: I will check that out.

eternaltyro:
> I'll try to find time to revise this PR soon (particularly the dockerfile). There are a few refinements I could include, plus Python 3.10.

Thank you, @spwoodcock . Let me know once you do, and I'll give it another review.

.circleci/config.yml
```diff
@@ -44,7 +44,8 @@ jobs:
           cd ${CIRCLE_WORKING_DIRECTORY}
           # Install Python dependencies
           pip install --upgrade pip
-          pip install -r requirements.txt
+          pip install toml && python -c 'import toml; c = toml.load("pyproject.toml")
+          print("\n".join(c["project"]["dependencies"]))' | pip install -r /dev/stdin
```
eternaltyro, line 48: I wanted to ask - why not simply do `pip install pdm && pdm install`  instead of parsing the toml file and piping it to pip?

spwoodcock:
I should get around to updating this at the weekend - got a pretty full work week ahead 👍

.circleci/config.yml
spwoodcock: Let me know if you have any thoughts! I haven't used CircleCI much to be honest - so it could just be my error 😅

eternaltyro: Thanks, I'll test with `pdm export`.

pyproject.toml
```diff
@@ -0,0 +1,83 @@
+[project]
+name = "HOTOSM Tasking Manager"
+version = "0.1.0"
+description = "The tool to team up for mapping in OpenStreetMap"
+authors = [
+    {name = "HOT Sysadmin", email = "sysadmin@hotosm.org"},
+]
+dependencies = [
```
Aadesh-Baral, line 8: Dev dependencies should be separated here. Once #5642 is merged, we can use it as a reference since it has a separation of dependencies in requirements.txt.   I think only direct dependencies will be listed in pyproject.toml because pdm.lock will handle all indirect dependencies.

Aadesh-Baral: We can do it in another PR if  we are merging this ahead of https://github.com/hotosm/tasking-manager/pull/5642.

eternaltyro:
@spwoodcock I think the changes are good to go from our end. Please let us know once you have resolved any conflicts and we'll give it a review again.

spwoodcock:
I'll be working on this now - will update soon.

Dockerfile
spwoodcock: If I may, I will update the Dockerfile a bit and base it on 3.10-alpine3.17.

spwoodcock:
As we are using Python image 3.10-alpine3.17, the system proj version is 9.x.
I had to update pyproj to 3.5.0.
As far as I can see there are few breaking changes: https://pyproj4.github.io/pyproj/stable/history.html#id13
so this shouldn't be an issue.

spwoodcock:
It's not possible for us to update to Python 3.10 due to the `schematics` dependency.
The `Iterable` class was removed in Python 3.10.

Traceback:
```
  File "/usr/src/app/manage.py", line 11, in <module>
    from backend.services.users.authentication_service import AuthenticationService
  File "/usr/src/app/backend/services/users/authentication_service.py", line 9, in <module>
    from backend.services.messaging.message_service import MessageService
  File "/usr/src/app/backend/services/messaging/message_service.py", line 12, in <module>
    from backend.models.dtos.message_dto import MessageDTO, MessagesDTO
  File "/usr/src/app/backend/models/dtos/message_dto.py", line 1, in <module>
    from schematics import Model
  File "/home/appuser/.local/lib/python3.10/site-packages/schematics/__init__.py", line 6, in <module>
    from . import deprecated
  File "/home/appuser/.local/lib/python3.10/site-packages/schematics/deprecated.py", line 8, in <module>
    from .types.serializable import Serializable
  File "/home/appuser/.local/lib/python3.10/site-packages/schematics/types/__init__.py", line 2, in <module>
    from .base import *
  File "/home/appuser/.local/lib/python3.10/site-packages/schematics/types/base.py", line 19, in <module>
    from collections import Iterable, OrderedDict
ImportError: cannot import name 'Iterable' from 'collections' (/usr/local/lib/python3.10/collections/__init__.py)
```

I will revert to Python 3.9.

spwoodcock:
I have updated the dockerfile and pyproject.toml, plus docs and CircleCI.
As well as PDM, I also made the container run as a non-root user for security.
Please review and let me know if I missed anything.

I think CircleCI is failing because there have been substantial changes to the CI config since I made this PR.
It seems to be running on the old config.yaml for CircleCI, not sure why (my version is in sync with `develop`, aside from two lines edited.

On a side note: does tasking manager currently run in a Kubernetes cluster, or with the docker-compose config?
If it runs with the docker-compose.yml in the root of the repo, then I would update the gunicorn `--workers` value to `4`.
If Kubernetes, then 1 worker is ideal (no change required).

spwoodcock:
With the currently dependency versions we can only use Python 3.9.
To upgrade to Python 3.10, I suggest we:

1) Merge this PR.
2) Selectively merge #5642:
  - All code edits.
  - Dependency versions, for compatibility with Python 3.10.
  - Skip edits to dockerfiles.
3) Import the new dependency versions using `pdm import -f requirements requirements.txt).
4) Remove `requirements.txt`.
5) Split out the dev dependencies `pdm add --dev ... ... ...`.

Dockerfile
```diff
@@ -1,12 +1,34 @@
-FROM quay.io/hotosm/base-python-image as base
-LABEL version=0.1
-LABEL maintainer="HOT Sysadmin <sysadmin@hotosm.org>"
-LABEL description="Builds backend docker image"
+ARG ALPINE_IMG_TAG=3.17
+ARG PYTHON_IMG_TAG=3.9
 
-WORKDIR /usr/src/app
+FROM docker.io/python:${PYTHON_IMG_TAG}-alpine${ALPINE_IMG_TAG} as base
```
eternaltyro, line 4: @spwoodcock Just for context, one of the reasons we switched from Docker to Quay was that Docker was rate limiting us. Our pipelines build docker images every push (probably a bad idea and will be fixed) and has been pulling from docker.io very often - hitting the rate limits. Just FYI, no action needed.

Dockerfile
```diff
@@ -20,29 +42,55 @@ RUN apk update && \
         proj-util \
         proj-dev \
         make
+# Setup backend Python dependencies
+COPY --from=extract-deps \
+    /opt/python/requirements.txt /opt/python/
+RUN pip install --user --no-warn-script-location \
```
eternaltyro, line 48: I'm assuming `--no-warn-script-location` because we are installing from `/opt/python/requirements.txt` which is not in PYPATH

Dockerfile
```diff
@@ -20,29 +42,55 @@ RUN apk update && \
...
-    -r requirements.txt
 
+FROM base as runtime
+ARG PYTHON_IMG_TAG
+WORKDIR /usr/src/app
+ENV PYTHONDONTWRITEBYTECODE=1 \
+    PYTHONUNBUFFERED=1 \
+    PYTHONFAULTHANDLER=1 \
+    PATH="/home/appuser/.local/bin:$PATH" \
+    PYTHON_LIB="/home/appuser/.local/lib/python$PYTHON_IMG_TAG/site-packages" \
```
eternaltyro, line 60: Does `/home/appuser` already exist?

Dockerfile
spwoodcock: I decided on that flag so long ago now I have forgotten why 😄 Presumably it's to suppress the warning, but I can't seem to reproduce a warning on the `python:3.9` image, so perhaps it's been updated since.

Dockerfile
```diff
@@ -1,12 +1,34 @@
...
+ARG PYTHON_IMG_TAG=3.9
 
-WORKDIR /usr/src/app
+FROM docker.io/python:${PYTHON_IMG_TAG}-alpine${ALPINE_IMG_TAG} as base
+ARG APP_VERSION=0.1.0
+ARG DOCKERFILE_VERSION=0.4.0
+ARG ALPINE_IMG_TAG
+ARG PYTHON_IMG_TAG
+ARG MAINTAINER=sysadmin@hotosm.org
+LABEL org.hotosm.tasks.app-version="${APP_VERSION}" \
```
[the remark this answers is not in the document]
spwoodcock: What environment are you running the pipeline? If using Github action, you can use variable `${{ github.ref_name}}`. Gitlab has something similar. If using Quay to build, then I honestly don't know.

eternaltyro (approved):
Good to go. @Aadesh-Baral Looks good to me. I have some questions for @spwoodcock . But regardless, I would like this merged. We can continue conversations post merge too. This should pave the way for #5642 to be reviewed and merged next.

cc: @ramyaragupathy .

Dockerfile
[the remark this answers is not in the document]
spwoodcock: I hardcoded the variables for now to make this merge-able. Passing in dynamic build args could be a later refinement 🤷

Dockerfile
spwoodcock: It is created below here: https://github.com/hotosm/tasking-manager/blob/7c6e2dc3a116a381cac3833b8d730cf479bf22ed/Dockerfile#L76

I do it in this order for caching reasons / image layering.

spwoodcock:
A note for the future:
- I kept Python with an alpine base as that was used before & I didn't want to be too obnoxious with my edits.
- However, Debian 11 would be a better base for a Python image, as Alpine isn't a great combo with Python: https://pythonspeed.com/articles/alpine-docker-python/

As we are close to the point of merging, I'm not suggesting we change now, but this could be a future PR.

Aadesh-Baral:
@spwoodcock Let's resolve the merge conflicts first, and then i will give this a final review.

eternaltyro:
> 
>     I kept Python with an alpine base as that was used before & I didn't want to be too obnoxious with my edits.
> 
>     However, Debian 11 would be a better base for a Python image, as Alpine isn't a great combo with Python: https://pythonspeed.com/articles/alpine-docker-python/

Agreed. Let's deal with this later.

eternaltyro:
@Aadesh-Baral  the PR removes `requirements.txt`. Is there a way we can negotiate this conflict from our end?

Aadesh-Baral:
In that case, lets update pyproject.toml with latest requirements.txt.

spwoodcock:
Let me know if there is anything I can do to help.

I could add the requirements file back in for now.

eternaltyro:
@spwoodcock Yes, please. Let's get this thing done.

Aadesh-Baral:
@spwoodcock  requirements.txt will not be required after this. I just want to have latest requirements.txt in pyproject.toml as we have updated some dependencies lately.

pyproject.toml
```diff
@@ -0,0 +1,94 @@
...
+    "Mako==1.2.2",
+    "Markdown==3.3.3",
+    "MarkupSafe==1.1.1",
+    "mccabe==0.6.1",
+    "mypy-extensions==1.0.0",
+    "newrelic==8.8.0",
+    "nose==1.3.7",
+    "oauthlib==3.2.0",
+    "pathspec==0.11.1",
+    "psycopg2==2.8.6",
```
Aadesh-Baral, line 53: 'pdm install' failed to install psycopg2 while setting up without docker. However, everything works fine with the docker-setup. Is this a problem with my system?

eternaltyro: @Aadesh-Baral what's the error message you get?

Aadesh-Baral: ERRORS:
add psycopg2 failed:
Traceback (most recent call last):
  File "/usr/lib/python3.8/concurrent/futures/thread.py", line 57, in run
    result = self.fn(*self.args, **self.kwargs)
  File "/home/aadesh/.local/lib/python3.8/site-packages/pdm/installers/synchronizers.py", line 236, in install_candidate
    self.manager.install(can)
  File "/home/aadesh/.local/lib/python3.8/site-packages/pdm/installers/manager.py", line 39, in install
    installer(str(prepared.build()), self.environment, prepared.direct_url())
  File "/home/aadesh/.local/lib/python3.8/site-packages/pdm/models/candidates.py", line 375, in build
    builder.build(build_dir, metadata_directory=self._metadata_dir)
  File "/home/aadesh/.local/lib/python3.8/site-packages/pdm/builders/wheel.py", line 30, in build
    filename = self._hook.build_wheel(out_dir, config_settings, metadata_directory)
  File "/home/aadesh/.local/lib/python3.8/site-packages/pyproject_hooks/_impl.py", line 209, in build_wheel
    return self._call_hook('build_wheel', {
  File "/home/aadesh/.local/lib/python3.8/site-packages/pyproject_hooks/_impl.py", line 311, in _call_hook
    self._subprocess_runner(
  File "/home/aadesh/.local/lib/python3.8/site-packages/pdm/builders/base.py", line 261, in subprocess_runner
    return log_subprocessor(cmd, cwd, extra_environ=env)
  File "/home/aadesh/.local/lib/python3.8/site-packages/pdm/builders/base.py", line 107, in log_subprocessor
    raise build_error(e) from None
pdm.exceptions.BuildError: Build backend raised error: Showing the last 10 lines of the build output:
It appears you are missing some prerequisite to build the package from source.

You may install a binary package by installing 'psycopg2-binary' from PyPI.
If you want to install psycopg2 from source, please install the packages
required for the build and try again.

For further information please check the 'doc/src/install.rst' file (also at
<https://www.psycopg.org/docs/install.html>).

error: command '/usr/bin/x86_64-linux-gnu-gcc' failed with exit code 1

eternaltyro: You might have to install `libpq-dev` in order to install from source. Can you try that?

Aadesh-Baral: Installing psycopg2-binary also does not resolve the problem.

Aadesh-Baral: `libpq-dev `already exists in my system.

spwoodcock: The Dockerfile uses Apline and install `postgresql-dev`, plus is running Python 3.10.

Looks like you are running Python 3.8, but from what distro of Linux?

I'm assuming Ubuntu, or Debian here.
The version of psycopg2 being installed is 2.8.6: https://packages.debian.org/bullseye/python3-psycopg2
So you will require libpq-dev v13.10: https://packages.debian.org/bullseye/libpq-dev

I would recommend attempting to replicate the environment of the docker build.
Run a `debian:bullseye` image and try to install again.

```bash
# Start with a fresh environment
docker run --rm -it -v $PWD:$PWD debian:bullseye bash
# apt install libpq-dev
apt update && apt install -y libpq-dev
# cd to your code
pip install pdm
pdm install
```

Hope this helps.

Dockerfile
eternaltyro: Yeah, that's a headache. I'll find a fix.

Dockerfile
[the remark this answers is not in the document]
eternaltyro: Makes sense. I will do this through circleCI parameters.

spwoodcock:
I just rebased with **develop** and added the `requirements.txt` file back in for now.

The file can be deleted after merging https://github.com/hotosm/tasking-manager/pull/5642 and updating the `pyproject.toml`.

Ready to merge 👍

eternaltyro:
Merged. All good so far. Thanks @spwoodcock for your time and labour. It's much appreciated 🙏🏼

pyproject.toml
Aadesh-Baral: Looks like installing psycopg2-binary==2.9.6 instead of psycopg2==2.8.6 will resolve this issue.

spwoodcock: I would advise against that & instead install the correct prerequisite libraries on your system 👍

 https://stackoverflow.com/questions/70330567/what-is-the-different-about-psycopg2-and-psycopg2-binary-python-package

Dockerfile
eternaltyro: New issue for this: https://github.com/hotosm/tasking-manager/issues/5788

Can we continue the discussion there?

eternaltyro:
PEP 582 has been rejected! - https://discuss.python.org/t/pep-582-python-local-packages-directory/963/430 

We will continue to use venv for now until our move to containers is done.

spwoodcock:
Yes it has, but that doesn't mean it's useless.
Here is a great summary: https://chriswarrick.com/blog/2023/01/15/how-to-improve-python-packaging/

PDM defaults to venv based installs. PEP582 support is optional, but is a much better alternative IMO.

The rejection shouldn't affect this PR.
PDM is used to export requirements for the dockerfile and to install via venv for running locally.

eternaltyro:
> Yes it has, but that doesn't mean it's useless. Here is a great summary: https://chriswarrick.com/blog/2023/01/15/how-to-improve-python-packaging/
> 
> PDM defaults to venv based installs. PEP582 support is optional, but is a much better alternative IMO.
>

I agree. But we have to play nice with standards and conventions that have buy in from the community also. 
 
> The rejection shouldn't affect this PR. PDM is used to export requirements for the dockerfile and to install via venv for running locally.

Agreed. I am only concerned about PEP582; the rest of the features PDM brings are still useful. Going forward, we will use venv, preferably managed by PDM and conforming to PDM defaults. All this is moot anyway with containers taking over.

PR #5187, opened 2022-06-19
Enable backend tests on circle ci

Aadesh-Baral:


Aadesh-Baral:
Thanks, @dakotabenjamin for setting up the test database. I can see some tests are failing I will work on that soon.

PR #5204, opened 2022-06-29
Show project title instead of project id in notifications

### LINKED ISSUE #2870: Improve message popup
* [ ] Link to project is not working well. A normal click doesn't work. On some browsers it can be opened with a right click, on others it is harder to get.
* [ ] Link doesn't jump to the comments section.
* [ ] Review `delete` button position
* [ ] Can we add more info on the project. The project title would give me much more context about this than just the project number.

image

when I open a message in my inbox that is a ‘mention’ type, I try clicking on the project where it originates but nothing happens. Right-clicking and opening in new tab does get me to the project but not the Q/A section.

Aadesh-Baral:
closes #2870 and #5203


Aadesh-Baral:
task_comment_project_id
invalidation_project_name
project_name_chat

PR #5207, opened 2022-06-30
Implement email templates from TM4 design

HelNershingThapa:
This PR has HTML email templates implemented from Figma.

@Aadesh-Baral take it from here and implement it on Jinja. Let me know if you need some help with HTML templating and styling.


Aadesh-Baral:
@HelNershingThapa just found out that many CSS properties like `display: flex, box-shadow` etc. are removed by email service providers as most of the email services still use HTML3. There might need some changes to the HTML template.

HelNershingThapa:
> @HelNershingThapa just found out that many CSS properties like `display: flex, box-shadow` etc. are removed by email service providers as most of the email services still use HTML3. There might need some changes to the HTML template.

sadge

Issue #5108, opened 2022-04-21
Feature request: Team Managers can enable notifications for Team join requests


SColchester:
**Is your feature request related to a problem? Please describe.**
When 'Invite only' is enabled for a team, team managers have to approve each join request that is made. At the moment the only way for a team manager to know that a request has been made is to regularly check the 'Manage Team' page and look for new requests. This is very time-consuming and may mean that users have to wait a long time until their request is approved. 

**Describe the solution you'd like**
It would be fantastic if there was an option to enable notifications on the 'Manage Team' page (ideally email notifications) so that each time an application to a team was made a notification is sent to the team manager(s). I suggest that by default this is toggled off, and if this were enabled for all teams on first being implemented it might upset lots of team managers who might never wish to have been disturbed. 

I've marked a green cross on a Team Manage page where I suggest the notification toggle is placed. It might have the text "Enable for team managers to receive (email) notifications each time a new join request is made":

![Inked2022-04-21 13_31_06-Window_LI2](https://user-images.githubusercontent.com/43822967/164451580-3785c026-96f7-460c-9694-0d45745d66f4.jpg)

**Describe alternatives you've considered**
Weekly roundup of join requests could be sent as a weekly notification instead.

ghimiresushma99:
Yes, the group join request email would be very much helpful for Team Manager and will save us much time. 

Since we already have a dedicated **Notifications** section in the **Settings** of **Tasking Manager**, I believe the option to enable this notification will fit in there better.

![image](https://user-images.githubusercontent.com/80664831/172123534-03a3bb3f-90fe-43e6-aa0b-c255199ede49.png)

Aadesh-Baral:
Closed by #5218

PR #5218, opened 2022-07-03
[rework] Added switch field to enable/disable notifications

Aadesh-Baral:
This PR is a rework on #5172 due to a change in the implementation of #5108.


HelNershingThapa (approved):
LGTM

PR #5220, opened 2022-07-04
Allow updating team member role

### LINKED ISSUE #5217: Cannot update team member role from member to manager and vice versa
**Describe the bug**
An error message is shown while changing the team member's role. Team members must be removed from the particular role and then added to the new one to update. If a team has only one member it's almost impossible to change their role. 

This has to do with the backend as the following error was thrown by the backend when trying to change user role
`"Error": " User is already a member of this team or has already requested to join",  "SubCode": "UserAlreadyInList"`

**To Reproduce**
You need to have one of the Admin, Organization admin, or team manager permission to reproduce this error.
1. Go to Manage
2. Click on Teams
3. Click on any team you manage
4. Add any user as a manager who is already a member of the team or vice versa
5. You will see the error as shown in the screenshot below

**Expected behavior**
Team member role must be changed instead of showing an error.

**Screenshots**
update team member role

Aadesh-Baral:
closes #5217

### DIFF
--- a/backend/models/postgis/team.py
+++ b/backend/models/postgis/team.py
@@ -44,6 +44,15 @@ class TeamMembers(db.Model):
         db.session.delete(self)
         db.session.commit()
 
+    def update(self):
+        """ Updates the current model in the DB """
+        db.session.commit()
+
+    @staticmethod
+    def get(team_id: int, user_id: int):
+        """ Returns a team member by team_id and user_id """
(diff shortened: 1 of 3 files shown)


Aadesh-Baral:
@HelNershingThapa Assigning to you as members and managers section also needs to be updated on frontend after updating role.

PR #5227, opened 2022-07-06
FIX Add two new SMTP vars

dakotabenjamin:
@eternaltyro needs review

### DIFF
--- a/scripts/aws/cloudformation/tasking-manager.template.js
+++ b/scripts/aws/cloudformation/tasking-manager.template.js
@@ -128,6 +128,16 @@ const Parameters = {
     Description: 'TM_SMTP_PORT environment variable',
     Type: 'String'
   },
+  TaskingManagerSMTPSSL: {
+    Description: 'TM_SMTP_USE_SSL environment variable',
+    Type: 'Number',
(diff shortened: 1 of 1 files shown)


eternaltyro (changes requested):
Can we make TM_SMTP_USE_SSL 1 by default?

eternaltyro:
For flask_mail

SSL with port 465  (STARTSSL??)
TLS with port 25 or 587 (TLS Wrapper??)


For AWS SES:

STARTSSL - 25, 587, or 2587
TLS Wrapper - 465 or 2465

Have to set ssl as enabled by default.

Both variables cannot be true at the same time.

Aadesh-Baral:
flask_mail has port as 25 and both use_tls and use_ssl as False by default.

dakotabenjamin:
After updating staging and pushing this hotfix to production, I have updated the defaults and added AllowedValues parameter as well.

Aadesh-Baral:
@dakotabenjamin is this complete? 
P.S. I am not receiving any emails from staging.

dakotabenjamin:
Waiting on review- will need to check staging again as well

PR #5245, opened 2022-07-13
Replace textfield with dropdown for searching by project ID for 'My Tasks' tab

### LINKED ISSUE #2173: My Contributions Page: search by project
Under my contribution section - there is an option to filter tasks by projects. Text box here takes in every keyboard stroke and performs an API query for each entry. Instead, this should pre-populate IDs based on the result already available in the page.

image

HelNershingThapa:
Closes #2173 

Screenshot from 2022-07-13 16-08-37

Uses the `projects/queries/{username}/touched/` endpoint to pre-populate dropdown options with project IDs


Ichchhie:
As shown in the screenshot I have multiple projects worked on but only 1 project id is listed in the dropdown. Which marks the need of further work @HelNershingThapa @Aadesh-Baral 
Screenshot_328

PR #5281, opened 2022-08-01
Fix readme

zlavergne:
This PR does 2 things:
1. fixes the paths in the README for the new docs structure (these will need to change again once the new doc content is finalized)
2. it adds two more necessary env vars for circle tests to pass. Most notably, [`test_variable_replacing` in `TestTemplateService`](https://github.com/hotosm/tasking-manager/blob/49632fbad1633effbc6237205271576ae01a9e60/tests/backend/unit/services/messaging/test_template_service.py#L19-L20) needs the org env vars to be able to pass


dakotabenjamin (approved):
Thank you @zlavergne for the contribution to our documentation!

PR #5304, opened 2022-08-14
Add rate limit to endpoints

### LINKED ISSUE #5122: Rate Limit API calls
There is currently nothing stopping someone from performing a DOS attack unintentionally by writing a script that GETs big projects (MBs of data, in some cases) in a loop and overloads the database connections. We would not know which user (even if they are using an authentication token!) is performing the requests, or have a way to stop it. This is a risk for reliability of the website. 

We would need to implement a rate limiting feature at least for the weightier requests. In addition we could also begin to attach user id info to requests as well in the logging system. 

I am wondering if a package like https://flask-limiter.readthedocs.io/en/stable/ would work

Aadesh-Baral:
closes #5122


ramyaragupathy:
@kaditya97 @prabinoid - something to consider as part of tech debt work? cc @manjitapandey

PR #5370, opened 2022-10-05
Feature / Auto Close Issues

### LINKED ISSUE #5369: How to archive multiple issues via script or automated way
I am adding this issue as a suggestion for Hacktoberfest as I wasn't able to find a solution easily..

Task: 
There are a lot of  issues in our backlog going back to 2017. We want to archive issues from 2017 to Jan 2020. However, before archiving and closing the issue we want to contact users (tag on Github) and give them two weeks notice to respond if they have any objections. We no response/ objection, we'd like to close the issue automatically. If response is provided by the person who added the issues, we will follow-up/discuss and address accordingly before closing. We discussed in one of our Tasking Manager Meet Ups and the group was in agreement that as long as we notify the person who raised the issue and give them enough time to response, we can proceed with archiving and closing the issues.

Solution:
Is there a way to automate the process vis script/another method so that we don't have to go to each issue one by one.

Thanks in advance!

kshitijrajsharma:
- Resolve #5369

### DIFF
--- a/.github/workflows/issue_checker.yml
+++ b/.github/workflows/issue_checker.yml
@@ -1,0 +1,22 @@
+name: Close inactive issues
+on:
+  schedule:
+    - cron: "0 0 * * *" # Runs every Every day at 0 UTC
+
+jobs:
+  close-issues:
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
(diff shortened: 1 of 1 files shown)


spwoodcock:
This is neat & should be merged after a cleanup of the current issues IMO.

ramyaragupathy:
I would not want to get this in at the moment

PR #5375, opened 2022-10-10
Add options to team join methods

### LINKED ISSUE #2913: Methods to join teams
We need more options for how one applies to/joins a team. 

1. Need a 'Group is public, but can join only by invite' - this is different than the current 'by invite' - what I mean is I want to only add people, not let them request to join (i.e. only those who receive party invitation can join). (i.e. similar to private team, but public so team can still be found, added to projects, etc. - i.e. super group of project managers - add us if you need help with projects, but only select people are invited to the group)
2. Then we turn the current 'by invite' into what it really is - 'by request'; i.e. user requests to join and an admin chooses accept/deny. 
3. Anyone can just join, no one is watching the door.
4. Follow link to a form/survey/website with further instructions; then an admin will add you if you 'pass the test' (i.e. this is how we would like to keep doing validator vetting, minus the form - acts similar to 1).

Aadesh-Baral:
Closes #2913.
This PR implements 3 team join methods for user to join a team. The team join method has to be set by team manager on team management page.
### Types of team join method:
Type ANY: This can be joined by anyone which already exists in TM. This team visibility will be Public.
Type BY_INVITE: This will be public/private group that can be set by manager and user cannot request to join this group, only managers can add. The user will not receive option to accept/reject the invite, but will get a notification that they have been added to the team.
Type BY_REQUEST: This will be current "invite_only" group that will be termed as "by_request" where users can request to join group and managers have option to accept/deny this request. This team visibility can only be Public.

_**On #2913 there was request to one more team type i.e. BY_REQUEST team but user must fill a form or have certain badge/level to make a join request. But this hasn't been addressed in this PR.**_
team_join_methods


HelNershingThapa (changes requested):
When a team's join method is set to _invite only_, the button to leave the team is hidden from view. Let's display this button.

A tooltip describing the two options under visibility would also be helpful, especially for managers.

frontend/src/components/teamsAndOrgs/teams.js
```diff
@@ -163,38 +172,56 @@ export function TeamInformation(props) {
...
+              width={12}
+              height={12}
+              className="blue-grey v-mid pb1 ml2"
+              data-tip={intl.formatMessage(messages["privateDescription"])}
+            />
+            <ReactTooltip place="bottom" className="mw6" effect="solid" />
+          </div>
         </div>
-      </div>
+      ) : null}
```
HelNershingThapa, line 224: Was there any reason to utilize a ternary operator instead of the logical && operator? For the conditional rendering part, I would have used the logical && rather than the ternary operator because the underlying component already renders something from the code above.

I assume we could also map these two visibility values from an array, similar to how you did with the join methods above this code.

frontend/src/components/teamsAndOrgs/management.js
```diff
@@ -45,10 +45,15 @@ export function VisibilityBox({ visibility, extraClasses }: Object) {
...
+  const teamJoinMethods = {
+    ANY: 'anyoneCanJoin',
+    BY_REQUEST: 'byRequest',
+    BY_INVITE: 'byInvite',
+  };
   return (
-    <div className={`tc br1 f7 ttu ba red b--red ${className}`}>
-      <FormattedMessage {...messages.inviteOnly} />
+    <div className={`tc br1 f7 ttu ba red b--red ${props.className}`}>
+      <FormattedMessage {...messages[teamJoinMethods[props.joinMethod]]} />
```
HelNershingThapa, line 56: You can also consider removing the message from the message file if it is no longer being utilized.

HelNershingThapa (approved):

PR #5378, opened 2022-10-12
Add feature to extend task lock time

### LINKED ISSUE #3539: Extend lock time endpoint
We have a timer on the map/validation page showing how much time the user has to finish mapping or validation, but it's only informative and we don't have any action triggered from there.

We could add an endpoint to extend the lock time, so we can show a modal to ask if the user wants to continue working on that task.

Other possibility is to ask the user to unlock the task and redirect them to the task selection page, so they can lock again if needed.

Aadesh-Baral:
closes #3539 

- [x] Add endpoint to extend task lock time.
- [x] Display popup on frontend  5-10 minutes before task lock time is expired from where user can extend lock time.
- [x]  Display popup on frontend after task lock time is expired from where user can relock tasks.


Aadesh-Baral:
@HelNershingThapa Found this issue while review: 
After a task expires, a popup appears from which the user can re-lock it; however, on the second expiry, a popup stating "Your task has been relocked" appears instead of the same popup from which the user can re-lock it. 
Screenshot from 2022-11-01 13-51-22

PR #5386, opened 2022-10-19
Implement a simple markdown editor with toolbar

### LINKED ISSUE #5130: Wysiwyg editor for comments boxes
Currently to format text in the tasks comments section of TM you have to use Markdown. That has a few disadvantages:
If you do not have some sort of developer background Markdown could be too hard to use.

There is not even a hint accompanying the text box that Markdown can be used. So those that know Markdown will not use it because they simply do not know they can.

The result of this is that all comments look like boring unformatted blocks of text, unattractive to read.

The ideal solution would be a text editor with a toolbar with buttons to render bold, italic and lists (numbered, not numbered) etc. 

**Additional context**
The idea was martien's

HelNershingThapa:
Closes #5130

Implements on four different sections:
- Project's _Questions and Comments_ section
- Task's status update section (both mapping and validation)
- Task history section (by clicking on the first icon on the tasks list)
- Message team members section

Screenshot (from Project's _Questions and Comments_ section)

questions-comments


wireguy:
@HelNershingThapa could a button be added for attach an image ?
(or make a new issue?)     It is  already available with drag and drop.

HelNershingThapa:
@wireguy sure, would you mind creating the issue?

martien-176:
https://github.com/hotosm/tasking-manager/pull/5104

Hi @HelNershingThapa there is also a not much known possibility to add Youtube (instruction) video's  in project's description and instructions.

Would that also be an option to add to the toolbar?

If yes i can make a new issue, or add that to @wireguy issue.

HelNershingThapa:
I concur. Thank you for bringing this to attention, @martien-176 . This is due to the fact that the textarea in the project's description and instructions section differs from the one we used in this pull request.

Regarding the YouTube symbol in the toolbar, we've already implemented it on the front end, the client side - you'll see a YouTube video when previewing it, but not when you post it. I've included a sneak peek, `::youtube[BxV14h0kFs0]`. 
Peek 2022-10-21 16-14

Because we use the same input for emails and the Tasking Manager, these implementations have yet to be improved so that they perform as expected - you'll have to wait until we have a workaround. I expect the editor will eventually be added to all of TM's major textareas.

HelNershingThapa:
@martien-176 @wireguy I just created an issue https://github.com/hotosm/tasking-manager/issues/5390 to track the enhancements for the editor.

PR #5399, opened 2022-11-04
Add tm version to db using github workflow

### LINKED ISSUE #5314: Add TM Version to database
When the database schema updates, it can break some third party queries. If the version is in the database then those users can check easily if the schema changes. 

Add a table with the current version of TM to the DB and update it when new versions come out. 

**Additional context**
Bene from Heigit requested this feature. We provide weekly database dumps (scrubbed user data) for their analyses. When we update the DB schema their processing software needs updating, so it would be helpful to have that information available on the database. 

It could also be used on the frontend to inform users of the current TM version.

Aadesh-Baral:
closes #5314.
This implementation creates a new github workflow which will be triggered when a new release is published. This workflow will makes a POST request to a new endpoint `https://tasking-manager-tm4-production-api.hotosm.org/api/v2/system/release/` which will fetch latest release from github and saves to database of production instance. Release tag name and published date will be saved to `release_version` table in the databse.

In terms other instances then production, a post request to `/system/release/` should be done manually to save a release on database.

Additionally the release vesion can be checked using `/system/heartbeat/` endpoint.


Aadesh-Baral:
@dakotabenjamin can you verify if this would be correct implementation?

.github/workflows/add-version-to-db.yml
```diff
@@ -0,0 +1,11 @@
...
+on:
+  release:
+    types: [published]
+jobs:
+  add_release:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Add release version to db
+      run: |
+        curl -X POST "https://tasking-manager-tm4-production-api.hotosm.org/api/v2/system/release/"
```
dakotabenjamin, line 11: Could someone abuse this POST request? I can see below that the request will update the release version from GitHub itself, so it would probably be difficult to accidentally or otherwise insert incorrect information here. What are your thoughts?

backend/api/system/general.py
```diff
@@ -227,3 +235,40 @@ def post(self):
...
+        )
+        try:
+            tag_name = response.json()["tag_name"]
+            published_date = response.json()["published_at"]
+            release = ReleaseVersion.get()
+            if release is None:
+                release = ReleaseVersion()
+            release.tag_name = tag_name
+            release.published_at = published_date
+            release.save()
```
dakotabenjamin, line 268: If the release hasn't changed when the request is made (say, by someone trying to call the function outside of the GH action), we should exit out rather than save over.

backend/api/system/general.py
```diff
@@ -227,3 +235,40 @@ def post(self):
...
+          - system
+        produces:
+          - application/json
+        responses:
+          201:
+            description: Saved version successfully to database
+          502:
+            description: Couldn't fetch latest release from github
+          500:
+            description: Internal server error
```
dakotabenjamin, line 255: Can we authenticate/authorize only certain roles or people/bots with keys to call this API and return 403 forbidden otherwise?

dakotabenjamin (changes requested):
I posted some questions in-line. Overall this looks good, with some improvements needed to prevent any possible edge cases.

.github/workflows/add-version-to-db.yml
Aadesh-Baral: As you mentioned this endpoint internally fetches the version from GitHub it will be not be possible to insert incorrect information here. So i don't think people can actually abuse this request. 

We can make this endpoint restricted to sys admin role only, but for that we need to generate token with no expiry or longer expiry date as our session token expires every 7 days. Since people can't abuse this request i think we can leave it open to all users.

backend/api/system/general.py
Aadesh-Baral: Sounds good

backend/api/system/general.py
Aadesh-Baral: We can make this endpoint restricted to sys admin role only, but for that we need to generate token with no expiry or longer expiry date as our session token expires every 7 days. Since people can't abuse this request i think we can leave it open to all users.

.github/workflows/add-version-to-db.yml
dakotabenjamin: OK, sounds good to me

dakotabenjamin (approved):

PR #5408, opened 2022-11-15
Clarify refresh update language

dakotabenjamin:
This changes the language of the update popup to force a user to reload their page. @d-rita I was unable to test getting the dialogue to show, but didn't get any build errors. How can we get a new service worker locally to test these changes?


d-rita:
The popup rendered looks like this:
update dialog
I think it would be better to center the `Refresh` button in the middle of the popup. 

@HelNershingThapa are you familiar with testing a service worker locally?

HelNershingThapa:
@d-rita You could try enabling the service worker flag from your environment variable `TM_ENABLE_SERVICEWORKER=1` and then serve the build. `yarn build && npx serve -s build`. maybe

dakotabenjamin:
Rather than center the button, moved the dialogue to two lines. I think that looks nicer that a centered button on a left-aligned block. Thoughts? 

BTW I still can only manage to get the popup to show by temporarily removing the conditionals- but none of the logic for showing the popup has changed so not sure if that needs testing. @HelNershingThapa suggestion did not work.

ramyaragupathy:
@royallsilwallz @varun2948 @manjitapandey

Issue #3918, opened 2020-11-23
Handle "enforce random task selection" option on task selection page


willemarcel:
We have an option on the project edit page that should enforce the task selection page to always select a random task to the user. That feature existed since TM3, but was not ported to TM4, so even when the project has that option activated, the task selection page is not enforcing a random task lock for the user.

It seems it's not used by HOT community, but only by some companies.

PR #5409, opened 2022-11-16
Handle random task selection enforcement on task selection page

HelNershingThapa:
Closes https://github.com/hotosm/tasking-manager/issues/3918

_Screenshot:_

alert component

### DIFF
--- a/frontend/src/components/taskSelection/index.js
+++ b/frontend/src/components/taskSelection/index.js
@@ -245,7 +245,10 @@ export function TaskSelection({ project, type, loading }: Object) {
           setTaskAction(getTaskAction(user, project, null, userTeams.teams, userOrgs));
         } else {
           // if there are multiple tasks selected, remove the clicked one
-          setSelectedTasks(selected.filter((i) => i !== selection));
+          const selectedTasksTemp = selected.filter((i) => i !== selection);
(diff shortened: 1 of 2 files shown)


d-rita (commented):
@HelNershingThapa, I tested the feature and it does work as expected, i.e. a user is assigned a random task when they click on the "map a task" button. The alert gives more clarity on what is happening with the project which is good. 
Maybe we can also show the particular assigned task on the task map since currently it is not outlined.
cc: @Ichchhie @ramyaragupathy

Ichchhie:
> @HelNershingThapa, I tested the feature and it does work as expected, i.e. a user is assigned a random task when they click on the "map a task" button. The alert gives more clarity on what is happening with the project which is good. Maybe we can also show the particular assigned task on the task map since currently it is not outlined. cc: @Ichchhie @ramyaragupathy

Imo, even at present, when we directly click `Map a task` button without selecting any particular task, it's randomly selected one. Once the task is submitted, we are also zoomed into submitted one in the task map. So, taking care of the feature not being that prominently used nowadays as mentioned in the [issue ](https://github.com/hotosm/tasking-manager/issues/3918) itself, I think having this message will be enough for now.

ramyaragupathy:
I get your point @Ichchhie - since the random task selection is already in place, a message suffices.

PR #5416, opened 2022-11-21
Use SwitchToggle to enable/disable comment preview

willemarcel:
image

I believe that style fits better with the app design system.

### DIFF
--- a/frontend/src/components/projectDetail/messages.js
+++ b/frontend/src/components/projectDetail/messages.js
@@ -133,13 +133,13 @@ export default defineMessages({
     id: 'project.detail.questionsAndComments.login',
     defaultMessage: 'Log in to be able to post comments.',
   },
-  write: {
-    id: 'project.detail.questionsAndComments.write',
-    defaultMessage: 'Write',
-  },
   preview: {
     id: 'project.detail.questionsAndComments.preview',
     defaultMessage: 'Preview',
+  },
+  previewComment: {
+    id: 'project.detail.questionsAndComments.previewComment',
+    defaultMessage: 'Preview comment',
   },
   post: {
     id: 'project.detail.questionsAndComments.button',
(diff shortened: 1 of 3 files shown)


emi420:
I think this solution is good as it puts the preview option closer to the "Post" button and it saves space, removing the "Write" tab which looks redundant.

I'll change the label to "Preview" and work in with the avatar to integrate it more into the comment box, for example:

Screenshot 2023-03-29 at 11 22 07

Ichchhie:
I like how @emi420 has moved the user image out from the markdown editor box. But regarding the placement of the preview button, I feel the current markdown we have i.e. in the screenshot below is okay. Now, the write and preview buttons are together which is also similar to the GitHub markdown editor, and keeping them together is more appropriate in terms of accessibility too Screenshot_926

willemarcel:
If you prefer a tab, it's ok, I would just recommend to improve the design. Right now, it seems like a dirty text floating above the text input.

Tabs need to connect with what is above or below it, like in
image

or in the tasks page
image


or in the github example
image

emi420:
Good analysis @HelNershingThapa ! I didn't realized that switching between write and preview also changes the position of the bottom section, that's definitely something that breaks usability.

I agree with @willemarcel that tabs design must be improved, because it doesn't look good right now. 

For making good design changes we must take into account the whole comment box and its context, as all UI elements are related, maybe we should create an issue for "comment box redesign" , what do you think? CC @ramyaragupathy

ramyaragupathy:
sounds good @emi420 - @royallsilwallz, medium priority, definitely on our to do list

ramyaragupathy:
cc @manjitapandey

royallsilwallz:
Hi @emi420 , I think you can add your design inputs to this anytime you are free so that we can move ahead with the implementation of this PR. We can also create a new issue ticket and discuss the new design there. What are your thoughts?
cc - @ramyaragupathy @manjitapandey

PR #5425, opened 2022-11-29
Uniform contributor count across project list and project detail views

### LINKED ISSUE #4899: Divergent Contributor Counts
Here is a screen capture of the Explore page on the Tasking Manager:

Explore

Looking at project 6665, you see "total contributors" with a value of 3573

Here is a screen capture of the very same project, in the project view, just moments later:

Project

Towards the bottom you see "contributors" preceded by the value of 1069

It is my assumption that on the Explore page, the value is actual "total contributions", i.e. each time the project is 'worked on', i.e. if I map a few tasks, then come back and validation a few task - that is 2 contributions, by 1 contributor. The value on the project page is I think actually showing "Unique Contributors". However we go about it, we need to clarify (and make sure we use correct terminology) so these either both show "Unique Contributors" or "Total Contribution Sessions" or whatever stat makes the most sense here.

d-rita:
- Fixes #4899


Aadesh-Baral (changes requested):
I think this does not resolve the issue because we are still considering only users who mapped, validated, or marked as bad imagery as contributors in the project detail page. We need to make changes on the project detail page instead because we decided to count every user who has locked a task as a contributor rather than just state change.
project_detail_contrib
explore_project_contrib

backend/services/stats_service.py
```diff
@@ -311,7 +335,14 @@ def get_user_contributions(project_id: int) -> ProjectContributionsDTO:
...
                     validated=r.validated,
-                    total=r.total,
+                    split=r.split,
+                    marked_bad_imagery=r.bad_imagery,
+                    invalidated=r.invalidated,
+                    total=r.mapped
+                    + r.validated
+                    + r.bad_imagery
+                    + r.invalidated
+                    + r.split,
```
Aadesh-Baral, line 345:
```suggestion
                    total=r.mapped + r.validated,
```
Remove the addition of bad_imagery, invalidated, and split tasks from the total count since this shows up on tasks validated and invalidated by a user in the frontend.
split_task_counted

d-rita: Implemented that.

Aadesh-Baral (approved):

PR #5439, opened 2022-12-03
fix JOSM remote control so url= is regular http(s) URL

mnalis:
Otherwise, as `imageryUrl` is (as required) in format like `tms[20]:https://tms.osm-hr.org/knin-2007/{zoom}/{x}/{-y}.png`, that value would be sent as `url` parameter to [JOSM remote control](https://josm.openstreetmap.de/wiki/Help/RemoteControlCommands#imagery) `imagery` load command, which would thus fail to load imagery (in current testing JOSM version 18583 at least)

This PR fixes it so `url=` parameter to remote control is regular http(s) URL like `https://tms.osm-hr.org/knin-2007/{zoom}/{x}/{-y}.png`, as it should be.

### DIFF
--- a/client/app/project/project.controller.js
+++ b/client/app/project/project.controller.js
@@ -1678,10 +1678,11 @@
                 hasImagery = true;
             }
             if (hasImagery) {
+                const fixedUrl = imageryUrl.substr(imageryUrl.indexOf(":")+1);
                 var imageryParams = {
                     title: encodeURIComponent('Tasking Manager Imagery - #' + vm.projectData.projectId),
                     type: imageryUrl.toLowerCase().substring(0, 3),
-                    url: encodeURIComponent(imageryUrl)
+                    url: encodeURIComponent(fixedUrl)
                 };
                 editorService.sendJOSMCmd('http://127.0.0.1:8111/imagery', imageryParams)
                     .catch(function() {


ramyaragupathy:
hey @mnalis - can you please share more details on your Tasking Manager instance? Looks like this PR is specific for v3.

mnalis:
Yes, that PR was made for `v3`, as that was what we were using (and where we had a problem) in 2022. 

That specific instance (https://tasks.osm-hr.org) had since been upgraded to `v4` (for other reasons, i.e. Oauth2) which (on a cursory glance) does not seem to have that problem  (although I haven't done extensive testing).

ramyaragupathy:
@mnalis - do you know of other v3 instances that we could use to test this behaviour? If none available, let's verify the behaviour in v4 and close it here

cc @royallsilwallz @manjitapandey

mnalis:
> do you know of other v3 instances that we could use to test this behaviour? If none available

As v3 only supports Oauth1, which is [no longer supported by openstreetmap.org](https://wiki.openstreetmap.org/wiki/2024_authentication_update), if there are any running TM3 instances left, you wouldn't be able to log in into them any more (unless someone forked the TM3 and backported Oauth2 support, of course)

mnalis:
@ramyaragupathy I'm appreciative of your support, but I'm not sure if using your energy on this PR is warranted any more since 1 July 2024 - when TM V3 was rendered **inoperative**. 


The fact that external imagery also doesn't work (which this PR addressed at the time) is of lesser importance now that V3 doesn't work **at all**.  Also, V3 seems not be updated anymore anyway (last commit seems to have been almost [5 years ago](https://github.com/hotosm/tasking-manager/tree/version-3.x)), so it seems somewhat unlikely that someone will implement Oauth2 for it and new V3 release made; and unless it happens, any other improvements (such as this one) on V3 branch are mostly pointless, as nobody would be able to use them.

So, in my humble opinion - all PRs/issues that relate to V3-only (like this one) and are not applicable to V4, should probably best closed as outdated. (but if you do see some use for V3, I'd be delighted to hear!)

ramyaragupathy:
@mnalis - thanks for sharing your thoughts. I am of the same opinion too, but was mindful of the time you have given here. Mainly due to OAuth, this change has to be archived now. Looking forward to collaborate further in the future!

Here's the [high level roadmap](https://github.com/hotosm/tasking-manager/discussions/6688) for your reference. Would be happy to hear your feedback on this when you have some time. Thank you!

Issue #5390, opened 2022-10-21
Enhancements for the markdown editor


HelNershingThapa:
Tracking anticipated enhancements for the markdown text editor here:

- [ ] Add support to upload image by selecting an image with a button on the toolbar (https://github.com/hotosm/tasking-manager/pull/5386#issuecomment-1286011769)
- [ ] [BUG] Default list is not shown when @ button is clicked on the toolbar
- [ ] Include support to add YouTube videos from the toolbar

martien-176:
@HelNershingThapa, doesn't the addition of a preview tab to the editor also belong to this list?

HelNershingThapa:
@martien-176 If you're talking about the editor in other sections besides the one mentioned in this pull request https://github.com/hotosm/tasking-manager/pull/5386, then it doesn't belong here. This editor differs from the ones found in the project edit sections. We'll indeed migrate the same editor to those sections in the coming months; thank you for your patience.

dakotabenjamin:
There is a bug while using the `#author` tag accidentally stylizing when it's the first text of the line: 
![image](https://github.com/hotosm/tasking-manager/assets/1847818/93bdca23-4fd6-4f7b-b0d3-09c44c251bc0)

https://hotosm.slack.com/archives/C319P09PB/p1696963590721069

PR #5453, opened 2022-12-06
Display default contributors list on markdown editor

d-rita:
- Related to  #5390; Show default list when @ button is clicked on the toolbar

_Progress_:
The current solution is adapted from the [library's example](https://github.com/zurb/tribute/blob/252ec344a75b624268aef1cdbf80afc30b3cbe2b/example/index.html#L295) on how to trigger the mention functionality. Still getting inconsistent displays of the tribute menu when @ icon is clicked vs typed in.

### DIFF
--- a/frontend/src/components/comments/commentInput.js
+++ b/frontend/src/components/comments/commentInput.js
@@ -80,19 +80,7 @@ export const CommentInputField = ({
         setComment(e.target.value);
       });
     }
-    // cleanup function
-    return () => {
-      mentionIcon.removeEventListener('mousedown', function (e) {
-        e.preventDefault();
-        // eslint-disable-next-line react-hooks/exhaustive-deps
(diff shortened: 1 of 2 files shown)


HelNershingThapa (changes requested):
I noticed that the comment was erased when I clicked the toolbar icon.

text-removed

HelNershingThapa (changes requested):
- The menu is rendered where the textarea container begins rather than where the text caret is. The menu should appear directly beneath the caret from where the icon is clicked.
options-should-render-below-expected-at
- Text should be added after the caret--it gets added at the end of the line.
mention-should-begin-where-the-caret-is
- The menu list should be hidden when a user clicks outside the menu container.

ramyaragupathy:
stale

PR #5087, opened 2022-04-09
Make docker defaults more useful for local testing setup and update documentation

SlowMo24:
This PR updates the docker documentation that seems to lag behind the docker setup development. It does so by adapting the docker-compose defaults as well as the settings (.env) defaults.

In addition it updates and extends the documentation. This documentation update is quite brisk, so feel free to adapt anything where you don't agree. Especially I was not sure what would be the 'best-practice' to get administrator privileges on the self hosted Tasking Manager and how detailed to describe the procedure.

closes #5008

PR #5463, opened 2022-12-09
Update backend dependency

Aadesh-Baral:
Updates Python to 3.10.8 and all other dependencies to  latest version including flask to 2.2.2

Simplifies docker-compose from PR #5087


eternaltyro:
Related in part to #5705

Aadesh-Baral:
Cosing this since #5642 has addressed this.

PR #5493, opened 2022-12-20
Circleci update resource class and limit cache for frontend deps

dakotabenjamin:
Two separate issues were preventing CircleCI builds from running properly. 
1. The CI container was running out of memory. To resolve, we increased the resource class for the container (4 -> 8 GB memory) 
2. The node-modules folder was being cached from a hotfix branch that didn't have the proper updates. We were too liberal in the caching rules, so we removed one of the keys for checking whether the pull from the cache.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,82 +3,84 @@ jobs:
   build:
     working_directory: /home/circleci/app
     docker:
-    - image: cimg/python:3.7.14-node
-      environment:
-        SQLALCHEMY_DATABASE_URI: postgresql://taskingmanager@localhost/test_tm
-    - image: cimg/postgres:14.2-postgis
-      environment:
-        POSTGRES_USER: taskingmanager
-        POSTGRES_DB: test_tm
+      - image: cimg/python:3.7.14-node
+        environment:
+          SQLALCHEMY_DATABASE_URI: postgresql://taskingmanager@localhost/test_tm
+      - image: cimg/postgres:14.2-postgis
+        environment:
+          POSTGRES_USER: taskingmanager
+          POSTGRES_DB: test_tm
     steps:
-    - checkout
-    - setup_remote_docker
-    - run:
-        name: Install modules
-        command: |
-          sudo apt-get update
-          sudo apt-get install -y libgeos-dev # Required for shapely
-          sudo yarn global add @mapbox/cfn-config @mapbox/cloudfriend
-          pip install awscli --upgrade
-    - run:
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -3,82 +3,83 @@ jobs:
...
-    - run:
-        name: Set folder permissions
-        command: |
-          chown -R circleci:circleci ${CIRCLE_WORKING_DIRECTORY}
-          chmod -R 755 ${CIRCLE_WORKING_DIRECTORY}
-    - restore_cache:
-        keys:
-        - node-v1-{{ .Branch }}-{{ checksum "frontend/yarn.lock" }}
-        - node-v1-{{ .Branch }}-
-        - node-v1-
```
dakotabenjamin, line 37: Removed this key

.circleci/config.yml
```diff
@@ -3,82 +3,83 @@ jobs:
...
+          path: tests/backend/results
+      - store_artifacts:
+          path: tests/backend/results
+      - save_cache:
+          key: node-v1-{{ .Branch }}-{{ checksum "frontend/yarn.lock" }}
+          paths:
+            - frontend/node_modules
+            - env
+    # The resource_class feature allows configuring CPU and RAM resources for each job. Different resource classes are available for different executors. https://circleci.com/docs/2.0/configuration-reference/#resourceclass
+    resource_class: large
```
dakotabenjamin, line 82: Added the resource class

.circleci/config.yml
```diff
@@ -88,262 +89,264 @@ jobs:
         description: "The 40 char hash of the git commit"
         type: string
     working_directory: /home/circleci/tasking-manager
+    resource_class: medium
```
dakotabenjamin, line 92: Added this resource class

.circleci/config.yml
```diff
@@ -88,262 +89,264 @@ jobs:
...
+          name: Configure AWS default region
+          command: |
+            aws configure set region $AWS_REGION \
+            --profile default
+      - run:
+          name: Deploy Frontend to S3
+          command: |
+            cd ${CIRCLE_WORKING_DIRECTORY}/frontend/
+            export TM_ENVIRONMENT=<< parameters.stack_name >>
+            yarn
```
dakotabenjamin, line 225: We should probably use the cache here instead of rebuilding.. maybe there is a risk of using the wrong cached folder though...

eternaltyro (approved):
LGTM

PR #5502, opened 2022-12-28
Revise styles for the user settings page

HelNershingThapa:
This PR revises overall styles for the user settings page. Two major changes here would be - reduction of the switch toggle component size and replacement of the locale selector component with that from react-select.

Large screens:
user-settings-large

Medium screens:
user-settings-medium

Small screens:
user-settings-small


Aadesh-Baral (commented):
Everything looks excellent, however I liked how each part was previously distinguished by a box-shadow. However, this is only my opinion.

PR #5605, opened 2023-02-27
Set mapping level upgrade notifications as system generated

Aadesh-Baral:
Addresses the issue raised in: #5567 
This PR sets mapping level upgrade notifications as system generated notification type and also modifies the notification templete of this notfication type.

Now the mapper level upgrade notifications will look like:
mapper-Level_upgrade-popup
mapper-Level_upgrade


Ichchhie:
LGTM

PR #5632, opened 2023-03-14
Add similar projects at the bottom of the project detail page.

### LINKED ISSUE #5626: Recommend related projects to user on project detail view
From discussion with @ramyaragupathy @HelNershingThapa @Ichchhie @d-rita 

**Create a frontend and backend to suggest related projects to users when they visit the project detail page.**

On the frontend, use the Figma design to display related projects at the bottom of the project detail page:
related_projects

Backend logic must be developed to identify similar projects using techniques such as cosine similarity or nearest neighbor.
This will be non-personalized recommendations, and every user, regardless of their activities or experience, will see the same result for a given project.

Aadesh-Baral:
closes #5626
This PR adds similar projects at the bottom of the project detail page.
realted_projects

**Implementation**:
- Similarity between the projects and given project will be calculated using cosine similarity and top 4 projects will be returned by default.
- Similarity between projects is calculated using these fields. -> `default_locale, difficulty, mapping_types, country, categories`
- All published projects except private ones are displayed  on non logged in view while only projects in which user has permission to map are shown for a logged in user.
- The function to create project matrix, which is the most significant part of calculating project similarity will be cached for a day. 

**How to test:**
- Install backend dependencies as new dependencies has been added in this PR.
- Start frontend and backend locally.
- Go to project detail page of any published projects
- Check if Similar projects button is visible on the footer alongside question & comments, contributors ....
- Click on similar projects button. You should be scrolled to Similar projects section.
- On the similar projects section 4 other similar projects should be displayed.


Aadesh-Baral:
To find similar projects, we analyzed the correlation between different project metadata fields and iterated through them. For instance, we observed that organizations tend to create similar projects based on their country and category. However, considering the organization for the similarity score led to the display of projects from the same organization, which affected the diversity of our displayed projects. Hence, we excluded organization as a field. Furthermore, we found some fields that didn't significantly contribute to our results, so we discarded them to enhance performance.
Then the result was tested with some test users who found it to be satisfactory.

Aadesh-Baral:
Adding @HelNershingThapa to review frontend changes and @d-rita to review backend changes.

HelNershingThapa (changes requested):
When you click the related project card, the page should scroll to the top so that users feel like they are viewing another project when they navigate to the related project. Use the useFetch hook to fetch linked projects as well, making the component look cleaner. Other than that, LGTM, I'll pushed with the requested changes.

HelNershingThapa (approved):

d-rita (commented):
@Aadesh-Baral will the limit of the projects returned always be 4? We have a limit parameter that we don't actually give the user a chance to use in the endpoint

backend/api/projects/resources.py
```diff
@@ -1241,3 +1242,52 @@ def get(self):
...
+              description: Base64 encoded session token
+              required: false
+              type: string
+              default: Token sessionTokenHere==
+            - name: project_id
+              in: path
+              description: Project ID to get related projects for
+              required: true
+              type: integer
+              default: 1
```
d-rita, line 1269: The limit parameter should be included in this part of the documentation

d-rita (approved):
Tested this, and it works as expected. The tests are good too. Overall, LGTM

ramyaragupathy (approved):
Looks good to me @Aadesh-Baral

PR #5636, opened 2023-03-20
Remove redundant TeamsRestAPI Post method

d-rita:
**What does this PR do?** 
Under the TeamsRestAPI, we have two methods doing the same job, i.e. POST and PATCH. This PR removes the redundant POST method that is not in use anywhere and is essentially a copy of what the PATCH method does. 
The PATCH method is retained since the RFC standards dictate that the PATCH method is what is used to update a resource and not POST.

### DIFF
--- a/backend/__init__.py
+++ b/backend/__init__.py
@@ -671,7 +671,7 @@ def add_api_endpoints(app):
     api.add_resource(
         TeamsRestAPI,
         format_url("teams/<int:team_id>/"),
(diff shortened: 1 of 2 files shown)


eternaltyro (approved):
Looks Good. How will this change the API version?

d-rita:
> Looks Good. How will this change the API version?

There might be no need to change the API version since the method itself was not in use at any point, i.e. not mapped to any endpoint.

Aadesh-Baral (approved):

PR #5641, opened 2023-03-22
Add Comments Module API Tests

d-rita:
This PR handles the Comments module-related API endpoint tests.

**How to test:**

- create a test database following the instructions [here](https://github.com/hotosm/tasking-manager/blob/develop/docs-old/setup-development.md#create-the-database-user-and-database). The database name is preceded by `test_`
- pull and checkout test branch using `git fetch origin && git checkout chore/add-comments-tests`
- run tests using `python3 -m unittest discover tests/backend`

### DIFF
--- a/backend/api/comments/resources.py
+++ b/backend/api/comments/resources.py
@@ -54,7 +54,7 @@ class CommentsProjectsRestAPI(Resource):
         """
         authenticated_user_id = token_auth.current_user()
         if UserService.is_user_blocked(authenticated_user_id):
-            return {"Error": "User is on read only mode.", "SubCode": "ReadOnly"}, 403
+            return {"Error": "User is on read only mode", "SubCode": "ReadOnly"}, 403
 
         try:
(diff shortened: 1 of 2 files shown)


tsmock:
Can you add an `__init__.py` file to the new test directory?

Almost every test directory has one (the only ones that don't were added in the past year).

You might be wondering _why_. `nose` (which is what the TM CI uses to [run tests](https://github.com/hotosm/tasking-manager/blob/develop/.circleci/config.yml#L67)) hasn't been updated since 2016, and completely fails to run on recent [python versions](https://github.com/nose-devs/nose/issues/1099). I'm currently working on updating pretty much all of the dependencies for the TM (see #5642), and I'm using `pytest` instead of `nose`. I would anticipate `nose2` having the same problem. And `nose2` recommends `pytest` over itself as well.

Aadesh-Baral (approved):

Issue #5483, opened 2022-12-15
[BUG] Error while login on instance setup using docker-compose


Aadesh-Baral:
/system/authentication/login throws error after the authorization on instance setup using docker. This issue is probably related to gevent and this issue occurs after the update of gevent and greenlet in #5316 which had to be done to resolve #5300.

Error log:
```
ERROR in app: Exception on /api/v2/system/authentication/callback/ [GET]
backend       | Traceback (most recent call last):
backend       |   File "/usr/local/lib/python3.7/site-packages/flask/app.py", line 1950, in full_dispatch_request
backend       |     rv = self.dispatch_request()
backend       |   File "/usr/local/lib/python3.7/site-packages/flask/app.py", line 1936, in dispatch_request
backend       |     return self.view_functions[rule.endpoint](**req.view_args)
backend       |   File "/usr/local/lib/python3.7/site-packages/flask_restful/__init__.py", line 468, in wrapper
backend       |     resp = resource(*args, **kwargs)
backend       |   File "/usr/local/lib/python3.7/site-packages/flask/views.py", line 89, in view
backend       |     return self.dispatch_request(*args, **kwargs)
backend       |   File "/usr/local/lib/python3.7/site-packages/flask_restful/__init__.py", line 583, in dispatch_request
backend       |     resp = meth(*args, **kwargs)
backend       |   File "/usr/src/app/backend/api/system/authentication.py", line 72, in get
backend       |     code=authorization_code,
backend       |   File "/usr/local/lib/python3.7/site-packages/requests_oauthlib/oauth2_session.py", line 350, in fetch_token
backend       |     **request_kwargs
backend       |   File "/usr/local/lib/python3.7/site-packages/requests_oauthlib/oauth2_session.py", line 522, in request
backend       |     method, url, headers=headers, data=data, **kwargs
backend       |   File "/usr/local/lib/python3.7/site-packages/requests/sessions.py", line 529, in request
backend       |     resp = self.send(prep, **send_kwargs)
backend       |   File "/usr/local/lib/python3.7/site-packages/requests/sessions.py", line 645, in send
backend       |     r = adapter.send(request, **kwargs)
backend       |   File "/usr/local/lib/python3.7/site-packages/requests/adapters.py", line 450, in send
backend       |     timeout=timeout
backend       |   File "/usr/local/lib/python3.7/site-packages/urllib3/connectionpool.py", line 710, in urlopen
backend       |     chunked=chunked,
[...]

Issue #3638, opened 2020-09-26
Postpone COPY in Dockerfile


frafra:
Executing `COPY` after `apt` would save bandwidth and time, as every time a file changes, the package manager has to install all the packages.

https://github.com/hotosm/tasking-manager/blob/d94ea226e2142da7e2b4b33755ab12ce41797458/scripts/docker/Dockerfile.backend#L13-L18

It would be even better just to run apt, then copy `requirements.txt`, running `pip` just after that and everything else later, to further optimize the image.

PR #5642, opened 2023-03-22
Update dependencies (backend)

### LINKED ISSUE #4906: Cannot use Docker due to tzlocal.utils.ZoneInfoNotFoundError
```
backend       | [2021-10-19 17:28:43 +0000] [9] [ERROR] Exception in worker process
backend       | Traceback (most recent call last):
backend       |   File "/usr/local/lib/python3.7/site-packages/gunicorn/arbiter.py", line 583, in spawn_worker
backend       |     worker.init_process()
backend       |   File "/usr/local/lib/python3.7/site-packages/gunicorn/workers/ggevent.py", line 162, in init_process
backend       |     super().init_process()
backend       |   File "/usr/local/lib/python3.7/site-packages/gunicorn/workers/base.py", line 119, in init_process
backend       |     self.load_wsgi()
backend       |   File "/usr/local/lib/python3.7/site-packages/gunicorn/workers/base.py", line 144, in load_wsgi
backend       |     self.wsgi = self.app.wsgi()
backend       |   File "/usr/local/lib/python3.7/site-packages/gunicorn/app/base.py", line 67, in wsgi
backend       |     self.callable = self.load()
backend       |   File "/usr/local/lib/python3.7/site-packages/gunicorn/app/wsgiapp.py", line 49, in load
backend       |     return self.load_wsgiapp()
backend       |   File "/usr/local/lib/python3.7/site-packages/gunicorn/app/wsgiapp.py", line 39, in load_wsgiapp
backend       |     return util.import_app(self.app_uri)
backend       |   File "/usr/local/lib/python3.7/site-packages/gunicorn/util.py", line 358, in import_app
backend       |     mod = importlib.import_module(module)
backend       |   File "/usr/local/lib/python3.7/importlib/__init__.py", line 127, in import_module
backend       |     return _bootstrap._gcd_import(name[level:], package, level)
backend       |   File "<frozen importlib._bootstrap>", line 1006, in _gcd_import
backend       |   File "<frozen importlib._bootstrap>", line 983, in _find_and_load
backend       |   File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked
backend       |   File "<frozen importlib._bootstrap>", line 677, in _load_unlocked
backend       |   File "<frozen importlib._bootstrap_external>", line 728, in exec_module
backend       |   File "<frozen importlib._bootstrap>", line 219, in _call_with_frames_removed
backend       |   File "/usr/src/app/manage.py", line 77, in <module>
backend       |     cron = BackgroundScheduler(daemon=True)
backend       |   File "/usr/local/lib/python3.7/site-packages/apscheduler/schedulers/base.py", line 87, in __init__
backend       |     self.configure(gconfig, **options)
backend       |   File "/usr/local/lib/python3.7/site-packages/apscheduler/schedulers/base.py", line 126, in configure
backend       |     self._configure(config)
backend       |   File "/usr/local/lib/python3.7/site-packages/apscheduler/schedulers/background.py", line 29, in _configure
backend       |     super(BackgroundScheduler, self)._configure(config)
backend       |   File "/usr/local/lib/python3.7/site-packages/apscheduler/schedulers/base.py", line 697, in _configure
backend       |     self.timezone = astimezone(config.pop('timezone', None)) or get_localzone()
backend       |   File "/usr/local/lib/python3.7/site-packages/tzlocal/unix.py", line 205, in get_localzone
backend       |     _cache_tz = _get_localzone()
backend       |   File "/usr/local/lib/python3.7/site-packages/tzlocal/unix.py", line 163, in _get_localzone
backend       |     tzenv = utils._tz_from_env()
backend       |   File "/usr/local/lib/python3.7/site-packages/tzlocal/utils.py", line 125, in _tz_from_env
backend       |     ) from None
backend       | tzlocal.utils.ZoneInfoNotFoundError: 'tzlocal() does not support non-zoneinfo timezones like %s. \nPlease use a timezone in the form of Continent/City'
backend       | [2021-10-19 17:28:44 +0000] [9] [INFO] Worker exiting (pid: 9)
backend       | [2021-10-19 17:28:44 +0000] [1] [INFO] Shutting down: Master
backend       | [2021-10-19 17:28:44 +0000] [1] [INFO] Reason: Worker failed to boot.
```

tsmock:
This PR intends to make the TM compatible with the defaults of Debian Bookworm. This means that the backend needs to be compatible with Python 3.11.

In general, the code changes are intended to be limited to compatibility changes. In the event that there needs to be a reformat due to formatting tool updates, those will be in separate commits (to make rebasing onto a new develop HEAD easier).

This PR replaces backend PRs #5649, #5629, #5591, #5590, #5563, #5480, #5457, #5420, #5385, #5348, #4591, and #4588.
EDIT: Additional backend PRs that are replaced are: #5770, #5761 (1.26.9 -> 1.26.15; sentry-sdk [does not](https://github.com/getsentry/sentry-python/issues/2070) support 2.x), #5777, and #5747.

I highly recommend using the `Hide Whitespace` option in the `Diff view` settings.

EDIT: This also fixes #4906 and #5483 (see the `Fix gevent warning when run with gunicorn` commit).


Aadesh-Baral:
Hello, @tsmock. Thank you for all your efforts. We were also working on updating backend dependencies on #5463; just to let you know if you are still working on backend dependencies and on updating React to 18, there is another PR #5510 that was halted due to some bundle size issues. Also it would be easy for reviewer if we could separate frontend and backend dependencies update PRs.

tsmock:
I did see #5510 -- I haven't seen the same bundle size issues (the largest I saw was 3 MB, prior to removing polyfills).

I _did not_ see #5463 .

I can go ahead and split the frontend/backend PRs.
I'll look through #5463 and cherry-pick some changes I didn't already have (I'd be surprised if #5463 passes the test suite right now -- I ran into issues with newer libraries having a max precision of 6 decimal places by default).

Aadesh-Baral:
#5463 is still a draft and yet to be finalized. There are also some changes to docker and circle configs in that PR which are not discussed yet, so might not need to be cherry picked them.

tsmock:
There are some things that it looks like I need to check/look into:

* `gevent.monkey.patch_ssl()` (AKA, I need to try to run my branch in a container to see if I need to add that)
* `TIMEZONE = os.getenv("TIMEZONE", "UTC")` (I didn't need this in a dev environment, maybe in a prod environment?)
* `accepted_licenses = db.relationship("License", secondary=user_licenses_table, overlaps="users")` (I don't think this is needed, but I could be wrong -- I'll check with the backend running via gunicorn)

tsmock:
> * `gevent.monkey.patch_ssl()` (AKA, I need to try to run my branch in a container to see if I need to add that)

This turned out to be necessary to silence a gevent warning.
> * `TIMEZONE = os.getenv("TIMEZONE", "UTC")` (I didn't need this in a dev environment, maybe in a prod environment?)

I haven't figured out why this was in the other PR -- I did find a separate issue with the docker image (`ENV KEY VALUE # COMMENT` can be interpreted as `ENV KEY="VALUE # COMMENT"` which can cause issues).
> * `accepted_licenses = db.relationship("License", secondary=user_licenses_table, overlaps="users")` (I don't think this is needed, but I could be wrong -- I'll check with the backend running via gunicorn)

This is why I converted it back to a draft PR. I cherry-picked the migration from #5463, and then ran into `sqlalchemy.exc.ProgrammingError: (psycopg2.errors.DuplicateTable) relation "idx_projects_centroid" already exists`. I think this is something that is automatically added in a prior migration, but while doing a `flask db downgrade`/`flask db upgrade` to figure out _when_ it was added, I found some migrations that don't work both ways (on a populated database).

tsmock:
> [I then] ran into `sqlalchemy.exc.ProgrammingError: (psycopg2.errors.DuplicateTable) relation "idx_projects_centroid" already exists`. I think this is something that is automatically added in a prior migration, but while doing a `flask db downgrade`/`flask db upgrade` to figure out _when_ it was added, I found some migrations that don't work both ways (on a populated database).

It turns out that geoalchemy2 has some [recommended](https://geoalchemy-2.readthedocs.io/en/latest/alembic.html) changes for usage with alembic.

I've adapted the https://github.com/hotosm/tasking-manager/blob/develop/migrations/env.py file to match alembic best practices, while keeping the current (custom) code.
I've additionally added an ignore method for `idx_username_lower` since that was explicitly added in https://github.com/hotosm/tasking-manager/blob/develop/migrations/versions/a7c617755721_.py.

I've done some manual testing as follows, watching container logs (run via gunicorn) as follows:
* Create a new project
* Publish the project by adding required information (mostly `Test <field>`)
* Create an organization
* Create a license
* Create a campaign
* Create a category
* Do a bit of mapping in the new project
  * This involved marking projects as completed, not completed, bad imagery, splitting, and validating (note: it is apparently possible to self-validate if the task is selected via the task list instead of map, this is probably not intended)

tsmock:
There is a reason why I explicitly said to look at the PR with the `Hide whitespace` option enabled -- that is all churn from updating `black`, and was done in a separate commit. That takes the number of files to review from 136 to 81 (40% reduction in files to review).

Of the remainder, the following can be ignored:
* .github/dependabot.yml -- this was removing older versions from the ignore lists

The following categories are for files with single-type changes (so the file only has the _specified_ change and no more, besides whitespace).

These files were adapted for flask changes (and only flask changes):
* `send_file` `attachment_filename` -> `download_name` parameter name change
  * backend/api/projects/actions.py
  * backend/api/tasks/resources.py
* `python manage.py` to `flask`
  * docker-compose.yml
  * docs-old/migration.md
  * docs-old/review-pr.md
  * docs-old/setup-development.md
  * docs/developers/development-setup.md
  * scripts/aws/cloudformation/tasking-manager.template.js
  * scripts/database/export-import-projects/README
  * scripts/database/migration-from-tm2-postgres.sql
  * scripts/install/install_on_debian_10_buster.sh
  * scripts/install/install_on_ubuntu_18_04.sh (this is a bit of a cheat -- I also corrected `Assamble` to `Assemble` in a comment)
These files were adapted for sql changes (and only sql changes):
* `paginate` argument changes
  * backend/models/postgis/project_chat.py
  * backend/services/team_service.py
  * backend/services/validator_service.py
* Wrap raw SQL with `sqlalchemy.text` (usually seen as `sa.text`)
  * migrations/versions/0a6b82b55983_.py
  * migrations/versions/14340f1e0d6b_.py
  * migrations/versions/7937dae319b5_.py
  * migrations/versions/7bbc01082457_.py
  * migrations/versions/7d55a089b5bc_.py
  * migrations/versions/84c793a951b2_.py
  * migrations/versions/a43b9748ceee_.py
  * migrations/versions/bfcf4182dcb5_.py
  * migrations/versions/c40e1fdf6b70_.py
  * migrations/versions/deec8123583d_.py
  * migrations/versions/f86698c827cc_.py
These files were adapted for geojson changes (and only geojson changes):
* `geojson.is_valid` -> `obj.is_valid` and geojson error logging
  * backend/models/postgis/task.py
  * backend/services/project_admin_service.py
* `geojson` now trims to 6 decimal places by default
  * tests/backend/helpers/test_files/clipped_feature_collection.json (note: this file doesn't follow the geojson spec)
  * tests/backend/helpers/test_files/non_square_split_results.json
  * tests/backend/integration/services/test_mapping_service.py (caused test hashes to change, I did validate the gpx string prior to changing hashes) 
These files were adapted to account for equal shapes but with different coordinate orders (`str(shape)` -> `shape`, using `assertDeepAlmostEqual` which was moved to the base class from resources.py):
* tests/backend/unit/services/grid/test_grid_service.py
* tests/backend/integration/api/projects/test_actions.py
* tests/backend/integration/services/grid/test_split_service.py

That leaves ~30 files that need more in-depth examination, although most of those are also going to have one or more of the above changes as well. One that is fairly common in the files calling sql is `with db.engine.connect() as conn: conn.execute(<something>)`. The `db.engine.execute` command was deprecated, and that replaces it.

Files that can probably be ignored:
* tests/backend/helpers/test_files/feature_collection.json -- removes 

Files that should definitely be looked at (non-trivial or data sensitive):
* `migrations/` (anything non-trivial regarding migrations should always be looked at)
  * migrations/alembic.ini -- just adds a manually added index to ignore for generating migrations
  * migrations/env.py -- adaptions so that erroneous migrations are _less likely_ to be generated with geoalchemy2
  * migrations/versions/0eee8c1abd3a_.py -- `if not project_centroid` was added since during testing and migrating up/down the db path, there were issues.
  * migrations/versions/42c45e74752b_.py -- new migration
* manage.py -- the compatibility code for `python3 manage.py runserver` specifically
* requirements.txt -- the whole point of this PR
* tests/backend/base.py -- moved assertDeepAlmostEqual from test_resources.py and added a section for checking shapes

.circleci/config.yml
```diff
@@ -179,7 +180,7 @@ jobs:
     working_directory: /home/circleci/tasking-manager
     resource_class: large
     docker:
-      - image: cimg/python:3.7.14-node
+      - image: cimg/python:3.10.10-node
```
Aadesh-Baral, line 183: Since cimg/python:3.10.10-node users node v18 frontend build might not pass until frontend is compatible with node v18. Let's use a different images for frontend and backend jobs here.

.env
```diff
@@ -0,0 +1 @@
+FLASK_APP=manage.py
```
Aadesh-Baral, line 1: Let's add a comment here on why this .env file is required and mention all other variables other than FLASK _APP needs to be set in tasking-manager.env file.

.circleci/config.yml
```diff
@@ -7,10 +7,10 @@ jobs:
   build:
     working_directory: /home/circleci/app
     docker:
-      - image: cimg/python:3.7.14-node
+      - image: cimg/python:3.10.10-node
         environment:
           SQLALCHEMY_DATABASE_URI: postgresql://taskingmanager@localhost/test_tm
-      - image: cimg/postgres:14.2-postgis
+      - image: cimg/postgres:15.2-postgis
```
Aadesh-Baral, line 13: Is this version of Postgresql tested?

.github/dependabot.yml
Aadesh-Baral, line 1: Can these changes be done in separate PR?

backend/services/organisation_service.py
```diff
@@ -195,9 +195,7 @@ def get_organisation_stats(organisation_id: int, year: int) -> OrganizationStats
         ).filter(Project.organisation_id == organisation_id)
         if year:
             start_date = f"{year}/01/01"
-            projects = projects.filter(
-                Project.created.between(start_date, datetime.today())
-            )
+            projects = projects.filter(Project.created.between(start_date, func.now()))
```
Aadesh-Baral, line 198: If this is not related to dependency updates and needs change. Let's do this in another PR.

backend/services/stats_service.py
```diff
@@ -547,10 +546,14 @@ def get_task_stats(
                 TaskHistory.action_text,
                 func.DATE(TaskHistory.action_date).label("day"),
             )
+            .order_by(
+                TaskHistory.project_id,
+                TaskHistory.task_id,
+                TaskHistory.action_text,
+                TaskHistory.action_date,
+            )
```
Aadesh-Baral, line 554: Same here.. If this is not related to dependency updates and needs change. Let's do this in another PR.

.circleci/config.yml
tsmock: It looks like I can use cimg/python:3.10.7-node -- that image has node node 16.17.0.

.github/dependabot.yml
tsmock: Yes, they could be. All I did was remove versions older than what has been updated. And removed a duplicate section.

backend/services/organisation_service.py
tsmock: This was actually needed to fix a test. I didn't check to see if the test was failing prior to this PR, but I can check that.

backend/services/stats_service.py
tsmock: Works for me. I'll check to make certain that it doesn't break tests first though.

tsmock:
> Hi @tsmock As an initial assist, can you please split `8d29a635b4cf0880b5d566980a1c9dfc02be2ed4` out into its own PR?

I can just drop 8d29a635b4cf0880b5d566980a1c9dfc02be2ed4  -- it was the result of running `black manage.py migrations backend tests`.

backend/services/organisation_service.py
tsmock: Without that change, `test_get_all_org_includes_stats_if_omit_stats_set_false` in tests/backend/integration/api/organisations/test_resources.py fails. I can drop it, if that is still desired.

.circleci/config.yml
eternaltyro: I don't know how relevant this is, but PostgreSQL 15 has now security restricted the usage of public schema willy-nilly. It's a good thing but we just have to make sure it doesn't affect the app. I would recommend staying with the latest v14 which currently is v14.7. 

I will defer to @Aadesh-Baral for the final decision.

tsmock: In my testing, this did not appear to be an issue.

Specifically, I ran the tests against a database created with the following parameters:
* Docker images: `cimg/postgres:15.2-postgis` and `postgis/postgis:latest`
* POSTGRES_PASSWORD=hottm
* POSTGRES_USER=hottm
* POSTGRES_DB=test_tasking-manager

Aadesh-Baral: Let's use 14.7 for the time being and upgrade to v15 later after investigating the issue @eternaltyro mentioned.

backend/services/organisation_service.py
Aadesh-Baral: In that case, let's keep this change.

.circleci/config.yml
tsmock: I've moved back to 14.7 for now.

backend/services/organisation_service.py
tsmock: I've reworded the commit in question.

Aadesh-Baral: Thanks @tsmock.

spwoodcock:
Not 100% sure where this should live (if there is an issue or PR tracking the update to Python 3.10 mentioned by @eternaltyro.

But #5725 is currently blocking the upgrade. We can use Python 3.9 for now.

tsmock:
This PR updates schematics from 2.1.0 to 2.1.1 (see [requirements.txt#L34](https://github.com/hotosm/tasking-manager/pull/5642/files#diff-4d7c51b1efe9043e44439a949dfd92e5827321b34082903477fd04876edb7552R34)). But I wasn't able to reproduce using python 3.11 and schematics==2.1.0.

Looking at the stacktrace from #5725 , it _should_ be happening _reliably_ whenever the `schematics` library is imported. I wasn't able to reproduce with this branch. Can you (@spwoodcock) try to reproduce with this branch?

tsmock:
> `tzlocal.utils.ZoneInfoNotFoundError: 'tzlocal() does not support non-zoneinfo timezones like UTC # Fix timezone (do not change - see issue #3638). \nPlease use a timezone in the form of Continent/City'`

This should be fixed by b19e537d10d6661429b9b257eef1d07e008717de, but it looks like I missed `scripts/docker/Dockerfile.backend` (I assume that is the dockerfile you used). I'll fix it in a few minutes and rebase onto current develop.

spwoodcock:
Good stuff, thanks - I will test again later.

The backend is going through refactoring in various places - I significantly updated the dockerfiles in #5184, but I'm sure they will be consolidated eventually 👍

spwoodcock:
Seems that updating to schematics==2.1.1 solved the issue, for the imports in **manage.py** that is (during the app init).
I haven't been able to test the full suite of features on tasking-manager to assess if there are any other lurking issues.

I now get an issue with the `geojson` package and Python 3.10, so the dependencies definitely need a major overhaul.
Perhaps it's ok to update to Python 3.10 in #5184 after this PR has been reviewed and merged.

We still need to replace `schematics` soonish, as it's unmaintained (#5725) - compatibility will certainly break with Python 3.11.

eternaltyro:
Now that #5184 is merged, I'm eyeing this next. Reviewing this will take some time and back-and-forth communication. But the good news is that @ramyaragupathy - the project manager for TM - has carved out an entire sprint focused mostly on fortifying foundations and reviewing community contributions. So it shouldn't be too long before this is merged.

On cursory glance, the backend changes seem fine. I will do a more thorough review early next week. Some initial notes: 

- install_on_ubuntu_18_04.sh can be repurposed for maybe 20.04 or better, 22.04. - 18 is going into security-only-maintenance mode in two weeks. So not much point supporting it at this juncture.
- If we can wrangle schematics 2.1.1 to work with Python 3.10, then I think we should do that and worry about 3.11 later. Thoughts?
- `.circleci/config.yml` is going to be in flux as I fix/improve the config. If having to resolve conflicts often is a bother, then I can take any suggestions and make those changes myself if you prefer. 
- Since venv was replaced with PEP582 some changes to binary paths need to be replaced with local dir paths.

ad astra!

tsmock:
> * install_on_ubuntu_18_04.sh can be repurposed for maybe 20.04 or better, 22.04. - 18 is going into security-only-maintenance mode in two weeks. So not much point supporting it at this juncture.

I would agree with that. I didn't update the install_on_ubuntu_18_04.sh script to 22.04 since I figured that ought to be a separate PR (and I don't know if anyone actually _uses_ it).

> * If we can wrangle schematics 2.1.1 to work with Python 3.10, then I think we should do that and worry about 3.11 later. Thoughts?

In my testing schematics 2.1.1 has worked with Python 3.10 and Python 3.11. At the very least, all tests are passing with Python 3.11. I had done some more in-depth (end-to-end) testing with Python 3.10.

> * .circleci/config.yml is going to be in flux as I fix/improve the config. If having to resolve conflicts often is a bother, then I can take any suggestions and make those changes myself if you prefer.

I can go ahead and leave it alone. I don't mind rebasing/resolving conflicts often. I've been doing that on a weekly basis with this PR already. :)

> * Since venv was replaced with PEP582 some changes to binary paths need to be replaced with local dir paths.

I'll check this out.

eternaltyro:
> > * install_on_ubuntu_18_04.sh can be repurposed for maybe 20.04 or better, 22.04. - 18 is going into security-only-maintenance mode in two weeks. So not much point supporting it at this juncture.
> 
> I would agree with that. I didn't update the install_on_ubuntu_18_04.sh script to 22.04 since I figured that ought to be a separate PR (and I don't know if anyone actually _uses_ it).
> 

You are right. Let's do it in a separate PR.

> > * If we can wrangle schematics 2.1.1 to work with Python 3.10, then I think we should do that and worry about 3.11 later. Thoughts?
> 
> In my testing schematics 2.1.1 has worked with Python 3.10 and Python 3.11. At the very least, all tests are passing with Python 3.11. I had done some more in-depth (end-to-end) testing with Python 3.10.
>

TBD
 
> > * .circleci/config.yml is going to be in flux as I fix/improve the config. If having to resolve conflicts often is a bother, then I can take any suggestions and make those changes myself if you prefer.
> 
> I can go ahead and leave it alone. I don't mind rebasing/resolving conflicts often. I've been doing that on a weekly basis with this PR already. :)
> 

It's up to you.
 
> > * Since venv was replaced with PEP582 some changes to binary paths need to be replaced with local dir paths.
> 
> I'll check this out.

Thanks, merging #5184 introduced a few conflicts that you need to resolve.

tsmock:
> Thanks, merging #5184 introduced a few conflicts that you need to resolve.

Sorry, I thought I pushed the updated branch. I'll push it in a few minutes.

eternaltyro:
PEP582 has been rejected! - https://discuss.python.org/t/pep-582-python-local-packages-directory/963/430

We will continue to use venv for now.

tsmock:
I thought that was known. It was literally the first thing I found when I was looking into how to use it with PDM (see https://pdm.fming.dev/latest/usage/pep582/ ). In my documentation changes, I ended up using `pdm run ${command}` which _should_ work regardless of what installation method was used.

eternaltyro:
> I thought that was known. It was literally the first thing I found when I was looking into how to use it with PDM (see https://pdm.fming.dev/latest/usage/pep582/ ). In my documentation changes, I ended up using `pdm run ${command}` which _should_ work regardless of what installation method was used.

I missed that somehow. Anyway, we will continue using PDM's default mode which uses venv.

tsmock:
It looks like I have to fiddle with the CI changes a bit. Sorry about the noise.

scripts/docker/tasking-manager/Dockerfile
```diff
@@ -1,10 +1,11 @@
-FROM python:3.7-buster
+ARG PYTHON_IMG_TAG=3.10
+FROM python:${PYTHON_IMG_TAG}-bullseye
```
eternaltyro, line 2: Maybe in a different PR, but if we can get away with using  `python:3.10-slim-bullseye` for this, we will save ourselves a lot of security headaches later.

tests/backend/base.py
```diff
@@ -9,31 +14,89 @@ def clean_db(db):
...
 
     def tearDown(self):
         super(BaseTestCase, self).tearDown()
         self.db.session.rollback()
         self.db.session.close()
         self.app_context.pop()
+
+    # Code modified from https://github.com/larsbutler/oq-engine/blob/master/tests/utils/helpers.py
+    # Note: Was originally in test_resources.py (author Aadesh-Baral)
+    def assertDeepAlmostEqual(self, expected, actual, *args, **kwargs):
```
eternaltyro, line 64: nice

tests/backend/integration/api/projects/test_resources.py
```diff
@@ -259,14 +259,16 @@ def assert_project_response(project_response, expected_project, assert_type="ful
...
+                geojson.dumps(project_response["areaOfInterest"])
+            ).is_valid
             assert ["type", "coordinates"] == list(
                 project_response["areaOfInterest"].keys()
             )
             if assert_type == "notasks":
                 assert "tasks" not in project_response
             else:
-                assert geojson.is_valid(project_response["tasks"])
+                assert geojson.loads(geojson.dumps(project_response["tasks"])).is_valid
```
eternaltyro, line 271: I observed this earlier. What does this accomplish?

eternaltyro (approved):
@tsmock Firstly, thank you for this PR, it is significantly impactful work and I can only imagine the amount of time and effort you put into this. 

I did my best to review the PR. Everything looks good, and I think this is good to merge. I have added some comments and questions. Please address them when you can. I am approving this change as is. 

@Aadesh-Baral I added some flags for you as well to double check some things I am don't feel confident reviewing. Please go through them when you can. All that notwithstanding, this PR is good to merge. 

I reviewed test cases only superficially. But they look good. I did not look at the JSON files at all. For the rest, I was as thorough as I can be with what little Python skills I have.

eternaltyro:
Sonalcloud is flagging the fact that we are using a Docker image that runs things as root, we can create a separate issue to track that. Let's ignore it for this review.

eternaltyro:
@tsmock Output of functional tests run by CI - https://gist.github.com/eternaltyro/3b86de6467df37b766816221802c1b64

tsmock:
> @tsmock Output of functional tests run by CI - https://gist.github.com/eternaltyro/3b86de6467df37b766816221802c1b64

I know that a test is failing (see the edit in https://github.com/hotosm/tasking-manager/pull/5642#issuecomment-1553416994). I've been trying to reproduce locally so I can fix the test failure. Unfortunately, that test keeps passing locally. I've been trying to replicate the test environment such that I can debug the failing test.

backend/models/postgis/project.py
```diff
@@ -267,11 +267,18 @@ def set_country_info(self):
             current_app.config["OSM_NOMINATIM_SERVER_URL"], lat, lng
         )
         try:
-            country_info = requests.get(url).json()  # returns a dict
+            response = requests.get(url)
+            response.raise_for_status()
```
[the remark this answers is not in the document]
tsmock: This happened when I was debugging a flaky test. It turns out the test is hitting the Nominatim servers directly, which can be flaky at times. The test really ought to be mocking the response from Nominatim, but I felt that was out of scope for this PR.

migrations/versions/14340f1e0d6b_.py
```diff
@@ -20,9 +21,11 @@ def upgrade():
     tm3_pm_team = "TM3-project-managers"
     conn = op.get_bind()
     # Create an undefined organisation
-    conn.execute("insert into organisations (name) values('" + tm3_org_name + "');")
+    conn.execute(
+        sa.text("insert into organisations (name) values('" + tm3_org_name + "');")
+    )
     fetch_org_id = "select * from organisations where name = '" + tm3_org_name + "';"
-    org_id = [r[0] for r in conn.execute(fetch_org_id)][0]
+    org_id = [r[0] for r in conn.execute(sa.text(fetch_org_id))][0]
```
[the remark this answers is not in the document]
tsmock: Short answer: Probably.

Longer answer: There are a bunch of things in the `migrations` directory that could be done better/clearer. I felt that changing a bunch of migrations would be out of scope for this PR. I tried really hard to keep the diff in the `migrations` directory as small and as semantically equivalent as possible.

scripts/docker/tasking-manager/Dockerfile
tsmock: We'd probably want to use `python:3.11-slim-bookworm` (Debian Bookworm is being released on June 10th), assuming all tests pass.

But I want to say someone was looking at reworking the `Dockerfile`s at some point. As a specific example, I don't think this Dockerfile is being used (`RUN curl -sL https://deb.nodesource.com/setup_10.x | bash -`). I think the `Dockerfile.{backend|frontend}` is what is used elsewhere.

tests/backend/integration/api/projects/test_resources.py
tsmock: `geojson.is_valid` is no longer a function.
The code could just as easily be written as `geojson.loads(json.dumps(project_response["tasks"])).is_valid`. Essentially, I'm converting a `json` object into a `geojson` object.

tsmock:
Can someone check and see if c3bd7b8cf22f462370dd261e049df9606b474c32 is needed for other people? It might just be a configuration issue on the CI end (from [docs/sysadmins/ci-cd.md](https://github.com/hotosm/tasking-manager/blob/develop/docs/sysadmins/ci-cd.md), "[each] environment has its own set of environment variables which are stored as secrets in the CircleCI Organization Settings under Contexts.")

scripts/docker/tasking-manager/Dockerfile
eternaltyro: Re: bookworm, agreed. 


Re: Dockerfile, yes, a review / rewrite is due. I think we have ignored frontend Dockerfile for a while mostly because it doesn't affect our own deployments. We deploy to AWS S3/Cloudfront currently. I'll make a new issue for it.

eternaltyro:
Merged. The changes will be deployed to staging. I'll report back if there are any problems.

scripts/docker/tasking-manager/Dockerfile
eternaltyro: Incidentally, tomorrow is the date for a feature freeze - https://lists.debian.org/debian-devel-announce/2023/04/msg00007.html

eternaltyro:
FWIW, CircleCI is all green - 

image

eternaltyro:
We need `newrelic==8.8.0` for our deployments - introduced again here: https://github.com/hotosm/tasking-manager/pull/5824/commits/85dc839621d9a226f83404e9c94936e83b603d3d

tsmock:
> I'm not sure. I'll find out. Is that what was causing the backend test failures?

Yes, that is what was causing the most persistent backend test failure, and it was a pain to track down since it would pass if `ORG_NAME` or `ORG_CODE` was set in an environment file (which was set in my standard dev environment via a `tasking-manager.env` file).

> We need newrelic==8.8.0 for our deployments

I probably would have done the `newrelic` install at https://github.com/hotosm/tasking-manager/blob/develop/scripts/aws/cloudformation/tasking-manager.template.js#LL383C21-L383C23 instead of adding it as a dependency in the pyproject.toml file.

eternaltyro:
> I probably would have done the `newrelic` install at https://github.com/hotosm/tasking-manager/blob/develop/scripts/aws/cloudformation/tasking-manager.template.js#LL383C21-L383C23 instead of adding it as a dependency in the pyproject.toml file.

That's fair. I'll make the change.

eternaltyro:
Sentry SDK now supports urllib3 >= 2.x https://github.com/getsentry/sentry-python/releases/tag/1.25.0

PR #5644, opened 2023-03-24
Fix project completion issues with project summary caching

### LINKED ISSUE #5578: [BUG] Validate a task button is shown even when there are no tasktos to validate in a project
**Describe the bug**
Validate a task button is shown even when there are no any tasks to validate in a project. And even after selecting a task validate a task button is not changed to  validate/map selected task button. But while clicking on button it right error message is shown i.e. task state is not ready for validation which is right.

The `/tasks` endpoint is returning task status as READY and same is the case for tasks latest activity endpoint i.e. `/latest`.
**Expected behavior**
If all tasks are validated in a project validate a task button should be disabled and if there are tasks that are available for mapping the button should be change to map a task. If a task is selected the button should be changed to map/validate selected task based on the task status. These expected behavior happens most of the time but seems to break in certain situations. Need to re-examine logic for this button text thing.

**Screenshots**
validate_task

Aadesh-Baral:
closes #5578 and #3841

**#5578**
With this PR, we moved the function used to compute project completion out of cache, allowing it to be updated in real time. This will fix a number of side effects, including the issue described in #5578 and the difference in completeness between the project card and the project detail page.

**#3841**
Furthermore, the function to compute project completion has been set as an instance method rather than a static method, so that all project task statistics do not need to be supplied as arguments to calculate project completion statistics. When this is set as the instance method, all of the project tasks' statistics can be retrieved via the project instance itself.

How to test **locally**? _**(Caution: Do not do this in live instance)**_
- Go to any tasks view of any project. i.e. `/projects/:id/tasks` and note the project completion percentages.
- Go to project edit.
- Go to action.
- Click on map all tasks.
- Go to tasks view of same project. The project completion should indicate that project is 100% mapped. This should display live completion stats.

### DIFF
--- a/backend/models/postgis/project.py
+++ b/backend/models/postgis/project.py
@@ -660,27 +660,9 @@ class Project(db.Model):
             .filter(ProjectChat.project_id == self.id)
             .count()
         )
-        project_stats.percent_mapped = Project.calculate_tasks_percent(
-            "mapped",
-            self.total_tasks,
-            self.tasks_mapped,
-            self.tasks_validated,
-            self.tasks_bad_imagery,
-        )
-        project_stats.percent_validated = Project.calculate_tasks_percent(
-            "validated",
-            self.total_tasks,
-            self.tasks_mapped,
-            self.tasks_validated,
-            self.tasks_bad_imagery,
-        )
-        project_stats.percent_bad_imagery = Project.calculate_tasks_percent(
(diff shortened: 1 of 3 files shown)


Aadesh-Baral:
@d-rita Updated the testing guidelines.

d-rita (approved):

PR #5651, opened 2023-03-29
Use previously used color for mapped progress bar

HelNershingThapa:
The color for the mapped progress bar was revised from the Figma design in #5516. Following the [remarks on Slack](https://hotosm.slack.com/archives/C319P09PB/p1680076639458289), this PR reverses the change to use the previously used color `b--blue-grey` instead of `bg-mask`.

bg-mask bg-blue-grey


emi420:
I think that, for being accessible to everyone, including the color blind, it should have better contrast. Maybe a stripped line could be better.

Screenshot 2023-03-29 at 10 04 02

PR #5666, opened 2023-04-04
Fix contribution stats

d-rita:
This PR fixes issue #4899 on divergent contributor counts on the project detail and project list views. 

**How to test:**
- fetch and checkout branch using `git fetch origin && git checkout fix/contribution-stats`
- install dependencies
- start both the frontend and backend servers
- Go to the http://127.0.0.1:3000/explore to see project list. Note the contributor counts of different projects.
- Click on different projects to compare the contributor counts.

### DIFF
--- a/tests/backend/integration/api/projects/test_contributions.py
+++ b/tests/backend/integration/api/projects/test_contributions.py
@@ -1,7 +1,7 @@
 from datetime import datetime
 
 from tests.backend.base import BaseTestCase
-from tests.backend.helpers.test_helpers import create_canned_project
+from tests.backend.helpers.test_helpers import create_canned_project, return_canned_user
 from backend.models.postgis.task import Task, TaskStatus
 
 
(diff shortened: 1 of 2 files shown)


Aadesh-Baral (commented):
@d-rita The main issue raised here is that when a task is invalidated, it is still counted in the contributions section as mapped tasks for a user who mapped the task. [[More info](https://hotosm.slack.com/archives/C319P09PB/p1680342103558969)]

During my investigation, I found that the task is counted as mapped for the user who mapped it the most recently, indicating that the issue you mentioned does not exist. can you verify this?

> When a user invalidates a task, and it is mapped by another, the stats do not reflect the new mapper as the contributor to the project.

Aadesh-Baral (commented):
The issue with contributor statistics seems to to be resolved, but the issue with divergent contributor counts between project card and project detail remains. We can leave the issue with divergent contributor counts for now and target the next release since we’re targeting this fix for deployment on April 12th and will freeze the push to deploy branch from today.

To avoid complications, we can simply undo the changes made in #5425.

Aadesh-Baral (approved):
LGTM. The number of contributors appears to be consistent across project card and project detail page. Also, the statistics in the user contributor section appear to be unaffected by the change.

tests/backend/integration/api/projects/test_contributions.py
Aadesh-Baral, line 1: Do you think tests cover all the cases related to contributor counts and contributor stats?

ramyaragupathy (approved):

PR #5670, opened 2023-04-05
Switch to OIDC to authenticate CircleCI to AWS

eternaltyro:
CircleCI used to authenticate to AWS using API key pair credentials. This proved insecure due to exposures in breaches. This commit initiates the move from API Key based approach to OpenID Connect (OIDC) based approach to authenticate to AWS from CircleCI

Circle-CI requires that jobs run within contexts for OIDC auth to work. I am using a global context for common variables like AWS Account ID in addition to existing contexts specific to instances of Tasking Manager.

TODO:
- Investigate and move *all* workflows (that connect with AWS) to OIDC
- Move workflows to the top and add semantic names for each workflow
- Only trigger CI jobs when appropriate file paths are changed

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,8 @@
 version: 2.1
+
+orbs:
+  aws-cli: circleci/aws-cli@3.1
+
 jobs:
   build:
     working_directory: /home/circleci/app
@@ -94,6 +98,11 @@ jobs:
     steps:
       - checkout
       - setup_remote_docker
+      - aws-cli/setup:
+          role-arn: "arn:aws:iam::$ORG_AWS_ACCOUNT_ID:role/CircleCI-OIDC-Connect"
+          profile-name: "OIDC-Profile"
+          role-session-name: "backend-deploy"
+          session-duration: "1800"
       - run:
           name: Install modules
           command: |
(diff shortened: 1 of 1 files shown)


dakotabenjamin (commented):
Love the simplicity of it. Are you intending to apply the TODO items within this PR or in a later one? I'll hold on a full review if you plan to add more.

eternaltyro:
> Are you intending to apply the TODO items within this PR or in a later one?

@dakotabenjamin in a later one from a different branch. This is just to serve as a reminder.

.circleci/config.yml
```diff
@@ -195,27 +187,11 @@ jobs:
...
-      - run:
-          name: Configure AWS default region
-          command: |
-            aws configure set region $AWS_REGION \
-            --profile default
+      - aws-cli/setup:
+          role-arn: "arn:aws:iam::$ORG_AWS_ACCOUNT_ID:role/CircleCI-OIDC-Connect"
+          profile-name: "OIDC-Profile"
+          role-session-name: "frontend-deploy"
+          session-duration: "1800"
```
Aadesh-Baral, line 194: Session duration is set to 1800 on frontend and backend deploy jobs. Will this be enough? Since we create a db snapshot during backend deployment.

eternaltyro: Thank you for the review, @Aadesh-Baral . I see in past runs that backend deploy takes less than 30min. But I have  increased the duration to 45min anyway to include a buffer.

eternaltyro:
Related to #5539

PR #5686, opened 2023-04-12
Fix OpsGenie HTTP 422 error

eternaltyro:
Removed OpsGenie notification block since the curly braces added in an earlier commit did not fix the problem. We can bring back the notification once Opsgenie ORB is able to work as expected.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -331,6 +331,3 @@ workflows:
           context:
             - org-global
             - tasking-manager-tm4-production
-notify:
-  webhooks:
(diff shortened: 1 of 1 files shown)


dakotabenjamin:
@AfiMaameDufie has been working on this issue since a few weeks. Afi can you give an update on the current status of this issue?

AfiMaameDufie:
Currently, the Opsgenie orb doesn't work properly.
The suggestion from their support team is to add the notify statements to the individual jobs rather than the workflow.
Will try that out and let you know 
@dakotabenjamin

eternaltyro:
Related to #5539

PR #5687, opened 2023-04-12
Enhance/remove codedeploy pipelines

eternaltyro:


dakotabenjamin:
I'd like to keep the backend_deploy job named "test" as a generic code block that we could use for future deployment testing. What I have done in the past is change the filter for the branch from which it is triggered to the branch I am working on, spin up a `tasking-manager-test` environment on cloudformation, and then push to that branch to test the deployment and update processes. 

The rest of the code changes are good to remove.

eternaltyro:
The `test` job would be recreated in my workflow refresh eventually. I'm trimming it down so I have a minimal working config to build on top of. @dakotabenjamin

eternaltyro:
Related to #5539

PR #5689, opened 2023-04-12
Parameterise Backend AMI and Instance Type

eternaltyro:
In the CloudFormation template we manage, we had hardcoded the instance type of the backend EC2 VM and the AMI. This change parameterises it so that the values can be updated later by passing it to cfn-config

- Update PostgreSQL version from 11.12 to 11.19 - this change has already been made in RDS manually. This is a noop change.
- Update AMI to the latest Ubuntu 20.04. This will cause instances to be replaced.
- Update instance type to c6a.large. This will make our instances cheaper and more performant.


dakotabenjamin (approved):
LGTM

PR #5693, opened 2023-04-12
Add IPv6 route to Frontend CloudFront CDN

eternaltyro:
After enabling IPv6 support on the CDN itself, we need to create DNS AAAA records to be able to reach the frontend CDN using IPv6. This change creates the resource records in AWS Route53 - the DNS service for hotosm.org

Replaces a RecordSet resource with a RecordSetGroup resource


dakotabenjamin (approved):

eternaltyro:
Thanks @dakotabenjamin .

PR #5704, opened 2023-04-13
Syntactic sugar - simplify template

eternaltyro:


scripts/aws/cloudformation/tasking-manager.template.js
```diff
@@ -75,7 +75,7 @@ const Parameters = {
   },
   ELBSubnets: {
     Description: 'ELB subnets',
-    Type: 'String'
+    Type: "List<AWS::EC2::Subnet::Id>"
```
dakotabenjamin, line 78: Does this change the format of the parameter input?

dakotabenjamin (approved):

scripts/aws/cloudformation/tasking-manager.template.js
eternaltyro: No, it doesn't. The only difference is that Cloudformation automatically interprets `"subnet-abcd1234,subnet-efef1234"` as `["subnet-abcd1234", "subnet-efef1234"]`.

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html#aws-specific-parameter-types

PR #5708, opened 2023-04-14
Wait for RDS resources to become available

eternaltyro:
This change introduces wait-to-become-available logic  before performing snapshot copy or snapshot creation for RDS Database Servers


dakotabenjamin (approved):
LGTM

Issue #5442, opened 2022-12-05
Complete endpoint level testing 


Aadesh-Baral:
The Tasking Manager API endpoints aren't being tested. As a result, any dependency updates or major changes to the existing code are extremely challenging. To avoid problems in these instances, we need to have at least complete tests at endpoint level.

- [x] Projects #5489 
- [x] Users #5597 
- [x] Tasks #5572 
- [x] Organisations #5570 
- [x] Comments #5641 
- [x] Notifications #5715 
- [x] Teams #5598 
- [x] System #5776
- [x] Interests #5616 
- [x] Campaigns #5592 
- [x] Licences #5638
- [x] ~Grid~
- [x] ~Annotations~
- [x] Countries #5637 
- [x] ~Favourites~
- [x] Issues #5665
- [x] ~Mapping~

ramyaragupathy:
@Aadesh-Baral - in order to prioritise the endpoints, let's take data on query insights from the RDS. Working w/ @dakotabenjamin to see if this could be retrieved as a sql dump for us to map to specific endpoints.

Otherwise as discussed let's moved the `api` folder to `integration tests` section.

Aadesh-Baral:
@ramyaragupathy yes sure query insights will help. I've moved `api` folder to `integration` tests section on #5489.

Aadesh-Baral:
closing this as all of the tasks in this issue are complete.

PR #5676, opened 2023-04-07
Add feature to mark all notifications as read

Aadesh-Baral:
closes #5593 

Major changes to the frontend:
- Adds a new component `ActionButtons`, that combines actions for deleting and marking multiple notifications as read.
- Adds a new component, `SelectAllNotifications`, that appears when all notifications in a page are selected and more notifications are under the active category tab. This component allows users to select all notifications under the active notification category and perform actions such as deleting and marking multiple notifications as read.
- Implemented refetching of notifications in the notification popup component when new notifications are available for the user. This change may potentially solve the issue mentioned in https://github.com/hotosm/tasking-manager/issues/5717
- Adds test cases for the newly introduced components.

PR #5715, opened 2023-04-18
Add notification API tests

d-rita:
This PR adds tests related to the Notifications API endpoints. 

**Related issue:**
- #5442

**How to test:**
- create a test database following the instructions [here](https://github.com/hotosm/tasking-manager/blob/develop/docs-old/setup-development.md#create-the-database-user-and-database). The database name is preceded by `test_`
- pull and checkout test branch using `git fetch origin && git checkout chore/add-notifications-tests`
- run tests using `python3 -m unittest discover tests/backend`

### DIFF
--- a/tests/backend/integration/api/notifications/test_resources.py
+++ b/tests/backend/integration/api/notifications/test_resources.py
@@ -1,18 +1,22 @@
-from datetime import datetime
+from datetime import datetime, timedelta
 
 from tests.backend.base import BaseTestCase
+from backend.models.postgis.message import MessageType
 from tests.backend.helpers.test_helpers import (
     create_canned_user,
     generate_encoded_token,
     return_canned_user,
     create_canned_message,
     create_canned_notification,
+    create_canned_project,
 )
 
 TEST_SUBJECT = "Test subject"
 TEST_MESSAGE = "This is a test message"
 MESSAGES_NOT_FOUND = "No messages found"
 NOT_FOUND = "NotFound"
+OLDER_TEST_SUBJECT = "Older Test Subject"
+OLDER_TEST_MESSAGE = "This is an older test message"
 
 
 class TestNotificationsRestAPI(BaseTestCase):
@@ -143,7 +147,7 @@ class TestNotificationsAllAPI(BaseTestCase):
         self.assertEqual(response.status_code, 401)
         self.assertEqual(response_body["SubCode"], "InvalidToken")
 
-    def test_get_messages_returns_200(self):
+    def test_get_messages_no_query_params_returns_200(self):
         """
         Test that endpoint returns 200 when authenticated user accesses their messsage notifications
         """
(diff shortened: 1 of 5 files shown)


Aadesh-Baral:
@d-rita is this ready for review?

d-rita:
> @d-rita is this ready for review?

@Aadesh-Baral not yet, two failing tests for the get and post unread count api endpoints

tests/backend/integration/api/notifications/test_resources.py
```diff
@@ -0,0 +1,237 @@
...
+    def test_get_message_notifications_returns_401(self):
+        """
+        Test that endpoint returns 401 when an unauthenticated user accesses messsage notifications
+        """
+        response = self.client.get(self.url)
+        response_body = response.get_json()
+        self.assertEqual(response.status_code, 401)
+        self.assertEqual(response_body["SubCode"], "InvalidToken")
+
+    def test_get_messages_returns_200(self):
```
Aadesh-Baral, line 146: Also, test all of the query parameters to ensure that they all have the expected impact on the response.

Aadesh-Baral:
@d-rita What's the status of this PR?

d-rita:
@Aadesh-Baral this is ready for review.
- I have included tests for the query parameters attached to the NotificationsAllApi.
- I have also updated the message_type filter implementation for a user to enter a string value of the message type, i.e. `/api/v2/notifications/?messageType=system`

Aadesh-Baral (commented):
Test cases for query param sortBy and sortDirection seems to be missing for NotificationsAllAPI. Other than that LGTM.

backend/services/messaging/message_service.py
```diff
@@ -706,8 +706,17 @@ def get_all_messages(
...
+                key.strip().lower() for key in message_type.split(",")
+            ]
+
+            filter_values = [
+                key.value
+                for key in MessageType
+                if key.name.lower() in message_type_filters
+            ]
+
+            query = query.filter(Message.message_type.in_(filter_values))
```
Aadesh-Baral, line 719: Can you explain this change?

d-rita: While testing, I could not retrieve any notifications if I specified the message type. Not when I used numbers (for the mapped integer value assigned to a type under the MessageType enum class), e.g. 1 (system), 2 (broadcast)., etc. Nor when I typed in the string value itself, e.g. system/SYSTEM/System. That was with the previous implementation.

So, this new implementation lets a user retrieve the specific messages when they filter by the string value of the message type, e.g. system, mention_notification, broadcast, etc. 

Could you test to confirm if my findings were accurate or not? @Aadesh-Baral

Aadesh-Baral: This change has introduced a problem on the frontend. In our frontend code, we have four notification types that are associated with different combinations of MessageTypes. You can find more information about this issue in the following link: [GitHub Issue Comment](https://github.com/hotosm/tasking-manager/pull/5676#issuecomment-1506579190).

For instance, when applying a teams filter in the frontend, a query parameter is passed as "types=6,7,11" in the NotificationsAllAPI. To filter and retrieve the notifications for the provided message types on the backend, this value needs to be split by comma (',') and converted to integers.

However, due to this change, the value is being split correctly but is not being converted to integers, resulting in empty results being returned.

d-rita: Noted. I have adjusted the tests, removed this change, and added tests for the sortBy and sortDirection query parameters. 
Sidenote: For future reference, we should probably include in the documentation the expected values for the message_type filter and what type the respective numerical values correspond to.

Aadesh-Baral: > For future reference, we should probably include in the documentation the expected values for the message_type filter and what type the respective numerical values correspond to.

Sounds good

Aadesh-Baral: > For future reference, we should probably include in the documentation the expected values for the message_type filter and what type the respective numerical values correspond to.

@d-rita Let's do this in this PR itself.

Aadesh-Baral (approved):

PR #5720, opened 2023-04-19
Use window object to generate copy URLs for tasks and organizations

### LINKED ISSUE #5719: [BUG] Task link copy generates incorrect url
Reported by Kraan64 through slack -

**Describe the bug**
Clicking on the link button like in the example below does no longer give a usable link like in the past but undefined/projects/14513/tasks/?search=1797 . Please repair it. In the past i could use it to include it in a mail to the mapper. No longer....

**Screenshots**
image

HelNershingThapa:
This PR does the following: 
- Closes #5719 . Use the browser window object instead of the React Router location object to generate task and organization URLs. The values of the location object must have changed, making the origin variable undefined when we migrated to React Router from Reach Router.
- Removes the dependency on `@lokibai/react-use-copy-clipboard` previously used for copying text to the clipboard. This is due to deprecated usage of the [document.execCommand](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand) function and lack of recent commits (last commit over 3 years ago). Instead, the `navigator.clipboard.writeText` function has been implemented to achieve the desired copy-to-clipboard functionality.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -10,7 +10,6 @@
     "@formatjs/macro": "^0.2.8",
     "@hotosm/id": "^2.21.1",
     "@hotosm/iso-countries-languages": "^1.1.2",
-    "@lokibai/react-use-copy-clipboard": "^1.0.4",
     "@mapbox/mapbox-gl-draw": "^1.3.0",
     "@mapbox/mapbox-gl-geocoder": "^5.0.1",
     "@mapbox/mapbox-gl-language": "^0.10.1",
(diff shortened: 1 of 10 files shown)


HelNershingThapa:
@ramyaragupathy, I could use some help with testing the functionality of copying the intended URL to the clipboard. We don't currently have code coverage for this feature, and we've been using [@lokibai/react-use-copy-clipboard](https://github.com/uniquexiaobai/react-use-copy-clipboard) to achieve the copy-to-clipboard functionality. However, I've encountered issues simulating the `navigator.clipboard` object to ensure that the expected text has been copied to the clipboard but haven't been able to develop a workaround. Can I go ahead and merge this PR for the time being and deal with the test cases later?

HelNershingThapa:
> @HelNershingThapa - do we know what altered the expected behaviour?

Yes, we do. When we switched from Reach Router to React Router, the value of the object we use to build these links changed, which is why.

d-rita (approved):
The functionality works and the tests are good. LGTM

PR #5724, opened 2023-04-21
Hide private teams on organisation detail view for users without manage permission

### LINKED ISSUE #5723: [BUG] Private team visibility update
Ensure private teams are visible only for authorized users in organisation detail page.

Aadesh-Baral:
closes #5723 

_**What to test?**_

Private teams should be hidden on the organization detail page if any of the following conditions are met:
- User is not logged in.
-  User does not have the permission to manage the organization.

Private teams should be displayed on the organization detail page if any of the following conditions are met:
- User has organisation manage permission
- User has system admin role

Public teams should be visible under all circumstances.

### DIFF
--- a/backend/api/organisations/resources.py
+++ b/backend/api/organisations/resources.py
@@ -17,6 +17,7 @@ from backend.services.users.authentication_service import token_auth
 
 
 class OrganisationsBySlugRestAPI(Resource):
+    @token_auth.login_required(optional=True)
     def get(self, slug):
         """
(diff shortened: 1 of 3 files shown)


ramyaragupathy (commented):
@d-rita and I reviewed the functionality. It works as expected. @d-rita  will test rest of the workflows and the code quality.

d-rita (approved):
This works as expected and the tests cover different cases. LGTM

Issue #4858, opened 2021-08-26
Develop Process For Handling Inappropriate Messages


russdeffner:
The Tasking Manager is missing a way for users to report messages as inappropriate and managers/admins to handle removal of messages deemed inappropriate.

My suggested solution is 3-part:
1. Add a 'button' to all messages to flag inappropriate content; akin to 'report this post' or similar actions in social media sites
2. Define who/what group of users receive that message
3. Create a way in the front-end for that/those user(s) defined in 2 to take action to remove inappropriate content

Important - this will need to work across all messages as even though there are no 'private' messages in TM4, a comment on a Task or message to a Team can potentially be 'hidden' from general users and sometimes creates a situation where the person getting mean or otherwise inappropriate messages may not know any other way/may not have any other contact to express the concern.

adiatmad:
I also need the feature to edit or delete comments that reveal personal information, such as full names or other privacy things. It will keep the atmosphere of the conversation better under control.

Aadesh-Baral:
I've divided this issue into two tasks:

- [ ] Add a feature to flag comments. More discussion on how to implement this feature is required.
- [ ] Add a feature to delete comments in the project chat section. Project managers can delete any comment, whilst other users can only delete their own. This has been addressed in #5742.

PR #5742, opened 2023-05-01
Add feature to delete project chat messages.

Aadesh-Baral:
Partially addresses #4858 by adding feature to delete message on project's question and comment section.

Only the project manager and the message's author will be able to delete a message. On frontend delete button will be shown to all messages if user has project manage permission else the delete button will only be shown on the comment which is posted by the logged in user.

image

**TO DO: Add a toast message to indicate the success or failure of comment deleting.**


HelNershingThapa (approved):

d-rita (approved):
Tested the functionality and it works fine

PR #5749, opened 2023-05-02
Display TM stats on the landing page simultaneously

### LINKED ISSUE #5738: [BUG] Inconsistent loading of stats on the landing page
**Describe the bug**
The stats section shows 5 different stats, but they don't load consistently or simultaneously. Three stats appear together, while the other two appear after some time, or vice versa.

**To Reproduce**
1. Go to [the landing page](https://tasks-stage.hotosm.org/). 
2. Scroll to the stats section.
3. Observe that all of the 5 stats don't load at the same time.

**Expected behavior**
All stats should be displayed simultaneously

**Screenshots**
Peek 2023-04-27 10-28

**Additional context**
The stats load inconsistently or simultaneously because they accumulate two different API endpoints.

HelNershingThapa:
Closes #5738 

stats

- Updates the component to load and display stats from two different APIs simultaneously. It does this by using the Promise.all() method, which allows concurrently fetching data from multiple sources. 
- Updated to show a dash ("-") instead of undefined values while the stats are still loading.

### DIFF
--- a/frontend/src/components/homepage/stats.js
+++ b/frontend/src/components/homepage/stats.js
@@ -5,7 +5,7 @@ import shortNumber from 'short-number';
 
 import messages from './messages';
 import { HOMEPAGE_STATS_API_URL } from '../../config';
-import { useFetchWithAbort } from '../../hooks/UseFetch';
+import { fetchLocalJSONAPIWithAbort } from '../../network/genericJSONRequest';
 
 export const StatsNumber = (props) => {
   const value = shortNumber(props.value);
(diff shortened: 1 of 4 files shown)


Aadesh-Baral (approved):
LGTM

HelNershingThapa:
> Works fine. The only issue I see in the future is if one of the endpoints is down, then no stats will be displayed.

Thanks for reviewing the pull request @d-rita and bringing up the potential issue with the endpoints being down. I've tracked this concern on a separate issue [here](https://github.com/hotosm/tasking-manager/issues/5763), so I'll go ahead and merge the pull request now.

PR #5755, opened 2023-05-03
Use FallbackComponent when route throws an exception while rendering

### LINKED ISSUE #5754: Display our custom FallbackComponent when exceptions are thrown
This would be a weird-looking error for the user.

235650082-499de215-1b5b-46fe-bf28-31f0cf0dc390

Instead, display our custom fallback component that offers the Contact Us and Return buttons.

HelNershingThapa:
Resolves #5754


frontend/src/App.js
```diff
@@ -33,11 +33,10 @@ const App = () => {
...
         <div className="w-100 base-font bg-white" lang={locale}>
           <main className="cf w-100 base-font">
-            <Suspense
-              fallback={<ReactPlaceholder showLoadingAnimation={true} rows={30} delay={300} />}
-            >
-              <RouterProvider router={router} />
+            <Suspense fallback={<ReactPlaceholder showLoadingAnimation rows={30} delay={300} />}>
+              <RouterProvider router={router} fallbackElement={<Preloader />} />
             </Suspense>
+            <RouterProvider router={router} fallbackElement={<Preloader />} />
```
Aadesh-Baral, line 39: It looks like there are two instances of the <RouterProvider> component being rendered vertically, which is causing the app to appear twice on the page. This must have been an oversight, and we should remove one of the instances to fix the issue.

Aadesh-Baral (changes requested):

Aadesh-Baral (approved):

frontend/src/App.js
HelNershingThapa: imagine shipping that haha. thanks for the catch @Aadesh-Baral. appreciate it

PR #5756, opened 2023-05-03
Display loading text when organisations and teams are loading under Permissions under project creation

### LINKED ISSUE #5741: [BUG] Teams dropdown showing "No options" instead of loading message while teams list loads
**Describe the bug**
The "No options" message is shown instead of a loading message while the teams list is loading in the Permissions section of the Edit Project page. This bug confuses users who may think the teams' list is empty while loading.

**To Reproduce**
1. Go to the Edit Project page of a project.
2. Navigate to the Permissions section.
3. Click on the Select a team dropdown.
4. Observe the "No options" message instead of a loading message while the team list is still loading.

**Expected behavior**
A loading message should be displayed in the dropdown while the teams' list is loading, indicating to users that the teams are still being loaded in the background and that they should wait for the process to complete.

**Additional context**
This issue occurs due to the absence of a loading message in the dropdown, confusing users who may think the teams list is empty. Nicole brought this issue up in our Slack channel. She reported having trouble adding teams to a project and that after clicking around multiple times, she could eventually select the teams. A screenshot was provided, which showed the message "No options" being displayed instead of a loading message while the teams list was still loading.
Screenshot 2023-04-29 at 18 05 44

HelNershingThapa:
Resolves #5741 

This PR does the following: 
- Adds a new message under the components directory to be utilized throughout the application. This message will be displayed in the select container while organisations and teams options are loading. It was really weird how the loading message won't be displayed if an empty array won't be passed as options.
- Replace fetchLocalJSONAPI with useFetchWithAbort to fetch organizations and teams for permissions.
- Addresses a warning by ensuring that the team ID is used as the key when mapping the teams' list.

Screenshot
Peek 2023-05-03 14-54

### DIFF
--- a/frontend/src/components/messages.js
+++ b/frontend/src/components/messages.js
@@ -104,4 +104,8 @@ export default defineMessages({
     defaultMessage:
       'Your browser does not support WebGL, which is required to render map components. Please try using a different browser or check that <a>WebGL is enabled</a>.',
   },
+  loading: {
+    id: 'common.loading',
+    defaultMessage: 'Loading...',
+  },
 });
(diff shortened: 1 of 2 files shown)


d-rita (approved):
Works as expected

PR #5760, opened 2023-05-04
 Refactor country search for more precise matching

### LINKED ISSUE #5751: [BUG] Country filter should filter only projects associated to the country
**Describe the bug**
While using TM filter by location, South Sudan projects come up for Sudan.

**To Reproduce**
1. Go to `Explore Projects` on tasks.hotosm.org
2. Click on `More filters`
3. Set the location filter to `Sudan`
4. You will see projects related to `South Sudan` listed in the grid view

**Expected behavior**
Show me only `Sudan` projects, instead of showing `South Sudan` projects

**Screenshots**
image

Aadesh-Baral:
Closes #5751 
This PR refactors the country search functionality in response to an issue where South Sudan projects were showing up under Sudan when filtering by location. The previous implementation used an ilike search to match country names in a nested array, which resulted in false positives.

The new implementation uses exact matching to make the search more precise and reliable. It replaces the ilike function with the equality operator, ensuring that only the projects with the exact country name are returned.

This change improves the accuracy of the search results and eliminates the issue like South Sudan projects appearing under Sudan.

### DIFF
--- a/backend/services/project_search_service.py
+++ b/backend/services/project_search_service.py
@@ -331,9 +331,9 @@ class ProjectSearchService:
             sq = Project.query.with_entities(
                 Project.id, func.unnest(Project.country).label("country")
             ).subquery()
-            query = query.filter(sq.c.country == search_dto.country).filter(
-                Project.id == sq.c.id
-            )
+            query = query.filter(
+                func.lower(sq.c.country) == search_dto.country.lower()
+            ).filter(Project.id == sq.c.id)
 
         if search_dto.last_updated_gte:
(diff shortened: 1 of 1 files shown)


d-rita (changes requested):
This is case-sensitive and hence it works well on the frontend since the information is already in the expected case, i.e. South Sudan or Sudan. 
However, when using the API directly/API docs, the search will not return any results. 

Previously:
south sudan case insensitive

Currently:
south_sudan case sensitive

d-rita (approved):
Works as expected. LGTM

PR #5786, opened 2023-05-15
Add basic backend commands on PDM scripts.

Aadesh-Baral:
With addition of  PDM we no longer use virtual environments and developers  need to enable pep 582 on their machine to read packages from` __pypackages__` directory. This can be troublesome sometimes. With pdm scripts we can just use `pdm run <command>` to perform the desired task. This commit adds following pdm scripts:
- `start`: Starts API server same as `python manage.py runserver -d -r`
- `migrate`: Performs database migration. Same as `python manage.py db migrate`
- `upgrade`: Performs database upgrade same as `python manage.py db upgrade`
- `downgrade`: Performs database downgrade same as `python manage.py db downgrade`
- `test`: Runs tests same as` python -m unittest discover`
- `lint`: Lints backend code same as `black manage.py backend tests migrations`
- `flake8`: Checks backend code standards same as `flake8 manage.py backend tests migrations`
- `coverage`: Checks tests coverage same as `coverage run -m unittest discover`

For example with this addition we can simply start backend server using **`pdm run start`**.


eternaltyro (approved):
LGTM

PR #5808, opened 2023-05-19
Fix app configuration in create_app for testing.

### LINKED ISSUE #5691: [BUG] Several errors in Backend CI even though the pipeline seems to pass
There are several errors in the backend tests - particularly when initiating the database - even though the pipeline seems to pass. For example, there are multiple instances of the error shown below:

```
...
  File "/home/circleci/.pyenv/versions/3.7.14/lib/python3.7/site-packages/sqlalchemy/engine/default.py", line 493, in connect
    return self.dbapi.connect(*cargs, **cparams)
  File "/home/circleci/.pyenv/versions/3.7.14/lib/python3.7/site-packages/psycopg2/__init__.py", line 127, in connect
    conn = _connect(dsn, connection_factory=connection_factory, **kwasync)
sqlalchemy.exc.OperationalError: (psycopg2.OperationalError) FATAL:  database "tm" does not exist

(Background on this error at: http://sqlalche.me/e/13/e3q8)
```

Aadesh-Baral:
When running tests for the function that involves creating a thread and an associated app context, we encountered an issue with the app configuration. The create_app function, responsible for creating the app context, was using the default `EnvironmentConfig` class even during testing.

To address this issue, we made a modification in the create_app function. We added a condition that checks if the TM_ENVIRONMENT environment variable is set to "test". If so, the configuration is set to `TestEnvironmentConfig` instead of the default `EnvironmentConfig`. 

By introducing this conditional logic, it ensures that the create_app function correctly sets the `TestEnvironmentConfig` for the app created across the project when the test environment is detected.

closes #5691

### DIFF
--- a/backend/__init__.py
+++ b/backend/__init__.py
@@ -73,6 +73,9 @@ def create_app(env="backend.config.EnvironmentConfig"):
     app = Flask(__name__, template_folder="services/messaging/templates/")
 
     # Load configuration options from environment
+    # Set env to TestEnvironmentConfig if TM_ENVIRONMENT is test
+    if os.getenv("TM_ENVIRONMENT") == "test":
+        env = "backend.config.TestEnvironmentConfig"
     app.config.from_object(env)
     # Enable logging to files
     initialise_logger(app)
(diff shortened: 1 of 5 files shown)


eternaltyro:
@Aadesh-Baral can you move the integration test out of this PR and into its own PR?

Aadesh-Baral:
> @Aadesh-Baral can you move the integration test out of this PR and into its own PR?

Done @eternaltyro

eternaltyro (approved):
LGTM

PR #5837, opened 2023-05-24
Use idempotent remote control commands for JOSM

tsmock:
Right now, there is a race condition when using JOSM remote control.

The TM currently makes the following calls, in order:
1. Import task boundary
 a. Create a new, empty layer with a hardcoded XML (I believe this is for the `upload="never"` option)
 b. Load the boundary into the new layer
2. Load data from OSM
 a. Create a new, empty layer with a hardcoded XML
 b. Load the data from a specified bounding box into JOSM

With the current implementation, the return information from JOSM is not checked. This means that if 2a fails, the OSM data will be loaded into the layer created in 1a. Additionally, if the user changes layers prior to the data load, the data may load into the wrong layer.

Having two sub-steps for either step is unnecessary. For the first, there is an `upload_policy=never` query parameter that can be used, instead of creating a new layer via an XML option. For the second step, there is a `layer_name` query parameter that can be used.

With this patch, I do the following:
1. Import task boundary with the `upload` and `download` policies set to `never`, and with the layer `locked` (this prevents users from moving the boundaries).
2. Load data from OSM (I just collapsed multiple remote control calls to a single call)

### DIFF
--- a/frontend/src/utils/openEditor.js
+++ b/frontend/src/utils/openEditor.js
@@ -135,19 +135,16 @@ function loadTasksBoundaries(project, selectedTasks) {
   const layerName = `Boundary for task${selectedTasks.length > 1 ? 's:' : ':'} ${selectedTasks.join(
     ',',
   )} of TM Project #${project.projectId} - Do not edit or upload`;
-  const emptyTaskLayerParams = {
+  const tmTaskLayerParams = {
     new_layer: true,
     layer_name: layerName,
-    data: '<?xml version="1.0" encoding="utf8"?><osm generator="JOSM" upload="never" version="0.6"></osm>',
-  };
-  const tmTaskLayerParams = {
-    new_layer: false,
+    layer_locked: true,
+    download_policy: "never",
+    upload_policy: "never",
     url: getTaskXmlUrl(project.projectId, selectedTasks).href,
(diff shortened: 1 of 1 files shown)


eternaltyro (approved):
LGTM

anthaas:
Hi, after this change, the boundary can't be added in TODO list plugin in JOSM.  Allowing the interaction with the TODO plugin is very useful tool in the multiple tasks validation workflow.

How to reproduce this issue: In JOSM set the boundary layer as active. Use search tool with the query 'type:way'. This selection is not allowed to add to todo list anymore.

tsmock:
@anthaas : The TODO plugin has been modified to allow TODO items to be added from more layer types. That should solve the issue that people are having. _If they still have the issue after updating plugins, they probably need to update to Java 17_.

If they used the installer for Microsoft Windows, they probably need to uninstall JOSM and reinstall JOSM -- there was a change in the installer awhile back, and they might be having issues with multiple JOSM installations.

If they use the `josm.jnlp` file, they need to install [OpenWebStart](https://openwebstart.com).

If they use the `josm.jar` file, they should (a) Uninstall Java 8 and (b) download [Azul Java](https://www.azul.com/downloads).

PR #5838, opened 2023-05-24
Add better defaults to backend config.py file

eternaltyro:
Add better defaults to make it easy for HOT to deploy Tasking Manager using containers thereby reducing the need to manipulate every environment variable.

Cast flags as bool instead of int wherever appropriate


Aadesh-Baral (approved):
LGTM

Issue #5864, opened 2023-06-01
Toggle Dropdown Functionality [BUG]


mtalhabaig3:
**Describe the bug**
When we repeatedly press the menu buttons, the dropdown menu keeps opening instead of toggling.

**To Reproduce**
1. Go to "Explore Projects" page from the home page.
2. Click multipe times on the menu buttons.
3. The drop down menu keeps opening

**Expected behavior**
The dropdown menu should toggle upon clicking the menu buttons

**Screenshots**

https://github.com/hotosm/tasking-manager/assets/57634631/154a4f86-c1b3-40e3-ba4a-eb1510f13322

**Desktop (please complete the following information):**
 - OS:  MAC
 - Browser chrome

PR #5865, opened 2023-06-01
solved the dropdown toggle bug

mtalhabaig3:
Fixes issue #5864 
I have fixed the issue by referencing the correct component for using the "useOnClickOutside" hook, previously DropDownContent button was being referenced which was causing the bug and the click on the menu buttons was also being listened as an outside click. I referenced the menu button itself so now only the outside clicks are listened.

https://github.com/hotosm/tasking-manager/assets/57634631/6244219b-7099-4a80-9d44-4c8b99d6db16

@mojodna please review

### DIFF
--- a/frontend/src/components/dropdown.js
+++ b/frontend/src/components/dropdown.js
@@ -1,11 +1,11 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import { useNavigate } from 'react-router-dom';
 
 import { useOnClickOutside } from '../hooks/UseOnClickOutside';
 import { ChevronDownIcon, CheckIcon } from './svgIcons';
 import { CustomButton } from './button';
 
-const DropdownContent = React.forwardRef((props, ref) => {
+const DropdownContent = (props, ref) => {
   const navigate = useNavigate();
   const isActive = (obj) => {
     return props.value === obj.value;
@@ -51,7 +51,6 @@ const DropdownContent = React.forwardRef((props, ref) => {
 
   return (
     <div
-      ref={ref}
       className={`db tl mt1 ba b--grey-light br1 absolute shadow-1 z-5 flex flex-column${
         props.toTop ? ' bottom-3' : ''
       }${props.options.length > 9 ? ' h5 overflow-y-scroll' : ''}`}
@@ -116,13 +115,15 @@ const DropdownContent = React.forwardRef((props, ref) => {
       ))}
     </div>
   );
-});
+};
 
 export function Dropdown(props) {
   const [display, setDisplay] = useState(false);
+
   const contentRef = React.createRef();
 
   const toggleDropdown = () => {
+    console.log('toggle');
     setDisplay(!display);
   };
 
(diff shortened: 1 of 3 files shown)


HelNershingThapa (changes requested):
Thank you for your contribution and for addressing the bug described in issue #5864. I can see that the dropdown content is now hidden when clicking the button, which is great progress!

However, upon reviewing the changes, I noticed that the dropdown selection functionality seems to have been unintentionally affected. When the select options are displayed, clicking anywhere triggers the useOnClickOutside function, causing the dropdown to close and **_preventing the selection of options_**.

Please take a look at the code and see if there's a way to modify the useOnClickOutside implementation to exclude the select options. Once this issue is addressed, we can merge your pull request with confidence. Keep up the good work!

mtalhabaig3:
@HelNershingThapa Thanks for the review. I am on the task 🙂

mtalhabaig3:
@HelNershingThapa I have solved the issue. Basically i did what you said, I excluded the click on the drop down menu from being an outside click in the useOnClickOutside hook by using the event.target.closest function. Now the functionality works as required. Kindly review 😊

HelNershingThapa (changes requested):
Thanks for fixing the issue and making the requested changes. I've gone through the code again, and I have a couple of points to discuss.

First, there are failing test cases in the dropdown file because of the new .dropdown-content className. Let's make sure all the tests pass before merging the code.

Secondly, I understand why you modified the useOnClickOutside hook, but I'm not a big fan of changing it. The hook is meant to be reusable, and modifying it for this specific case might limit its usefulness elsewhere. Here's my suggestion: let's get rid of the useOnClickOutside hook in the Dropdown component. Instead, we can handle the click events ourselves and check if the clicked element is outside the CustomButton and DropdownContent refs. If it is, we can close the dropdown content.

What do you think about this approach? It keeps things simple and maintains the hook's generic nature.

```js
  const handleClickOutside = (event) => {
    if (
      contentRef.current &&
      !contentRef.current.contains(event.target) &&
      buttonRef.current &&
      !buttonRef.current.contains(event.target)
    ) {
      setDisplay(false);
    }
  };

  useEffect(() => {
    document.addEventListener('click', handleClickOutside);
    return () => {
      document.removeEventListener('click', handleClickOutside);
    };
  }, []);
```

mtalhabaig3:
@HelNershingThapa yes, i think it is a better solution, this way we can make the hook generic and reusable! I will try to solve it with this approach.

mtalhabaig3:
@HelNershingThapa I have changed the implementation according to what you suggested, but there were some changes that were to be done in the code you provided. I had to add the two refs as the dependencies of the use effect and also had to change the logical operations from '&&' to '||'. Now the UI is working as expected. Kindly review my changes 🙂.

frontend/src/components/dropdown.js
```diff
@@ -1,4 +1,4 @@
-import React, { useState } from 'react';
+import React, { useEffect, useState } from 'react';
 import { useNavigate } from 'react-router-dom';
 
 import { useOnClickOutside } from '../hooks/UseOnClickOutside';
```
HelNershingThapa, line 4: Please consider removing the useOnClickOutside declaration as it is currently unused. This is causing the CI/CD build to fail.

```suggestion

```

frontend/src/components/dropdown.js
```diff
@@ -135,18 +157,21 @@ export function Dropdown(props) {
...
+        onClick={toggleDropdown}
+        className={`blue-dark ${props.className || ''}`}
+      >
         <div className="lh-title dib ma0 f6">{getActiveOrDisplay()}</div>
+
         <ChevronDownIcon style={{ width: '11px', height: '11px' }} className="pl3 v-mid pr1" />
       </CustomButton>
       {display && (
         <DropdownContent
+          ref={buttonRef}
```
HelNershingThapa, line 173: This is a small observation - it seems that `contentRef `and `buttonRef` have been defined the other way around. It might be more appropriate to pass `buttonRef` to the `CustomButton` component and `contentRef` to the dropdown component.

HelNershingThapa (commented):
Tested the `handleClickOutside` function locally, and it appears to be functioning correctly from my end. Good job on implementing it! Let's proceed with merging once the other changes are addressed.

frontend/src/components/dropdown.js
mtalhabaig3: Sure!

frontend/src/components/dropdown.js
mtalhabaig3: Yes, let me make the changes

mtalhabaig3:
@HelNershingThapa done!

HelNershingThapa (approved):

HelNershingThapa:
@mtalhabaig3 Thank you for this! Your changes has been merged into develop.

mtalhabaig3:
Your welcome!😊

PR #5881, opened 2023-06-06
Add confirmation pop up window when leaving a team

### LINKED ISSUE #5657: Add confirmation pop up window when leaving a team
Today I left a team by mistake. The button is in the same bottom right corner and the same color as is the contribution button so while I had opened the team tab and multitasking I automatically clicked in the corner on big red button. 

Since this action can't be taken back by the user I think the confirmation window would be appropriate.

image
image

A pop up window that will ask user to confirm their decision to leave the team.

Put the button in different spot.

**Additional context**
Yes I understand that this sounds silly but after spending over 100 hours on tasking manager my brain sometimes works on autopilot. Plus I found out that I am not the only one who did this before by mistake.

HelNershingThapa:
Closes #5657 

- Triggers a modal popup upon clicking the Leave Team button
- Adds a new component LeaveTeamConfirmationAlert that has the modal content. Styles were forged from the project edit page's previously implemented dialogs.
- Adds test cases for the new component.

### DIFF
--- a/frontend/src/components/teamsAndOrgs/leaveTeamConfirmationAlert.js
+++ b/frontend/src/components/teamsAndOrgs/leaveTeamConfirmationAlert.js
@@ -1,0 +1,33 @@
+import { FormattedMessage } from 'react-intl';
+
+import commonMessages from '../../views/messages';
+import messages from './messages';
+import { Button } from '../button';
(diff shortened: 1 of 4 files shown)


Aadesh-Baral (approved):
LGTM

PR #5889, opened 2023-06-07
Fix eslint warnings for valid-typeof

tsmock:
We were comparing to undefined (literal) instead of 'undefined' (string). `typeof` will always return a string, so `'undefined' !== undefined` would always return `true`.

For more details, see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof .

### DIFF
--- a/frontend/src/components/editor.js
+++ b/frontend/src/components/editor.js
@@ -11,7 +11,7 @@ export default function Editor({ setDisable, comment, presets, imagery, gpxUrl }
   const iDContext = useSelector((state) => state.editor.context);
   const locale = useSelector((state) => state.preferences.locale);
   const [customImageryIsSet, setCustomImageryIsSet] = useState(false);
-  const windowInit = typeof window !== undefined;
+  const windowInit = typeof window !== 'undefined';
   const customSource =
     iDContext && iDContext.background() && iDContext.background().findSource('custom');
 
--- a/frontend/src/components/teamsAndOrgs/teams.js
+++ b/frontend/src/components/teamsAndOrgs/teams.js
@@ -310,19 +310,13 @@ export function TeamSideBar({ team, members, managers, requestedToJoin }: Object
           return member.username.toLowerCase().includes(searchQuery.toLowerCase());
         })
       : members;
-  const organisations = useSelector((state) => state.auth.organisations);
-  const pmTeams = useSelector((state) => state.auth.pmTeams);
 
   return (
     <ReactPlaceholder
       showLoadingAnimation={true}
       type="media"
       rows={20}
-      ready={
-        typeof team.teamId === 'number' &&
-        typeof organisations !== 'undefined' &&
-        typeof pmTeams !== 'undefined'
-      }
+      ready={typeof team.teamId === 'number'}
     >
       <div className="cf pb2">
         <div className="w-20 pv2 dib fl">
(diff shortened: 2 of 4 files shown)


frontend/src/components/teamsAndOrgs/teams.js
```diff
@@ -318,8 +320,8 @@ export function TeamSideBar({ team, members, managers, requestedToJoin }: Object
       rows={20}
       ready={
         typeof team.teamId === 'number' &&
-        typeof organisations !== undefined &&
-        typeof pmTeams !== undefined
+        typeof organisations !== 'undefined' &&
```
HelNershingThapa, line 323: Not directly related to the pull request, but while reviewing the code, I observed that the `TeamSideBar` component doesn't need the `organisations` and `pmTeams` variables to render. Consequently, we can eliminate the usage of these two variables.

frontend/src/components/editor.js
```diff
@@ -11,7 +11,7 @@ export default function Editor({ setDisable, comment, presets, imagery, gpxUrl }
   const iDContext = useSelector((state) => state.editor.context);
   const locale = useSelector((state) => state.preferences.locale);
   const [customImageryIsSet, setCustomImageryIsSet] = useState(false);
-  const windowInit = typeof window !== undefined;
+  const windowInit = typeof window !== 'undefined';
```
HelNershingThapa, line 14: Isn't the `window` object always available? Is it possible for the `window` object to be `undefined` under any circumstances?

HelNershingThapa (approved):
I have left some comments that may appear unrelated to the pull request. My intention was to seek clarity and confirm my understanding of certain aspects.

frontend/src/components/teamsAndOrgs/teams.js
HelNershingThapa: Made the change https://github.com/hotosm/tasking-manager/commit/5ff810cd7d79ce7bee29680a61330d0001cc4ca6.

frontend/src/components/editor.js
tsmock: I think it is always available in a browser. I don't think it is in NodeJS.

frontend/src/components/teamsAndOrgs/teams.js
tsmock: Works for me, especially since I had to define the variables in the first place.

eternaltyro: @HelNershingThapa This change should ideally be in a separate PR.

tsmock: To clarify, the change of `typeof organisations !== undefined` -> `typeof organisations !== 'undefined'` should be in a separate PR or should the removal of `typeof organisations !== 'undefined'` be in a separate PR?

HelNershingThapa: @eternaltyro recommended that the changes made in https://github.com/hotosm/tasking-manager/commit/5ff810cd7d79ce7bee29680a61330d0001cc4ca6 should have been submitted as a separate PR. I agree that it would have been preferable to have it as a separate PR.  Nevertheless, given that it has already been synced and taking into account the situation, it has been decided to overlook it this time.

tsmock: I could unsync it. It isn't that difficult to do. :)

PR #5890, opened 2023-06-07
ReactJS: Fix an input component changing from uncontrolled to controlled

tsmock:
This is due to the "initial" value being `undefined` or `null` prior to the user data being fetched.


HelNershingThapa (approved):
LGTM. Thank you for your efforts in addressing those warnings and ensuring the inputs are _controlled_

PR #5721, opened 2023-04-19
Frontend dependency updates

tsmock:
This PR intends to make the TM compatible with the defaults of Debian Bookworm. This means that the frontend needs to be compatible with NodeJS 18.x.

In general, the code changes are intended to be limited to compatibility changes. In the event that there needs to be a reformat due to formatting tool updates, those will be in separate commits (to make rebasing onto a new develop HEAD easier).

This PR replaces frontend PRs #5712, #5640, #5586, #5585, #5583, #5540, #5529, #5510, #5474, #5467, and #5177.

This does not update Rapid to v2 -- that will need to be a separate PR.

Total bundle size has decreased from 14.26 MB to 8.07 MB.
main bundle size has decreased from 3.12 MB to 2.56 MB.
 * This can be additionally reduced by lazy-loading the map component

At one point, I had the main bundle size reduced to 700 KB (before I updated the react-scripts to 5.0.1, which required a workaround for https://github.com/facebook/create-react-app/issues/11770 ). Either that update or the workaround caused two bundles to merge; the first with our code, the second with most of the "heavy" modules.

<details><summary>Test runs from #5721</summary>

File | Test | RPi (Node 16, develop) | Mac (Node 16, updates) | Mac (Node 18, updates)
-- | -- | -- | -- | --
src/components/projectCard/tests/dueDateBox.test.js | test DueDate › with tooltip message |   |   | 1
src/components/projectDetail/tests/questionsAndComments.test.js | test if QuestionsAndComments component › renders tabs for writing and previewing comments |   | 1 | 1
src/components/projectDetail/tests/shareButton.test.js | test if shareButton › render open corresponding window popup |   |   | 1
src/components/projectDetail/tests/shareButton.test.js | test if shareButton › render shareButton for project with id 1 |   |   | 1
src/components/taskSelection/tests/actionSidebars.test.js | Appearance of unsaved map changes to be dealt with while mapping › when splitting a task |   | 2 | 1
src/components/taskSelection/tests/actionSidebars.test.js | Appearance of unsaved map changes to be dealt with while mapping › when submitting a task |   |   | 1
src/components/taskSelection/tests/actionSidebars.test.js | Completion Tab for Validation › should update status and comments for multiple tasks |   | 1 |  
src/components/taskSelection/tests/contributions.test.js | Contributions › clean user selection if we click on the selected tasks of the user | 1 |   |  
src/components/taskSelection/tests/contributions.test.js | Contributions › render users, links, and | 1 | 1 | 1
[...]

PR #5892, opened 2023-06-07
Update query-string to v8.1.0 from v7.1.3

tsmock:
The major change is that it is now pure ESM and no longer exports specific functions. It also requires Node.js 14+.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -40,7 +40,7 @@
     "mapbox-gl-draw-rectangle-mode": "^1.0.4",
     "marked": "^4.3.0",
     "osmtogeojson": "^3.0.0-beta.5",
-    "query-string": "^7.1.3",
+    "query-string": "^8.1.0",
     "react": "^17.0.2",
     "react-accessible-accordion": "^4.0.0",
     "react-calendar-heatmap": "^1.9.0",
--- a/frontend/src/components/projects/tests/moreFiltersForm.test.js
+++ b/frontend/src/components/projects/tests/moreFiltersForm.test.js
@@ -1,5 +1,5 @@
 import '@testing-library/jest-dom';
-import { parse } from 'query-string';
+import queryString from 'query-string';
 import { act, render, screen, waitFor } from '@testing-library/react';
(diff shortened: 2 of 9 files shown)


HelNershingThapa:
This babel-loader problem occurred to me. Does it compile successfully at your end?

```js
Failed to compile.

./node_modules/query-string/base.js 422:14
Module parse failed: Unexpected token (422:14)
File was processed with these loaders:
 * ./node_modules/babel-loader/lib/index.js
You may need an additional loader to handle the result of these loaders.
| 
|   return {
>     url: url_?.split('?')?.[0] ?? '',
|     query: parse(extract(url), options),
|     ...(options && options.parseFragmentIdentifier && hash ? {
```

tsmock:
It did. But I hadn't wiped the `node_modules` directory, so maybe there was a required component there. I'll debug and (probably) push a new yarn.lock file.

tsmock:
Well, `rm -rf node_modules && yarn install && yarn build` worked for me.

Maybe there is a version issue? My tool versions are:
* nodejs: 16.20.0
* yarn: 1.22.19

eternaltyro:
LGTM. @HelNershingThapa can you verify that the `rm node_modules` workaround fixes the issue for us?

eternaltyro:
@HelNershingThapa what exactly is the issue? Please include logs when you have issues .

tsmock:
Updating to CRA (`react-scripts`) to 5.0.1 fixes the issue with `yarn start`. I'll open a separate PR for that, since I'm going to need to fiddle with some code splitting (`main.*.js` is 6.07 MB, 2.35 MB from polyfill.js).

ramyaragupathy:
@tsmock @varun2948 would this be next in line after #5721? cc @royallsilwallz @emi420 @manjitapandey

tsmock:
No; #5721 also updates query-string and includes all of the changes from this PR. See [package.json#L43](https://github.com/hotosm/tasking-manager/pull/5721/files#diff-da6498268e99511d9ba0df3c13e439d10556a812881c9d03955b2ef7c6c1c655R43).

tsmock:
This PR is no longer necessary; the same changes were part of #5721.

PR #5899, opened 2023-06-09
Define custom exceptions for better handling error response

### LINKED ISSUE #5197: Improve error handling on backend and code readability
The current approach to handling backend exceptions has resulted in code readability issues and challenges in maintaining the codebase.

**1. Fix messed-up code readability #5911, #5909** 
The way exceptions are currently handled in the code has made it harder to read and comprehend. Combining SubCodes and error messages when raising exceptions can confuse new contributors trying to understand the error-handling logic.
For example, we have an endpoint that updates the project metadata. This is only permitted to users with Project Management permission. To return an error if the request is made by the user without Project Management permission we raise an exception like `ValueError` with the message “This action is only allowed to Project Managers”. Since we also have a `SubCode` parameter in the error response we are passing a subcode in the message itself like:
```python 
   raise ValueError("PM_ROLE_REQUIRD-  This action is only allowed to Project Managers")
```
On the view function, then we split the error string by `-` to get the subcode and error message like:
```python
try:
    update_project(request.json)
except ValueError as e:
    return {"Error": str(e).split("-")[1], "SubCode": str(e).split("-")[0]}, 403 
```
**Proposed Solution:**
Create a custom exceptions on which we can easily pass the extra information like sub-code, messages and any other args and the exception class should be able to prepare response based on the information passed.

**2. Excessive try-except blocks: #5912**
The code contains multiple try-except blocks to handle various types of exceptions. This can lead to code redundancy and reduced clarity, as each exception is managed separately.

For instance, if the view function expects NotFound, ValueError, and DataError exceptions, it handles them individually:
```python
try:
    my_function()
except NotFound:
    return {"SubCode": "RESOURCE_NOT_FOUND", "Error": "The requested resource was not found in the server"}, 404
except ValueError as e:
    return {"Error": str(e).split("-")[1], "SubCode": str(e).split("-")[0]}, 403
except DataError as e:
    return {"Error": str(e), "SubCode": "INVALID_REQUEST"}, 400
except Exception as e:
    return {"Error": str(e), "SubCode": "INTERNAL_SERVER_ERROR"}, 500
```

**Proposed Solution:**
The new custom exceptions should be handled using flask error handling mechanism instead of using try except blocks. All of the unhandled exceptions should be captured by flask generic error handler and return standard error response.

**3. More detailed and standardized error response: #5910, #5909**
Improving the error response to provide more information is essential. It can include detailed SubCodes, clear error messages, and additional context, making it easier for frontend developers and end-users to understand the issues.

For example, instead of a simple "RESOURCE_NOT_FOUND" SubCode, we can use "PROJECT_NOT_FOUND" and include the project ID in the response:
```json
{
    "error": {
        "message": "The requested project was not found.",
        "sub_code": "PROJECT_NOT_FOUND",
        "code": 404,
        "details": {
            "project_id": "12345"
        }
    }
}
```

In case of invalid request we can provide in which field the issue occurred like in case of missing required field we can raise following error response:
```json
{
    "error": {
        "message": "This field is required.",
        "sub_code": "FIELD_REQUIRED",
        "code": 400,
        "details": {
            "field_name": "name"
        }
    }
}
```

**4. Separation of error messages from the code: #5909**
Storing error messages in a separate JSON file provides a more organized and flexible approach. This allows easy updates and modifications of error messages without modifying the codebase.
Create an error_messages.json file to store all error messages, each with a unique sub-code, descriptive message, and additional information if required.
```json
{
    "PM_ROLE_REQUIRED": "This action is only allowed for Project Managers.",
    "PROJECT_NOT_FOUND": "The requested project was not found.",
    "INVALID_REQUEST": "Cannot validate the request.",
    "FIELD_REQUIRED": "This field is required."
}

```
- [x] #5909
- [ ] #5910
- [ ] #5911
- [ ] #5912

### LINKED ISSUE #5909: Define custom exceptions for better handling backend errors
Create custom exceptions that can accept various parameters such as code, sub-code, and error message. This will allow for more flexibility in error handling and provide better context to frontend applications.

Aadesh-Baral:
closes #5909 

This pull request (PR) introduces the necessary infrastructure to enhance backend error handling in the project. It includes the implementation of custom exception classes that extend the `BaseException` class from `werkzeug.exceptions.HTTPException`. These custom exceptions offer additional features for handling various types of errors effectively.

Moreover, this PR also introduces a mechanism to separate error messages from the code. The error messages are now stored in a JSON file named "error_messages.json". This approach simplifies the management and updates of error messages independently, providing more flexibility in maintaining and modifying them

Example:
We can raise a NotFound Exception as:
```python
raise NotFound(sub_code="PROJECT_NOT_FOUND", project_id=1)
```
This will return error response as:
```json
{
  "error": {
    "message": "The requested project was not found in the server."
    "code": 404,
    "sub_code": "PROJECT_NOT_FOUND",
    "details": {"project_id": 1}
}
````

For more info please have a look at individual commit messages.

**NOTE: This PR mainly focuses on setting up the infrastructure in order to better error handling; Hence these custom exceptions aren't used to handle errors.**

### DIFF
--- a/backend/__init__.py
+++ b/backend/__init__.py
@@ -17,7 +17,7 @@ import os
 import json
 from logging.handlers import RotatingFileHandler
 
-from flask import Flask, redirect
+from flask import Flask, redirect, request
 from flask_cors import CORS
 from flask_migrate import Migrate
 from requests_oauthlib import OAuth2Session
@@ -107,6 +107,39 @@ def create_app(env="backend.config.EnvironmentConfig"):
     mail.init_app(app)
 
     app.logger.debug("Add root redirect route")
+
+    @app.errorhandler(Exception)
+    def handle_generic_error(error):
+        """Generic error handler for all exceptions"""
+        from backend.exceptions import format_sub_code
+
+        app.logger.exception(error)
+
+        error_message = (
+            str(error)
+            if len(str(error)) > 0
+            else ERROR_MESSAGES["INTERNAL_SERVER_ERROR"]
+        )
+        error_code = error.code if hasattr(error, "code") else 500
+        error_sub_code = (
+            format_sub_code(error.name)
+            if hasattr(error, "name")
+            else "INTERNAL_SERVER_ERROR"
+        )
+        return (
+            {
+                "error": {
+                    "code": error_code,
+                    "sub_code": error_sub_code,
+                    "message": error_message,
+                    "details": {
+                        "url": request.url,
+                        "method": request.method,
+                    },
+                }
+            },
+            error_code,
+        )
 
     @app.route("/")
     def index_redirect():
--- a/backend/__init__.py
+++ b/backend/__init__.py
@@ -36,12 +36,26 @@ def sentry_init():
     """Initialize sentry.io event tracking"""
     import sentry_sdk
     from sentry_sdk.integrations.flask import FlaskIntegration
+    from backend.exceptions import (
+        BadRequest,
+        NotFound,
+        Unauthorized,
+        Forbidden,
+        Conflict,
+    )
 
     sentry_sdk.init(
         dsn=EnvironmentConfig.SENTRY_BACKEND_DSN,
         environment=EnvironmentConfig.ENVIRONMENT,
         integrations=[FlaskIntegration()],
         traces_sample_rate=0.1,
+        ignore_errors=[
+            BadRequest,
+            NotFound,
+            Unauthorized,
+            Forbidden,
+            Conflict,
+        ],  # Ignore these errors as they are handled by the API
     )
 
 
--- a/backend/__init__.py
+++ b/backend/__init__.py
@@ -14,6 +14,7 @@ except ImportError as e:
     logging.info(e)
 
 import os
+import json
 from logging.handlers import RotatingFileHandler
 
 from flask import Flask, redirect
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Review by d-rita — APPROVED** [review]
LGTM

File: backend/__init__.py [hunk 1]
```diff
@@ -27,16 +28,34 @@
 from backend.config import EnvironmentConfig
 
 
+# Storing error messages in a json file so that it is loaded only once at startup (Used in exceptions.py)
+ERROR_MESSAGES = json.load(open("backend/error_messages.json"))
```
eternaltyro, line 32: This is not the best way to handle file open since we usually forget to close. Please use the better approach to automatically close open file.

```
with open("backend/error_messages.json") as jsonfile:
  ERROR_MESSAGES = json.load(jsonfile)
```

backend/__init__.py
```diff
@@ -89,6 +108,39 @@ def create_app(env="backend.config.EnvironmentConfig"):
...
+                    "code": error_code,
+                    "sub_code": error_sub_code,
+                    "message": error_message,
+                    "details": {
+                        "url": request.url,
+                        "method": request.method,
+                    },
+                }
+            },
+            error_code,
```
eternaltyro, line 141: Why the dict and the `error_code` also separately?

eternaltyro (changes requested):
Please fix the file.close() issue. Otherwise, good to go.

backend/__init__.py
eternaltyro: Additionally, in a later PR. Please also consider using module.path style something for the JSON file location.

backend/__init__.py
Aadesh-Baral: Returned dict is response data while error_code is the HTTP status code associated with the error.

eternaltyro (approved):
LGTM. Good to go.

eternaltyro:
Merged.

PR #5908, opened 2023-06-12
Update react scripts

tsmock:
This updates react-scripts from 4.0.3. to 5.0.1. Due to some issues with react-scripts, CRACO was added to work around an issue with svg images. CRACO did cause the main bundle size to increase significantly to 6 MB, and the main bundle size can reduced via lazy loading in the router (see #5914). There is more in-depth analysis in that PR on decreasing the initial size of all bundles _needed_ for the initial render.


HelNershingThapa:
Shouldn't the CRACO configuration file have been included in this PR?

tsmock:
Yep. It looks like I messed up when I was cherry-picking the commit; I didn't run `git add` on the config file.

HelNershingThapa:
Let's also include the fix https://github.com/hotosm/tasking-manager/pull/5721/commits/9ed1d0c6608723b4d8b7f2aa356a8be4eb47c76d for the testimonial image in this PR?

frontend/package.json
```diff
@@ -84,11 +83,11 @@
     "patch-id": "bash -c \"cp patch/imagery.min.json public/static/id/data\"",
     "patch-rapid": "bash -c \"cp patch/rapid-imagery.min.json public/static/rapid/data/imagery.min.json\"",
     "preparation": "bash -c \"if (test -a ../tasking-manager.env); then grep -hs ^ ../tasking-manager.env .env.expand > .env; else cp .env.expand .env; fi\"",
-    "start": "npm run preparation && npm run copy-static && npm run patch-id && npm run patch-rapid && react-scripts start",
-    "build": "npm run preparation && npm run update-static && npm run patch-id && npm run patch-rapid && react-scripts build",
+    "start": "npm run preparation && npm run copy-static && npm run patch-id && npm run patch-rapid && craco start",
+    "build": "npm run preparation && npm run update-static && npm run patch-id && npm run patch-rapid && craco build --stats",
```
HelNershingThapa, line 87: To achieve quicker and leaner production bundles, wouldn't it be desirable to omit the `--stats` flag?

tsmock: > To achieve quicker [...] production bundles, wouldn't it be desirable to omit the --stats flag?

Surprisingly, no.
Here are some test build times (using `time`, and specifically the total time elapsed). Starting with run 22, the runs were with all other applications closed _and_ the computer locked.

<details><summary>Runtime statistics for <code>--stats</code> and no <code>--stats</code></summary>

Run | No stats (s) | Stats (s)
-- | -- | --
1 | 129.38 | 187.45
2 | 110.25 | 76.32
3 | 107.88 | 76.46
4 | 96.8 | 81.37
5 | 97.73 | 79.5
6 | 131.19 | 79.67
7 | 111.21 | 79.06
8 | 95.57 | 80.03
9 | 92.99 | 78.9
10 | 89.16 | 80.49
11 | 93.77 | 81.54
12 | 96.78 | 80.46
13 | 112.91 | 80.05
14 | 95.33 | 80.15
15 | 86.28 | 81.81
16 | 91.38 | 81.74
17 | 88.22 | 81.17
18 | 96.19 | 80.85
19 | 96.58 | 78.06
20 | 93.15 | 80.99
21 | 91.93 | 79.3
22 | 84.5 | 75.25
23 | 87.12 | 79.56
24 | 87.97 | 78.61
25 | 84.9 | 79.1
26 | 84.71 | 79.02
27 | 85.36 | 77.64
28 | 82.91 | 57.84

</details>

> To achieve [...] leaner production bundles, wouldn't it be desirable to omit the --stats flag?

`build` is 339M for `--stats` with 107M from the stats file, and 227M without the `--stats`. With that said, the numbers really need to be compared over multiple runs. And isn't what would reach the users _anyway_. For reference, it is 228M with current `develop`.

HelNershingThapa: But isn't it evident from the table that excluding the "--stats" flag generally led to shorter build times?

Could you kindly clarify why you mentioned that excluding the "--stats" flag might not be desirable? I value your insights and would appreciate further clarification.

tsmock: > But isn't it evident from the table that excluding the "--stats" flag generally led to shorter build times?

That wasn't the case. I probably should have flipped the description (`--stats` and no `--stats` -> no `--stats` and `--stats`) to match the table headers.

The average for no `--stats` was 96.5s and for `--stats` it was 82.5s. The median for no `--stats` was 93.46s and the median for `--stats` was 79.6s.

It _does_ produce a slightly larger `build` directory, most of which is from the `bundle-stats.json` file (which doesn't need to be served). I should probably do another 10 runs of `--stats` and no `--stats` to see how the `build` directory size changes and whether or not it is _consistent_ between runs.

HelNershingThapa: I was primarily only trying to understand whether including the `--stats` flag would make a significant difference. Considering the insights you've provided and the analysis conducted, it seems reasonable to proceed with merging the pull request as it is. Does anything need to be addressed to prevent the stats file from being served?

Thank you for your valuable contributions to this conversation. Are there any additional considerations before merging?

tsmock: > Does anything need to be addressed to prevent the stats file from being served?

Probably. I'll take a quick look into it. Realistically, we could just do an `rm` in the CI prior to deployment. But that probably isn't the best option. Something that might be better is a check, e.g. `if [ "${NODE_ENV}" == "PRODUCTION" ] ; then craco build; else craco build --stats; fi`. That works on mac/linux, but might not work on Windows.

With that said, I _think_ there is a ticket open for switching to Terraform and there is a repo for that.

HelNershingThapa:
@tsmock, would it be okay to merge the pull request and address the --stats flag in a subsequent update?

tsmock:
Works for me. I'm currently checking to make certain that tests didn't get borked on my last rebase.

tsmock:
I've removed the `--stats` and done a few (3) test runs. The tests have been passing.

tsmock:
We might want to hold off on merging this -- I just fixed a bug in @placemarkio/geo-viewport w.r.t. webpack 5. The fix has been merged, but is not yet released.

If you run `yarn build` and get `Module not found: Error: Default condition should be last one`, that is the problem that is fixed in current geo-viewport `main` but is not yet released.

tsmock:
`geo-viewport` released a fixed version, and I've updated it in this PR.

HelNershingThapa:
The tests pass successfully, but the CircleCI process `Run yarn test` exists before generating the build with `Exited with code exit status 1`.

tsmock:
Looking at https://github.com/jestjs/jest/issues/9324, it seems like there might be a test that is doing something wrong that might be causing the failure. I'm currently in the process of bisecting the test files, and I'm _probably_ going to have to do that multiple times.

varun2948:
Hi @tsmock what's the view of yours on transitioning to vite? Will it be a lot of work or something i don't know ? Just a query

tsmock:
Doing it manually will be a lot of work. I looked into it a bit awhile ago (sometime last year). I don't remember why I dropped it (it might have been not wanting to have to do another massive PR while I had one in progress).

In my investigations, I found that someone wrote a script to eject to vite (https://github.com/bhbs/viject), so that will take care of the bulk of the work.

Things to note:
* We use flow in some locations. We can use flow comments and/or convert to something else (typescript, jsdoc, etc). I think I converted flow typing to flow comments in one of my PRs. I don't remember which one. It isn't that difficult, and still works with flow (we literally just have to wrap the flow typing with `/*` and `*/`).
* We'll probably want to take advantage of SSR sometime (note: despite standing for "server-side rendering", it can be used to prerender static content to reduce resource usage on the client side).
* I _think_ we'll have to/want to use a different library for routing. IIRC, vite has its own routing library. I might be thinking of a different project though.

varun2948:
Hmmm the viject tools seems interesting, i will have a look into it soon. But yeah so for now we will be moving ahead with this PR ? and then onwards try to switch it to vite as i think we will be using alot of vite in other product of HOT aswell so maybe something to be considered ? Vite has a plugin for routing i guess which recently changed it's name to Vike.  At the moment for moving forward on the TM for vite or upgrading the react-scripts/craco what's your take , let me know further on this if its feasible or not?

tsmock:
This was largely copied from #5721 so that we could build/run on Node 18+ without `NODE_OPTIONS=--openssl-legacy-provider`, so if #5721 is merged, that would remove the need for this PR.

TBH, I was using `craco` to work around problems in `react-scripts` (which is effectively no longer maintained). This is by no means the optimal path.

What I would do is the following:
* Merge #5721 
* Run viject
* Fix problems
* Merge once things work "well enough" -- I think it would be a PITA to constantly rebase, since one of the changes done by viject is renaming files. I believe viject will work with react-scripts + craco.

Either way, I would strongly advocate moving away from `react-scripts`. It doesn't matter if it is next.js (Vercel), remix (Shopify), gatsby (Netlify), or vite.

With that said, it _sounds_ like other HOT projects are using vite. So that is probably the winner.

varun2948:
Hmmm sounds good will let ramya know on this aswell.

ramyaragupathy:
Still relevant @varun2948 ?

PR #5914, opened 2023-06-13
Use lazy loading for routes

tsmock:
This reduces the amount of date that must be transmitted in order for the initial page (whatever it is) to load.

For the purposes of this analysis, I am using statoscope.tech with download speeds equivalent to fast 3g (1.4 MBit/s).

||Total Size|Initial Size|Initial download time|Packages Size|Build Time|Modules|Duplicate Modules|Chunks|Assets|Packages|Package Copies|
| -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- |
|de65ce14c40f13583b4c0f26e8967630056f21e9|6.39 mb|17.33 mb|101.2s|23.49 mb|103.8s|4000|13|1597|3244|272|5|
|e11bc43bed0bc9c2ad8315aea33c3b54c5140052|3.70 mb|7.62mb|44.5s|23.48 mb|237.6s|4033|13|1631|3335|272|5|
|e11bc43bed0bc9c2ad8315aea33c3b54c5140052 with #5908|3.64 mb|3.64 mb|21.3s|23.15 mb|37.6s|4020|0|1638|1689|265|6|

Note: Build time seems to be highly variable between runs; de65ce14c40f13583b4c0f26e8967630056f21e9 took 75.2s and e11bc43bed0bc9c2ad8315aea33c3b54c5140052 took 76.9s in one run.

Initial chunks:

|Commit|File|Parsed Size|File Size|Download Time|
|--|--|--|--|--|
|de65ce14c40f13583b4c0f26e8967630056f21e9|main|3.29 mb|3.24 mb|18.9s|
|e11bc43bed0bc9c2ad8315aea33c3b54c5140052|main|2.55 mb|2.79 mb|16.3s|
|de65ce14c40f13583b4c0f26e8967630056f21e9|runtime-main|0|51.05kb|291ms|
|e11bc43bed0bc9c2ad8315aea33c3b54c5140052|runtime-main|0|52.99 kb|302ms|
|de65ce14c40f13583b4c0f26e8967630056f21e9|vendor packages|6.33 mb|3.09 mb|18.1s|
|e11bc43bed0bc9c2ad8315aea33c3b54c5140052|vendor packages|2.23 mb|881.48 kb|5.0s|

The top 3 dependencies in de65ce14c40f13583b4c0f26e8967630056f21e9 vendor packages are `refractor`, `mapbox-gl`, and `date-fns` (~3.5 mb). I believe this size is without minification.
The top 3 dependencies in e11bc43bed0bc9c2ad8315aea33c3b54c5140052 vendor packages are `@sentry`, `react-select`, and `@formatjs` (~1.2 mb). Those are also in de65ce14c40f13583b4c0f26e8967630056f21e9.

Both of the vendor packages are required for actual page content; it includes `react-dom` and `react-router`.

The `main` chunk is largely made up of locales (~67% of the size is from locale data for e11bc43bed0bc9c2ad8315aea33c3b54c5140052, and ~45% of the size is from locale data for de65ce14c40f13583b4c0f26e8967630056f21e9). Another PR should ensure we only load the locale(s) we need; this would reduce the amount of data sent for locales by 95% or more.

### DIFF
--- a/frontend/src/routes.js
+++ b/frontend/src/routes.js
@@ -1,110 +1,436 @@
-import { lazy } from 'react';
 import { createBrowserRouter, createRoutesFromElements, Route } from 'react-router-dom';
 
 import { Root } from './views/root';
-import { Home } from './views/home';
-import { AboutPage } from './views/about';
-import { LearnPage } from './views/learn';
-import { QuickstartPage } from './views/quickstart';
-import { UserDetail } from './views/userDetail';
-import {
-  UserProjectsPage,
-  ManageProjectsPage,
-  CreateProject,
-  ProjectsPage,
-  ProjectsPageIndex,
-  MoreFilters,
-  ProjectDetailPage,
-} from './views/project';
 import { Authorized } from './views/authorized';
-import { Login } from './views/login';
-import { Welcome } from './views/welcome';
-import { Settings } from './views/settings';
-import { ManagementPageIndex, ManagementSection } from './views/management';
-import {
-  ListOrganisations,
-  CreateOrganisation,
-  EditOrganisation,
-} from './views/organisationManagement';
-import { OrganisationDetail } from './views/organisationDetail';
-import { OrganisationStats } from './views/organisationStats';
-import { MyTeams, ManageTeams, CreateTeam, EditTeam, TeamDetail } from './views/teams';
-import { ListCampaigns, CreateCampaign, EditCampaign } from './views/campaigns';
-import { ListInterests, CreateInterest, EditInterest } from './views/interests';
-import { ListLicenses, CreateLicense, EditLicense } from './views/licenses';
-import { Stats } from './views/stats';
-import { UsersList } from './views/users';
 import { NotFound } from './views/notFound';
-import { SelectTask } from './views/taskSelection';
-import { MapTask, ValidateTask } from './views/taskAction';
-import { EmailVerification } from './views/verifyEmail';
-import { ProjectStats } from './views/projectStats';
-import { ContactPage } from './views/contact';
-import { SwaggerView } from './views/swagger';
-import { ContributionsPage, ContributionsPageIndex, UserStats } from './views/contributions';
-import { NotificationsPage } from './views/notifications';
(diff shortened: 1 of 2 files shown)


HelNershingThapa (approved):
This is great! I'm impressed with the reduction in the initial page load size you achieved with your changes. I have a couple of curiosities regarding your approach:

I'm curious about your thought process behind merging similar chunks into one (like "CreateLicense," "ListLicenses," and "EditLicense" into a single "license" chunk). I understand that we could have generated separate chunks for each route, but I'm interested in understanding the rationale behind your choice.

Additionally, in the PR description, the build time for https://github.com/hotosm/tasking-manager/commit/e11bc43bed0bc9c2ad8315aea33c3b54c5140052 is 237.6 seconds, which seems unusually high compared to the 61.5 seconds I observed when generating the build. Could you provide insights on why there might be such a significant difference in build times?

tsmock:
It doesn't look like splitting out additional chunks is currently possible; it _might_ be possible with `"type": "module"` in `package.json`, but that doesn't work for us right now.

HelNershingThapa:
Thank you for the explanation. I appreciate your approach to merging similar chunks based on file structure. It makes sense to keep related routes together in the same chunk, and splitting out `ListLicense` into its chunk could be beneficial if it doesn't require the components used by `CreateLicense` and `EditLicense`.

I must admit that I'm not well acquainted with these magic comments and webpack configurations myself, so your insights are valuable.

tsmock:
> It makes sense to keep related routes together in the same chunk, and splitting out ListLicense into its chunk could be beneficial if it doesn't require the components used by CreateLicense and EditLicense.

Right now, we would have to move `ListLicense` into its own file. We might not need to in the future though. For now, I think we would only want to split up particularly large and important routes. Except the heaviest components will still be in the more important routes (for example, react-md-editor @ 1.7 MB is used by ProjectDetailPage and ProjectEdit).

> I must admit that I'm not well acquainted with these magic comments and webpack configurations myself, so your insights are valuable.

Neither am I. I just did a lot of research.

PR #5920, opened 2023-06-15
Refactor request body validation using decorator

### LINKED ISSUE #5910: Return detailed request validation error messages
Enhance the error messages returned during request validation. Provide specific information about the validation failure, such as missing or invalid parameters, so that frontend can show users where and what actually went wrong. This will also assist developers in troubleshooting and resolving issues.

Tasks:

- [x] Create a decorator to validate request against provided dto class.  #5920
- [ ] Apply the decorator on route functions to validate request.

Aadesh-Baral:
This PR adds the necessary component to fix #5910 
- Creates a decorator which can be used to validate the request body against a DTO class.
    - This decorator can be applied to the route handler that requires request validation.
    - This decorator validates the request using the provided DTO class and stores the validated DTO instance in request.validated_dto.
- Adds a function to fetch validation errors from Schematics `DataError` exception class so  that we can return what and where error occurred during request validation.

**Sample validation error :**
```python
@validate_request(ProjectCreateDTO)
def post():
validated_dto = request.validated_dto # request.validated_dto will be set in the decorator after validating request against dto ProjectCreateDTO
```
If the validation error occurs due to missing project description and instruction we will have error response like:
```json
{
  "error":{
    "code": 400,
    "sub_code": "BAD_REQUEST",
    "message": "The request was invalid or cannot be otherwise served."
    "details": {
      "field_errors": {
        "field": "description", "message": "This field is required",
        "field": "instruction", "message": "This field is required"
      }
    }
  }
}
```

**Benefits:**
1. Now we can just add a decorator on the route handler (like in the example above) instead of following code to validate the request body.
```python
try:
    dto = ProjectCreateDTO(request.get_json())
    dto.validate()
except DataError as e:
   return {"Error": str(e), "SubCode": "BAD_REQUEST"}
````
2. The error message will have more information like what and where the issue occured
3. This approach improves code organization, promotes code reuse, and enhances readability.

**_NOTE_**: This decorator must me only used after decorator `token_auth.login_required` if both decorator are to be used.
```python
@token_auth.login_required
@validate_request(ProjectCreateDTO)
def post():
validated_dto = request.validated_dto 
```

### DIFF
--- a/backend/models/dtos/__init__.py
+++ b/backend/models/dtos/__init__.py
@@ -13,16 +13,43 @@ def get_validation_errors(e):
     ]
 
 
-def validate_request_body(dto_class):
-    """Decorator to validate request body against a DTO class"""
+def validate_request(dto_class):
+    """
+    Decorator to validate request against a DTO class.
+    --------------------------------
+    NOTE: This decorator should be applied after the token_auth decorator (if used) so that
+    the authenticated user id can be set on the DTO if the DTO has a user_id field.
+    --------------------------------
+    Parameters:
+        dto_class: DTO
+            The DTO class to validate against
+    """
 
     def decorator(f):
         @wraps(f)
         def wrapper(*args, **kwargs):
+            from backend.services.users.authentication_service import token_auth
+
             try:
-                dto = dto_class(request.get_json())
+                dto = dto_class()
+
+                # Set attribute values from request body, query parameters, and path parameters
+                for attr in dto.__class__._fields:
+                    if request.is_json and attr in request.json:
+                        setattr(dto, attr, request.json[attr])
+                    elif attr in request.args:
+                        setattr(dto, attr, request.args.get(attr))
+                    elif attr in kwargs:
+                        setattr(dto, attr, kwargs[attr])
+
+                #  Set authenticated user id if user_id is a field in the DTO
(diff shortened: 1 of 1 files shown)


backend/models/dtos/__init__.py
```diff
@@ -0,0 +1,33 @@
+from functools import wraps
+from flask import request
+from schematics.exceptions import DataError
+
+
+from backend.exceptions import BadRequest
+
+
+def get_validation_errors(e):
```
eternaltyro, line 9: @Aadesh-Baral  Can you give me a sample of `e` passed to `get_validation_errors`?

eternaltyro (commented):
Just need some context to understand the change.

d-rita (approved):
Given that this will give more context for validation errors, LGTM

Aadesh-Baral:
@d-rita @eternaltyro  I modified the decorator, so it now validates the DTO class using all of the path, query, and body parameters rather than just the body parameters. Could you kindly re-review the PR? The decorator is now more broadly applicable and may be used on views that takes path and query params. Take a look at the commit message for further details. 

**_I apologize for making changes following the review._**

eternaltyro (approved):

PR #5631, opened 2023-03-13
Add Project-Teams API tests

d-rita:
This PR handles project-teams related API endpoint tests.

**How to test:**
- create a test database following the instructions [here](https://github.com/hotosm/tasking-manager/blob/develop/docs-old/setup-development.md#create-the-database-user-and-database). The database name is preceded by `test_`
- pull and checkout test branch using `git fetch origin && git checkout chore/add-project-teams-tests`
- run tests using `python3 -m unittest discover tests/backend`

PR #5927, opened 2023-06-21
Apply newly defined `NotFound` exceptions to handle resource not found errors.

Aadesh-Baral:
This PR does the floowing:
- Apply new Notfound Exception replacing the prev implementation. i.e  `backend.exceptions.NotFound` (defined in #5909) replaces  `backend.models.postgis.utils.Notfound` all over the code.
- Removes NotFound exception from try except blocks as this new exception will be handled by flask itself and desired error response will be created on the exception class itself with the use of `subcode` and other provided parameters.
- Updates test cases to reflect the change.

**How to review:**
Going through each commit in turn is advised because this PR has numerous file modifications. The change is explained in the commit message.


Aadesh-Baral:
Before moving on to the next stages of this change, #5933 must be merged.

d-rita (commented):
Tested almost all endpoints (except annotations and some…) and the Not Found exception handler works very well.&nbsp;
However, I noted some issues (some of which may be beyond the scope of this PR as well) but have to be noted either way:

API Resource | Endpoint | Method | Issues | Status
-- | -- | -- | -- | --
Tasks | projects/project_id/tasks | GET | Getting particular tasks: non-existent task results in an internal server error | Fixed
Tasks | stop-validation | POST | Does not show 404 for non-existent mapping issue category | Out of scope
Tasks | tasks/xml, tasks/gpx | GET | Does not include non-existent task_id in error message | Fixed
Teams | project/teams | GET | Returns empty list for non-existent project | Fixed
Teams | project/teams/team_id | DELETE | 1. Docstrings do not include the project_id and team_id parameters. 2. user_id is not being captured hence resulting in an internal server error. | Out of scope
Teams | teams/team_id/actions/leave | POST | Remove existent but non-member from a team results in an internal server error | Fixed
Comments | projects/project_id/comments/tasks/task_id | POST, DELETE | Does not check if the project exists first. Hence error message does not show non existent project | Fixed
Projects | /projects | GET | Getting all organisations with varying non existent parameters results in an empty list| Needs Investigation (Out of scope)
Projects | /projects/project_id (get project including its area) | GET | Does not serve the 404 error. It is logged but it gives an internal server error for non existent projects | Fixed
Projects | projects/proj_id/actions/transfer-ownership | POST | Returns internal server error for non-existent project | Fixed
Projects | projects/proj_id/queries/nogeometries |   | Internal server error for non-existent project | Fixed
Organisations | /organisations | POST | 1. Add type parameter to docstrings since it is required. 2. If other parts of the request body already exist, it gives the same error, i.e. Name already exists. | Needs Investigation (Out of scope)
Organisations | /organisations/org_id | GET | Non existent org _id not availed in error message (which is inconsistent with the PATCH method 404 error where it is shown) | Fixed
Organisations | organisations/org_slug | GET | Non existent org_slug not availed in 404 error | Fixed
Organisations | /organisations | GET | Returns empty list for non-existent manager parameter | Needs Investigation (Out of scope)
Messages | notifications/message_id | GET, DELETE | MessageService.get_message() - raise NotFound | Fixed
Campaigns | organisations/org_id/campaigns/ | GET | Returns empty list for non-existent organisation | Fixed
Campaigns | organisations/org_id/campaigns/campaign_id | POST | Internal server error for non existent campaignInternal server error for non existent organisation | Fixed
Campaigns | organisations/org_id/campaigns/campaign_id | DELETE | Internal server error if both org and campaign are real but unassigned. | Fixed
Campaigns | projects/proj_id/campaigns/campaign_id | DELETE | Generic Not found for existing campaign and project that are not attached | Fixed
Comments | projects/project_id/comments/tasks/task_id | GET | Remove body from dosctrings | Out of scope
Grid | projects/actions/intersecting-tiles |   | Docstrings don’t show all required fields in the example schema. | Out of scope

backend/services/mapping_service.py
```diff
@@ -195,7 +200,11 @@ def add_task_comment(task_comment: TaskCommentDTO) -> TaskDTO:
         """Adds the comment to the task history"""
         task = Task.get(task_comment.task_id, task_comment.project_id)
```
d-rita, line 201: Would probably be good to check for the project first. The Not Found error only reflects if the task is not found even if the issue might be a non-existent project.

backend/services/messaging/chat_service.py
```diff
@@ -122,7 +126,11 @@ def delete_project_chat_by_id(project_id: int, comment_id: int, user_id: int):
             ProjectChat.id == comment_id,
         ).one_or_none()
         if chat_message is None:
```
d-rita, line 128: Same here. Does not show if the project not existing might be the issue.

backend/models/postgis/message.py
```diff
@@ -156,7 +158,10 @@ def get_all_messages(user_id: int) -> MessagesDTO:
         user_messages = Message.query.filter(Message.to_user_id == user_id).all()
 
         if len(user_messages) == 0:
-            raise NotFound()
+            raise NotFound(
+                sub_code="NO_MESSAGES_FOUND",
```
d-rita, line 162: Should be MESSAGE_NOT_FOUND according to the error_messages.json

d-rita (changes requested):

d-rita (commented):
We don't have tests for the Project Teams but the Project Teams API endpoints are failing since the user_id is not being read. That issue is solved for in this PR  https://github.com/hotosm/tasking-manager/pull/5631 containing the tests and fix.
cc: @Aadesh-Baral

Aadesh-Baral:
@d-rita Is https://github.com/hotosm/tasking-manager/pull/5631 ready for review?

d-rita (approved):
LGTM

Issue #5912, opened 2023-06-13
Get rid of excessive use of try except blocks for error handing.


Aadesh-Baral:
Eliminate the unnecessary use of try-except blocks throughout the codebase. Instead, leverage Flask's error handler to handle exceptions centrally. This will help streamline error handling and promote consistency across the application.

PR #5933, opened 2023-06-22
Remove unhandled exception handling blocks.

Aadesh-Baral:
Part of #5912 
As all of the unhandled errors will be now captured and processed by flask errorhandler defined on #5909, we no longer have to handle them manually using try except blocks. This PR removes these blocks that were set to handle these unhandled exceptions. 

https://github.com/hotosm/tasking-manager/blob/9582691a4ac2acc8a50b94deae82dd2a97ae7fa7/backend/__init__.py#L116-L147


d-rita (approved):
Tested endpoints via the swagger docs. Endpoints work fine. Odd cases that stood out were not related to this implementation. LGTM.

PR #5937, opened 2023-06-22
Easy dependency updates

tsmock:
This updates dependencies and dependencies of dependencies ("deep" dependencies) that do not require code modification.

Let me know if I need to squash the commits; they were done dependency-by-dependency to make bisecting problems easier.

Not all dependencies and their deep dependencies were updated. Specifically `@testing-library/jest-dom` and `@testing-library/react` dependencies were not updated because it caused `dom-accessibility-api` to be updated to a known-bad version (for us).

PR #5948, opened 2023-06-27
Add @babel/plugin-proposal-private-property-in-object to devDependencies

HelNershingThapa:
Ran into the following issue. 

```
One of your dependencies, babel-preset-react-app, is importing the "@babel/plugin-proposal-private-property-in-object" package without declaring it in its dependencies. This is currently working because "@babel/plugin-proposal-private-property-in-object" is already in your node_modules folder for unrelated reasons, but it may break at any time.

babel-preset-react-app is part of the create-react-app project, which is not maintianed anymore. It is thus unlikely that this bug will ever be fixed. Add "@babel/plugin-proposal-private-property-in-object" to your devDependencies to work around this error. This will make this message go away.
```

This PR is the workaround to make the message go away.

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -108,6 +108,7 @@
     ]
   },
   "devDependencies": {
+    "@babel/plugin-proposal-private-property-in-object": "^7.21.11",
     "@testing-library/jest-dom": "^5.16.5",
     "@testing-library/react": "^12.1.5",
     "@testing-library/react-hooks": "^8.0.1",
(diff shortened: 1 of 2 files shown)


tsmock:
Fair enough. I didn't do this in #5937 since it was fixed in `react-scripts`, and I was really hoping that they would do a release ASAP to fix it. Realistically we'll probably want to move off of CRA once we are on Node 18 and have a (relatively) stable and up-to-date base to work off of.

HelNershingThapa:
Had they done the release, it'd have been a more convenient solution for us. 

> Realistically we'll probably want to move off of CRA once we are on Node 18 

Next or Vite maybe?

PR #5950, opened 2023-06-27
Use https for outlink

### LINKED ISSUE #5925: [BUG] Broken website link of HOT
**Describe the bug**
The website link provided on the landing page appears to be broken.

**To Reproduce**
1. Go to 'https://tasks.hotosm.org/'
2. Click on '[hotosm.org](http://hotosm.org/)' link in top right

**Screenshots**
ezgif com-video-to-gif(4)

dakotabenjamin:
Partially fixes #5925 

This PR updates the protocol to https. I have also updated the ORG_URL in CircleCI Context for staging and production to be "www.hotosm.org"

The reason we can't easily setup a redirect is [from AWS](https://docs.aws.amazon.com/Route53/latest/DeveloperGuide/resource-record-sets-choosing-alias-non-alias.html):
> Unlike a CNAME record, you can create an alias record at the top node of a DNS namespace, also known as the zone apex. For example, if you register the DNS name [example.com](http://example.com/), the zone apex is [example.com](http://example.com/). You can't create a CNAME record for [example.com](http://example.com/), but you can create an alias record for [example.com](http://example.com/) that routes traffic to [www.example.com](http://www.example.com/) (as long as the record type for [www.example.com](http://www.example.com/) is not of type CNAME).

If this is not sufficient, we may be able to setup a redirect, but I would prefer not to, as this will likely change soon with the new hotosm website.

### DIFF
--- a/frontend/src/components/banner/index.js
+++ b/frontend/src/components/banner/index.js
@@ -34,7 +34,7 @@ export function Banner() {
 
   const privacyPolicyLink = (
     <a
-      href={`https://${ORG_PRIVACY_POLICY_URL}`}
+      href={`${ORG_PRIVACY_POLICY_URL}`}
       className="red underline link fw6"
       target="_blank"
       rel="noopener noreferrer"
@@ -53,7 +53,7 @@ export function Banner() {
           <a
             id="privlink"
             className="red link f4 fw6"
-            href={`https://${ORG_PRIVACY_POLICY_URL}`}
+            href={`${ORG_PRIVACY_POLICY_URL}`}
             target="_blank"
             rel="noopener noreferrer"
           >
--- a/frontend/src/components/footer/index.js
+++ b/frontend/src/components/footer/index.js
@@ -113,7 +113,7 @@ export function Footer() {
             </Link>
             {ORG_PRIVACY_POLICY_URL && (
               <div className="pt2 f6 lh-title">
-                <a href={`https://${ORG_PRIVACY_POLICY_URL}`} className="link white">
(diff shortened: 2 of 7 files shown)


HelNershingThapa (commented):
> I have also updated the ORG_URL in CircleCI Context for staging and production to be "www.hotosm.org"

This change resolves the issue at hand without requiring an additional protocol change. This change alone is sufficient to fix the problem, right?

Do you think it'd be important to consider the impact on forks that rely on HTTP rather than HTTPS for their sites? They may face difficulties in updating their site links. Is it accurate to say that this change could potentially affect these forks?

Instead of explicitly specifying the protocol in the code, why not retrieve the complete URL from an environment variable? By doing so, the protocol can be dynamically set based on the environment, providing more flexibility. Thoughts?

dakotabenjamin:
You are correct that this change is not required to fix the issue, we just need to push a deployment with the new context envvar. 

However, it is my personal opinion that we should not be supporting HTTP in general. Your suggestion to make the whole url with the protocol a part of the envvar is well taken, and I will push a change to reflect that. We will need to parse out the domain to make it look pretty, though, or call it "HOTOSM Homepage" or something to that effect. What do you think?

dakotabenjamin:
New change looks like this: 
image

HelNershingThapa (commented):
This test case should have failed, but it did not 🤨.

https://github.com/hotosm/tasking-manager/blob/fc5f3903d129abfced457defb18297d5bc007a13/frontend/src/components/header/tests/index.test.js#L25-L28

HelNershingThapa (approved):

PR #5955, opened 2023-06-29
Add feature to revert tasks set as unavailable by a specific user

### LINKED ISSUE #5943: Revert unavailable tasks set by a specific user
At present, we let advanced mappers mark bad imagery tasks. However, still the [validators notice some discrepancy](https://github.com/hotosm/tasking-manager/issues/5389) and spend a lot of time reverting these tasks to available state. An advanced user might have accidentally used a wrong imagery source or some other technical issue with the chosen imagery source, resulting in false positives.

Similar to reverting validated tasks of a specific contributor, a project manager/admin should have the ability to revert bad imagery tasks of a specific contributor.
image

Aadesh-Baral:
closes #5943 

Backend changes:
- Rename endpoint `projects/<int:project_id>/tasks/actions/reset-validated-by-user/` to `projects/<int:project_id>/tasks/actions/reset-by-user/`.
- Introduce a query param `action` in the endpoint above which is used to revert tasks of specific `action` performed by the user. The value of this action can only be `VALIDATED` and `BADIMAGERY`.
- Add and update test case for endpoint and its associated function.
- Better handling of exceptions on the particular endpoint.
- Add `badImagery` and `badImageryTasks` in the response of `projects/<project_id>/contributions/` endpoint.

unavailable_tasks


d-rita (approved):
LGTM!

PR #5966, opened 2023-07-06
Update react dropzone

tsmock:


HelNershingThapa (commented):
When interacting with the markdown editor container, an unintended behavior occurs where clicking on it triggers the file selection dialog. This issue seems to be specific to Chrome and does not occur in Firefox
filezone

tsmock:
I forgot about that, especially since I test in Firefox, not chrome. I think I fixed that in the massive PR, I'll just have to find it.

HelNershingThapa (approved):

PR #5978, opened 2023-07-12
Revert "Keep the footer at the bottom on a short page"

HelNershingThapa:
Reverts hotosm/tasking-manager#5930


HelNershingThapa:
Noticed a horizontal scrollbar when testing on the staging instance, so reverted it

scrolling

PR #5985, opened 2023-07-16
Make GPX name attribute string translatable

### LINKED ISSUE #5650: GPX name attribute cannot be translated
**Describe the bug**
When opening a task to edit, the box surrounding the AOI displays: "Task for project xxxx. Do not edit outside of this area!".

I would like to translate that string, and could not find here on Github neither on Transifex. Do you have more info where I can translate that?

This happens both on iD and RapiD (did not test on JOSM), but I suspect this string comes from the .gpx (attribute _name_).

Thanks!

d-rita:
Fixes #5650


HelNershingThapa (approved):
Tested the changes locally; translated strings appear. LGTM.

PR #6001, opened 2023-07-25
Don't allow creating/updating interest with empty name

### LINKED ISSUE #5946: [BUG] Interest/category with empty name can be created
By using the API, it is possible to create an interest or category without giving it a name, although doing so would break the manage categories page on front end.

**Expected behavior**
Creating a interest/category with empty name should not be possible. And also on the manage categories page this scenario should be handled.

Aadesh-Baral:
closes #5946


d-rita (approved):
Works as expected. LGTM

PR #6032, opened 2023-08-17
Add sentry replays and source maps

dakotabenjamin:
This PR adds sentry replays and source maps for improved frontend tracking

### DIFF
--- a/frontend/package.json
+++ b/frontend/package.json
@@ -15,8 +15,6 @@
     "@mapbox/mapbox-gl-language": "^0.10.1",
     "@placemarkio/geo-viewport": "^1.0.1",
     "@sentry/react": "^7.60.1",
-    "@sentry/tracing": "^7.60.1",
-    "@sentry/browser": "^7.60.1",
     "@tmcw/togeojson": "^4.7.0",
     "@turf/area": "^6.5.0",
     "@turf/bbox": "^6.5.0",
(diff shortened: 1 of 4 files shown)


frontend/package.json
```diff
@@ -14,8 +14,9 @@
     "@mapbox/mapbox-gl-geocoder": "^5.0.1",
     "@mapbox/mapbox-gl-language": "^0.10.1",
     "@placemarkio/geo-viewport": "^1.0.1",
-    "@sentry/react": "^7.56.0",
-    "@sentry/tracing": "^7.56.0",
+    "@sentry/react": "^7.60.1",
+    "@sentry/tracing": "^7.60.1",
+    "@sentry/browser": "^7.60.1",
```
HelNershingThapa, line 19: I couldn't find any instances where `@sentry/browser` package is being utilized. From the React | Sentry Docs 

> All methods available in `@sentry/browser` can be imported from @sentry/react.

Since we are already using `@sentry/react` in our project, it might be a good idea to consider removing the unused `@sentry/browser`.

frontend/src/index.js
```diff
@@ -16,8 +16,22 @@ if (SENTRY_FRONTEND_DSN) {
   Sentry.init({
     dsn: SENTRY_FRONTEND_DSN,
     environment: ENVIRONMENT,
-    integrations: [new BrowserTracing()],
+    integrations: [
+      new BrowserTracing(),
```
HelNershingThapa, line 20: On a separate note (unrelated to the changes made in this PR), the `BrowserTracing` import from `@sentry/tracing` is now deprecated. It's recommended to directly import it from our framework-specific SDK, `@sentry/react`. This lets us remove the `@sentry/tracing` package altogether.

PR #6046, opened 2023-09-01
Fix Missing task extension duration in action_text.

### LINKED ISSUE #6014: [BUG] Missing Task Lock Extension Duration in Action Text Column
Currently, our system stores an "action period" in the action_text column to track the duration of a user's task lock for mapping or validating. However, we have identified an issue with the implementation of the task lock extension feature.

The problem lies in the fact that we are not saving the information about how long the task was extended for. As a result, the action_text column remains empty when the task is in the "EXTENDED_FOR_MAPPING" status.

Here's an example of the issue:

| ID  | Task ID | User ID | Status            | Action Text   | Timestamp                 | Project ID |
| --- | ------- | ------- | ----------------- | ------------- | ------------------------- | ---------- |
| 1735| 11      | 6       | LOCKED_FOR_MAPPING| 00:01:11.887191| 2023-08-03 10:14:32.310950| 13526430   |
| 1736| 11      | 6       | EXTENDED_FOR_MAPPING| None        | 2023-08-03 10:15:44.211045| 13526430   |
| 1739| 11      | 6       | STATE_CHANGE      | MAPPED        | 2023-08-03 10:21:48.759124| 13526430   |

As you can see, when the task status changes to "EXTENDED_FOR_MAPPING," there is no duration information provided in the action_text column, unlike when the status is "LOCKED_FOR_MAPPING."

This missing data is crucial for tracking the total duration of task locks accurately, especially when a task is extended multiple times.

Proposed Solution:
To resolve this issue, we need to update our system to include the duration of the task lock extension in the action_text column when the status is "EXTENDED_FOR_MAPPING."

1. When a task lock is extended, calculate the duration of the extension.
2. Ensure that the duration is properly updated each time the task lock is extended.

**Steps to Reproduce:**
1. Lock a task for mapping/validation.
2. Observe that the action_text column records the lock duration.
3. Extend the task lock before its expiry.
4. Unlock the task.
5. Check the action_text column for the extended task; notice that the duration of the extension is not recorded.

Aadesh-Baral:
closes #6014 

In the previous implementation, there was an issue where the duration of a task extension was not being recorded when a task was extended and then unlocked/stopped. Prior to the implementation of the task lock extension feature, the last recorded action for a task was always either "LOCKED_FOR_MAPPING" or "LOCKED_FOR_VALIDATION," which matched the task status. To update the duration, we used the current task status as the last action to filter the task history and update the last entry with the lock duration.

However, after the task extension feature was introduced, two additional last actions were possible: "EXTENDED_FOR_MAPPING" and "EXTENDED_FOR_VALIDATION." This introduced a discrepancy between the last recorded action and the task status. For example, the task status could still be "LOCKED_FOR_MAPPING" even when a user had extended the task. As a result, the entry with the action "EXTENDED_FOR_MAPPING" in the task history was not being updated with the correct duration.

To address this issue, this PR fixes the problem by retrieving the last task action and using it as the filter criteria for updating the task history, rather than relying on the task status as the last action.

#### Cases:
- **Task is stopped mapping/validating after extension.** -> The entry with extension is updated with extension duration.
- **Task's state is changed after extension.** -> The entry with extension is updated with duration and new entry with state changed is added.
- **Task's lock is extended multiple times.**  -> After each extension previous entry with extension is updated with the duration and new entry with extension is added.
- **Task is auto unlocked after extension.** -> The entry with extension is updated with duration and name of the action is changed from "EXTENDED_FOR_MAPPING" to "AUTO_UNLOCKED_FOR_MAPPING".

### DIFF
--- a/backend/models/postgis/task.py
+++ b/backend/models/postgis/task.py
@@ -784,6 +784,16 @@ class Task(db.Model):
         self, user_id, new_state=None, comment=None, undo=False, issues=None
     ):
         """Unlock task and ensure duration task locked is saved in History"""
+        #  If not undo, update the duration of the lock
+        if not undo:
+            last_history = TaskHistory.get_last_action(self.project_id, self.id)
+            # To unlock a task the last action must have been either lock or extension
+            last_action = TaskAction[last_history.action]
+            TaskHistory.update_task_locked_with_duration(
+                self.id, self.project_id, last_action, user_id
+            )
+
+        # Only create new history after updating the duration since we need the last action to update the duration.
         if comment:
             self.set_task_history(
                 action=TaskAction.COMMENT,
@@ -822,12 +832,6 @@ class Task(db.Model):
             self.mapped_by = None
             self.validated_by = None
 
-        if not undo:
-            # Using a slightly evil side effect of Actions and Statuses having the same name here :)
-            TaskHistory.update_task_locked_with_duration(
-                self.id, self.project_id, TaskStatus(self.task_status), user_id
-            )
-
         self.task_status = new_state.value
         self.locked_by = None
         self.update()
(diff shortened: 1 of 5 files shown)


ramyaragupathy:
@Aadesh-Baral - could you confirm if this is ready for review? cc @kaditya97 @prabinoid @manjitapandey

Aadesh-Baral:
@ramyaragupathy Sorry for late reply, Yes this is ready for review 😄

ramyaragupathy:
@Aadesh-Baral - would you be able to look at this? I assume this is going to be a heavy lift because of the recent codebase migration. If this is not possible, please give a heads-up. We will plan alternatives.

cc @prabinoid @suzit-10 @manjitapandey

ramyaragupathy:
Fixed through https://github.com/hotosm/tasking-manager/pull/7214

Issue #5767, opened 2023-05-08
[BUG] NodeJS v16 used for frontend is EOL in September 2023


eternaltyro:
Tasking Manager frontend runs on NodeJS v16.x which is set to be EOL on 11 September 2023. We need to reprioritise our tasks to focus on making the frontend compatible with v18 - the current LTS version of NodeJS. This will give us one year before we need to upgrade again.

https://endoflife.date/nodejs

ramyaragupathy:
Per discussion w/ @HelNershingThapa this will be addressed through https://github.com/hotosm/tasking-manager/pull/5721

ramyaragupathy:
with https://github.com/hotosm/tasking-manager/pull/5721 - moved to node 18 cc @varun2948 @royallsilwallz @manjitapandey

PR #6053, opened 2023-09-08
Add a temporary workaround to get building on Node 18 since Node 16 EOL is 2023-09-11

tsmock:
This does the following:
* Add NODE_OPTIONS=--openssl-legacy-provider (not needed once we upgrade to react-scripts v5 or remove/replace react-scripts)
* Only adds NODE_OPTIONS when run on node 18 or later
* Performs frontend-code-test on node 18
* Does _not_ run backend_deploy/frontend_deploy with node 18. This can be changed via using cimg/python:3.10-node (versions after 3.10.8-node use node 18 instead of node 16)

Node 16 EOL Notice: https://nodejs.org/en/blog/announcements/nodejs16-eol
create-react-app --openssl-legacy-provider: https://github.com/facebook/create-react-app/issues/11562

This is intended to be a temporary fix for #5767.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -8,7 +8,7 @@ jobs:
   frontend-code-test:
     resource_class: large
     docker:
-      - image: cimg/node:16.20
+      - image: cimg/node:18.17
     working_directory: /home/circleci/tasking-manager
     steps:
       - checkout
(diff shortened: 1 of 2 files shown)


dakotabenjamin:
We built some node images for this issue while we work to get TM running on a newer version: https://quay.io/repository/hotosm/node?tab=tags

ramyaragupathy:
@varun2948 @tsmock - with the node upgrade coming from #5721 merge is this relevant?


cc @royallsilwallz @manjitapandey

tsmock:
Nope. We can close this.

PR #6054, opened 2023-09-08
Update Rapid to v2

tsmock:
* Use URL hash params for Rapid settings
  * This reduces fragility going forward since we don't have to worry about internal Rapid changes as much
* Use a CDN for Rapid script and data
  * This is set to a semver patch range for the Rapid version. This means that if there is a patch release, we don't have to update Rapid and do a release. It was _not_ a semver minor version since there were some internal Rapid changes that broke some of the code that we were using.
* Saves dom node to avoid recreating Rapid
  * Otherwise, switching between tasks can be tricky

I've tried to document everything.


royallsilwallz:
Hey @tsmock, I've reviewed the PR and it seems to work perfectly fine. I think it is ready to merge.

Regarding the SonarCloud Code Analysis quality gate failure - the linter is showing warnings on the docker files.
I don't think this should block us from merging the code since we have manual build setup and the warning on docker files should not affect the build process.

Furthermore, @mahesh-naxa is working on fixing these error as well in [this PR](https://github.com/hotosm/tasking-manager/pull/6087).

CC @ramyaragupathy @manjitapandey @varun2948 @konishon

royallsilwallz:
Hey @tsmock, after your latest push last week, one of the test case if failing. Do you have time to look into it?

PR #6087, opened 2023-10-20
add: sonarcube suggestions on disable package install recommends

mahesh-naxa:
Fixes Sunarcube Errors and Suggestions on CI

- Make sure automatically installing recommended packages is safe here.
- Make sure automatically installing recommended packages is safe here.
- Omitting --ignore-scripts can lead to the execution of shell scripts. Make sure it is safe here.

### DIFF
--- a/scripts/docker/tasking-manager/Dockerfile
+++ b/scripts/docker/tasking-manager/Dockerfile
@@ -15,7 +15,7 @@ RUN curl -sL https://deb.nodesource.com/setup_16.x | bash -
 # Install dependencies
 RUN apt-get update \
   && apt-get upgrade -y \
-  && apt-get install -y nodejs libgeos-dev \
+  && apt-get install -y --no-install-recommends nodejs libgeos-dev \
   && apt-get clean && rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 
 # Upgrade pip
@@ -35,7 +35,7 @@ RUN pip install --no-cache-dir pdm \
     && pdm install --prod --no-lock --no-editable
 
 # Setup and build frontend
-RUN cd frontend && npm install && npm run build
+RUN cd frontend && npm install --ignore-scripts && npm run build
 
 # INITIALIZATION
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -91,7 +91,7 @@ FROM runtime as prod
 USER root
 # Get the necessary bits for the health check
 RUN apt-get update && \
-	apt-get install -y curl && \
+	apt-get install --no-install-recommends -y curl && \
 	apt-get clean && \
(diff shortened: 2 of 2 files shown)


mahesh-naxa:
Add:
- a seperate user "appuser" that runs pdm executable .

scripts/docker/tasking-manager/Dockerfile
```diff
@@ -4,18 +4,13 @@ FROM python:${PYTHON_IMG_TAG}-bookworm
 RUN mkdir -p /usr/src/app
 WORKDIR /usr/src/app
 
-ENV PATH="/usr/src/app/__pypackages__/${PYTHON_IMG_TAG}/bin:$PATH" \
```
eternaltyro, line 7: why is this being removed?

scripts/docker/tasking-manager/Dockerfile
```diff
@@ -35,7 +34,7 @@ RUN pip install --no-cache-dir pdm \
     && pdm install --prod --no-lock --no-editable
 
 # Setup and build frontend
-RUN cd frontend && npm install && npm run build
+RUN cd frontend && npm install --ignore-scripts && npm run build
```
eternaltyro, line 37: Why would we ignore scripts when we can run preinstall and postinstall scripts? I am okay with this change in general, but let us add a dummy script in package.json with key `README` and value something like `DOCKERFILE NPM INSTALL IGNORES SCRIPTS` so that in the future it will make debugging easier if someone does intend to add a pre/post install script to the manifest.

eternaltyro (changes requested):

scripts/docker/tasking-manager/Dockerfile
mahesh-naxa: @eternaltyro  Hey, i had these modified as part of security recommendation from sonar cube. i had these tested as well. But looks like /scripts/docker/tasking-manager/Dockerfile is replaced by Dockerfile.backend & Dockerfile.frontend resp. We should be looking to remove this in future.

eternaltyro: Good for now.

scripts/docker/tasking-manager/Dockerfile
eternaltyro: @mahesh-naxa

mahesh-naxa: @eternaltyro python stuff are inherited from python:3.10-bookworm base image. idk why `PATH` and `PYTHONPATH` are updated to use `/usr/src/app/__pypackages__/${PYTHON_IMG_TAG}/bin` where nothing exists. 
It was also flagged by sonarcube. these changes are non destructive.

Issue #6109, opened 2023-10-26
Add Project Data Quality view


emi420:
**Is your feature request related to a problem? Please describe.**
There's no way to way to monitor mapping in near real-time, highlighting quality issues at the same time.

**Describe the solution you'd like**
Integrate [Underpass Data Quality Tools](https://github.com/hotosm/underpass) and using [Underpass UI components](https://github.com/hotosm/underpass-ui).

**Describe alternatives you've considered**
Another alternative could be to develop this feature for Tasking Manager only, but there's no way currently of getting data and analyze it every minute without using Underpass. Also this could be done without using Underpass DB/REST API and/or Underpass UI, and develop new endpoints or a new UI, but this components already exists because they are intended to be used on other projects.

**Additional context**

There were several discussions for the design of this idea:

- https://github.com/hotosm/tasking-manager/discussions/5617
- https://github.com/hotosm/tasking-manager/discussions/5685

Finally, we arrived to the conclusion that creating a live monitoring tool, while displaying automatically detected data quality issues, have higher value than other options.

This is the draft design for the view:

![image](https://github.com/hotosm/tasking-manager/assets/1226194/ae68f865-cf3a-41bb-9870-21da4becfaee)

It could be modified when doing the final implementation.

### Technical notes

- It will be available for 4 countries (1 per Hub) in the first version, in order to have time for testing and improvement
- Should be located in: `components/projectDataQuality.js`
- It will be implemented using Underpass React UI components, including the connection with an Underpass API for getting data
- It will need need access to some data for the TM project: `id`, `areaOfInterest` and `mappingTypes` for filtering results.

emi420:
I added the view in this branch:

https://github.com/hotosm/tasking-manager/tree/feature/liveMonitoring

<img width="1181" alt="Screenshot 2023-12-06 at 14 48 58" src="https://github.com/hotosm/tasking-manager/assets/1226194/a768d308-1d81-48c6-9602-7e22b1285cee">

varun2948:
Hi @emi420  are we fetching data in certain intervals? Cause we don't have web socket integration right now to get the realtime data right ?

emi420:
Hey @varun2948 , yes, when the "live" function is checked. I'll add websockets soon to the [API](https://github.com/hotosm/underpass/tree/master/python) and the UI (https://github.com/hotosm/underpass-ui)

varun2948:
Sounds great with the websocket implementation and realtime data.

[...]

PR #6151, opened 2023-12-07
Feature/live monitoring

emi420:
This PR adds a new view for live monitoring of mapping projects, providing data quality insights.

You can see how it looks here:

https://github.com/hotosm/tasking-manager/issues/6109#issuecomment-1843383333

There are some other tasks that needs to be completed in order to have this feature in production.


frontend/src/components/projectDetail/liveViewButton.js
```diff
@@ -0,0 +1,20 @@
+import React from 'react';
+import { Link } from 'react-router-dom';
+import { FormattedMessage } from 'react-intl';
+
+import messages from './messages';
+import { CustomButton } from '../button';
+
+export const LiveViewButton = ({ projectId, className, compact = false }: Object) => (
```
varun2948, line 8: Hi Emi, I see you using Object Type in this section and wanted to know if it's necessary as this file is not typescript and Object type is too vague for the types, isn't it?

frontend/src/views/projectLiveMonitoring.js
```diff
@@ -0,0 +1,233 @@
...
+import { ProjectHeader } from '../components/projectDetail/header';
+import { useSetTitleTag } from '../hooks/UseMetaTags';
+import { useParams } from 'react-router-dom';
+import { useFetch } from '../hooks/UseFetch';
+import ReactPlaceholder from 'react-placeholder';
+import centroid from '@turf/centroid';
+import "./projectLiveMonitoring.css";
+
+const config = {
+    API_URL: "https://underpass.hotosm.org:8000",
```
varun2948, line 12: shouldn't this be on the .ENV file ?

varun2948 (approved):

frontend/src/views/projectLiveMonitoring.js
emi420: Yes!

frontend/src/components/projectDetail/liveViewButton.js
emi420: You're right, it can be removed, I think I just used code from another view as a template and it was there :)

royallsilwallz:
Hey @emi420, 
Are you finished with this PR or is there some work remaining from your side? 
I saw some warnings showing `no-unused-vars` by eslint.
Are you planning to fix them?
CC: @varun2948

royallsilwallz:
Hey @emi420 I've addressed the issues by ignoring the eslint rules for the specific lines for now. 
Need to fix them when you are available later.
CC - @varun2948

royallsilwallz:
@kshitijrajsharma - @emi420 has used underpass api url as localhost:8000.
What shall we put on TM for the underpass api url? 
CC - @varun2948

kshitijrajsharma:
@royallsilwallz , For download service its a env variable in frontend and it doesn't need any API to communicate with it will directly point to s3 ! So different and unrelated !

royallsilwallz:
@kshitijrajsharma do you mean we don't have to pass the apiUrl prop to the UnderpassFeatureStats component ?

kshitijrajsharma:
I am not following @royallsilwallz , I believe for download service frontend @varun2948 will have more context

royallsilwallz:
@kshitijrajsharma I think we need underpass api url for this to work..

varun2948:
@kshitijrajsharma This is not related to the export tool but to the live monitoring part where EMI has used a localhost on the API part as emi has used underpass UI library.

kshitijrajsharma:
@varun2948 oh okay unrelated to download service , for config best approach is it pick from env variable and fallback to localhost 
Couldn't help but notice , I wonder why we are using mapbox though instead of maplibre

emi420:
> On separate topic , Couldn't help but notice , I wonder why we are using mapbox though instead of maplibre @emi420 ?

We're using Maplibre, Mapbox API KEY is for background imagery only

emi420:
> @kshitijrajsharma - @emi420 has used underpass api url as localhost:8000 in the PR What shall we put on TM for the underpass api url? CC - @varun2948

API URL is https://underpass.hotosm.org:8000 

Currently, it has data for one country (Libya) and is a development instance, it will be replaced later once we do the deployment (there are devops tasks already created for that)

kshitijrajsharma:
> > On separate topic , Couldn't help but notice , I wonder why we are using mapbox though instead of maplibre @emi420 ?
> 
> @kshitijrajsharma we're using Maplibre, Mapbox API KEY is for background imagery only

Would it make sense to remove dependency from mapbox completely ? I think there are existing baselayers for OSM that ca n be injected on maplibre , It would be wise to remove mapbox atleast on the tools / feature we developing recently !

emi420:
> @varun2948
> 
> > where there is localhost so what do you think about that part?
> 
> oh okay unrelated to download service , for config best approach is to pick from env variable if not found fallback to localhost

@kshitijrajsharma 

Yes, ENV variables are the best approach, I just leave that for the project developers. You can pass a config property to the Underpass UI component and the values on that property will be the ENV variables. Makes sense?

manjitapandey:
I have tested the feature by integrating it on my local device setup. I encountered few problems, which i am not quite sure if they are due to Underpass UI or feature issues. The team is digging further regarding this. 

- Number of tasks are not shown complete. If there are total of 6 unsquared buildings only 4 are seen on the list.

- The pop up of buildings are still seen after filter is removed.

emi420:
>     * Number of tasks are not shown complete. If there are total of 6 unsquared buildings only 4 are seen on the list.

I think some of them are hidden because a scroll/overflow problem, I'll fix it in the Underpass UI component.

>     * The pop up of buildings are still seen after filter is removed.

I'll check this and provide a solution.

royallsilwallz:
The build file of CSS has been changed from `452.4KB` (develop branch) to `6.54MB` (feature/liveMonitoring branch).

The size is coming from `@hotosm/underpass-ui/dist/index.css` (`5.9MB`). It has 336,016 lines of code, almost all coming from Tailwind CSS.

This requires optimization. 
The solution would be to minify this file.
Also, see if there are unwanted codes in the dist files.

Were you aware of this, @emi420 ?

**Screenshot**
- `develop` VS `feature/liveMonitoring` branch
image

royallsilwallz (approved):

royallsilwallz:
I can confirm that the size is optimized now. Nice work @emi420 !
I've upgraded the latest Underpass UI repo and pushed the code. 
I think we should merge this now. What do you say ?
CC - @ramyaragupathy @manjitapandey

ramyaragupathy:
Since there are conflicts and merge that's unrelated to this PR which makes the conflict resolution difficult, we will have @emi420 create a separate branch.

Also, we will include this feature only in `expert mode`. Once the new PR is up from @emi420, @royallsilwallz will implement the changes to support the view for expert mode features

emi420:
Closing this PR because we'll create a new one.

royallsilwallz:
Reopening because it was reverted in #6306

royallsilwallz:
Continued with #6299

PR #6176, opened 2023-12-26
Feat osm download

### LINKED ISSUE #6143: Let users download OSM data generated for each project
Related to [Export Tool integration](https://github.com/hotosm/tasking-manager/issues/6094) - there should be a download option available for all signed-in users to access data exports for each project.

Per the sketch here, seeing this fit in the contribution section.

@manjitapandey @varun2948 to add the mock-ups for reference.

image

Download option swill be available on all projects(published and archived, draft need not have this option). During the initial launch we will have files generated for archived projects as well and it will be a one time task. Later, on a daily basis, we will have new exports available only for active projects that are in published status. Suppose export is not available for a project then a pop-up with the message that no data available for download should be indicated for the user.

varun2948:
I have added an Export Tool with the env needed as mentioned in the sample env we @manjitapandey  tested on our side and would love some feedback after it gets used and will improve it if needed.

- fix #6143 
image


varun2948:
Fixed sonarlint issue too now good to go

Issue #6145, opened 2023-11-30
  Export tool integration: Setting up a lambda function to track active projects and schedule export jobs


ramyaragupathy:
Related to [Export Tool integration](https://github.com/hotosm/tasking-manager/issues/6094):
![image](https://github.com/hotosm/tasking-manager/assets/12103383/33ba73b9-d220-447e-ab70-ad5cac3ece80)

![image](https://github.com/hotosm/tasking-manager/assets/12103383/244f6d3e-4a76-4360-90cc-6bdf9b1b5fb8)

There will be a [new TM endpoint](https://github.com/hotosm/tasking-manager/issues/6144) to track active projects which will be used by a lambda function. Objective of this lambda function:
1. run everyday on 12 UTC cc @mahesh-naxa @eternaltyro 
2. hit the TM API to generate a list of published, active projects
3. using the list of active projects, send export jobs to [raw-data-api](https://github.com/hotosm/raw-data-api)

ramyaragupathy:
For now, we have the TM related lambda functions existing in the same repo: https://github.com/hotosm/tasking-manager/tree/develop/scripts/aws/lambda. If required, we should consider moving this to separate repos.

cc @dakotabenjamin @eternaltyro @mahesh-naxa @kaditya97

dakotabenjamin:
I say keep it in the repository. My major concern is cloud independence- we can default for now to using lambda but should also consider being able to run it as a tiny container/microservice should we need to.

ramyaragupathy:
Complete

PR #6189, opened 2024-01-08
Infra: tm-exporter lambda function using terragrunt.

### LINKED ISSUE #6146: Export tool integration: infrastructure tasks
For [export tool integration](https://github.com/hotosm/tasking-manager/issues/6094), @mahesh-naxa has identified the following devops related tasks.

- [x] Create cloudformation/terraform code for [lambda function](https://github.com/hotosm/tasking-manager/issues/6145) deployment & S3 Bucket Setup (use existing?).
- [ ] Update circleci to use needed orb & create test and deployment pipeline
- [x] Create Cloudwatch event to trigger [lambda function](https://github.com/hotosm/tasking-manager/issues/6145) (also add such to step 1)
- [x] Create a mechanism to monitor [lambda function](https://github.com/hotosm/tasking-manager/issues/6145) execution & their tracking 
- [x] Create a way to handle retries, could be aws sns or just plain [lambda](https://github.com/hotosm/tasking-manager/issues/6145) retries. Also test this.

cc @dakotabenjamin @eternaltyro @kaditya97 @varun2948 @royallsilwallz @manjitapandey

mahesh-naxa:
- Resolves #6146 #6145 

- Adds terraform snippet for creating lambda function, cloudwatch event, alarms & policies.
- Adds lambda layer for requests library (import from s3 or provide layer arn later. )

Note: 
- `TF_VAR_rawdata_api_auth_token`, Auth token for raw data api, Request [here](https://github.com/hotosm/raw-data-api/).
- `TF_VAR_raw_data_api`, API endpoint. Defaults to https://api-prod.raw-data.hotosm.org/v1
- `TF_VAR_active_projects_api_base_url`, Your [tasking-manager](https://github.com/hotosm/tasking-manager) instance. 
Defaults to https://tasking-manager-staging-api.hotosm.org/api/v2


mahesh-naxa:
hi @dakotabenjamin i have added added raw_data_api & active_projects_api_base_url as inputs on all sets of dev,stag,prod so, only `TF_VAR_rawdata_api_auth_token` remains to be set as an environment var.

scripts/aws/lambda/TM-Extractor/_envcommon/lambda.hcl
```diff
@@ -0,0 +1,46 @@
...
+# deploy mysql are defined here. This configuration will be merged into the environment configuration
+# via an include block.
+# ---------------------------------------------------------------------------------------------------------------------
+
+# Terragrunt will copy the Terraform configurations specified by the source parameter, along with any files in the
+# working directory, into a temporary folder, and execute your Terraform commands in that folder. If any environment
+# needs to deploy a different module version, it should redefine this block with a different ref to override the
+# deployed version.
+terraform {
+  # Sourcing from here rn, updating to a release from https://github.com/hotosm/TM-Extractor/ later.
```
dakotabenjamin, line 13: In order to make my tests work, I had to replace the source with my local build of TM-extractor where I set the required terraform version to `~>1.6.0`. See https://github.com/hotosm/TM-Extractor/issues/3 for more details.

scripts/aws/lambda/TM-Extractor/dev/env.hcl
```diff
@@ -0,0 +1,13 @@
+# Set common variables for the environment. This is automatically pulled in in the root terragrunt.hcl configuration to
+# feed forward to the child modules.
+
+locals {
```
dakotabenjamin, line 4: Should these be set with example defaults as a `env.hcl.example` with README instructions to replace and rename?

dakotabenjamin (changes requested):
I'm setting my review as "Request Changes" because everything looks good but I need some clarification. Once the questions are addressed here in in the TM-Extractor repo, I think this is good to merge.

scripts/aws/lambda/TM-Extractor/dev/env.hcl
mahesh-naxa: I think it would be up to us to have it already created. I have updated `README.md` within `scripts/aws/lambda/TM-Extrator/` with this information. 
@dakotabenjamin However i had no clue on what stag/prod input vars will be for `account_name`, `account_region`. Can you have that updated for stag/prod.

scripts/aws/lambda/TM-Extractor/_envcommon/lambda.hcl
mahesh-naxa: Yep, that would be nice to have minor patches allowed, only thing is that once we have our CircleCi config for deployments how would we make sure they align. Else i am all okay. 
I have updated [TM-Extractor](https://github.com/hotosm/TM-Extractor/tags)'s v.1.0.1 to be source. I see it has updated code on `=1.6.0`.

PR #6190, opened 2024-01-08
fix: Add "apidocs" target to produce API documentation

robsavoye:
This add a target to the Makefile to run Doxygen on the backend to produce API documentation. In addition, it adds the Doxyfile config file for Doxygen. As this doesn't effect anything else, it should be safe to merge.


robsavoye:
This will add API documentation for the TM backend.

PR #6191, opened 2024-01-08
Add Mkdocs support

robsavoye:
This adds initial support for using Mkdocs to publish documentation. When #6190 is merged, then the internal API documentation also gets published. Currently this is only a reworking of the existing documentation to support publication. I figure better to have a small PR, the doc updates will be future PRs.

For an example of how this looks, see the version in my TM fork:
https://robsavoye.github.io/tasking-manager/


robsavoye:
This will allow TM to publish all the documentation for TM to GH Pages.

PR #6192, opened 2024-01-08
Add doc on the database schema

robsavoye:
This docs is being moved from the tm-admin project, as it's TM specific. It includes content on all of the database tables, and their functionality. This is a work in progress, I'll be expanding this doc over time as I dig deeper into the database.


robsavoye:
Obviously adding a single doc won't break anything.

PR #6197, opened 2024-01-11
Start cleaning up docs-old

robsavoye:
Currently TM has a *docs-old* and a also a *docs* directories. This PR removes old versions after comparing them against the new versions for the common files like contributing or coding guidelines, used by other projects. Also the images directory has been renamed, and the diagrams from the techdocs git repository added as source and images.


robsavoye:
Rather than stack branches, this is needed before other work can proceed as it involves renaming and rearranging files.

PR #6206, opened 2024-01-18
fix: Fix relative links, add new docs to the menu

robsavoye:
Minor bug fixes in docs, and add new content.


robsavoye:
Changes nobody will notice unless they read the docs.

PR #6222, opened 2024-01-29
Feature/OSM Download meta data integration

varun2948:
## What does this PR do ? 
- Meta data shown to let user know it's updated date and size.

## Consideration 
-On previous approach , We didn't had any metadata, which was causing user to know last updated date and the actual size of the files.

## How to test ? 
-You can see two text one on side of the download with size of the file i.e. Geojson (2.4mb) and one below the filetype as last modified date as shown in the picture.

image


varun2948:
just fixed the frontend sonarlint issue @kshitijrajsharma, @kaditya97 can you see what is causing that issue ?

kshitijrajsharma (approved):

royallsilwallz (approved):

PR #5621, opened 2023-03-08
docker-compose deploy guide

jonnymccullagh:
Adding a guide to help other sysadmins who want to deploy a version of the Tasking Manager

PR #6234, opened 2024-02-02
Configurable devevelopment port & documentation updates

spwoodcock:
### Problem

- The docker compose setup for development uses Traefik on port 80.
- On many machines port 80 is not an option: either in use, or not available in rootless mode.
- The container names are also a mix of conventions: two being automatically generated, two being manually specified.
- The development setup guide is also missing a few things to get started easily.

### Solution

- Added a `TM_DEV_PORT` variable to the dotenv options, so the port can be configured.
- Updated all containers to have consistent names. I prepended `tm-` to prevent conflict with other tools that may have containers named `backend` and `frontend`.
- Updated the development setup guide:
  - Info on running via docker.
  - How to set up OSM OAuth creds.

### Context

- @robsavoye was having some issues setting up Tasking Manager for testing.
- I realised there isn't actually a guide for running the entire stack in Docker during development, which I assume some people would want to do.

----------------------------------------

**2024-02-02 15:35 Update/Edit**:
- I removed info about using pep582, as this was blocked by the python steering comity (😢)
- I noticed an unpinned pdm version during functional tests in CI. I pinned to the same as others in the repo. 
- Clarified info about running backend api only.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -81,8 +81,7 @@ jobs:
       - run: sudo apt-get update
       - run: sudo apt-get -y install libgeos-dev # Required for shapely
       - run: sudo apt-get -y install proj-bin libproj-dev
-      - run: pip install --upgrade pip pdm
-      - run: pdm config --global python.use_venv False
+      - run: pip install --upgrade pip pdm==2.7.4
       - run: pdm export --dev --without-hashes > requirements.txt
       - run: pip install -r requirements.txt
       - run: mkdir --mode 766 -p /tmp/logs
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,7 +15,7 @@ services:
   # Main application
   backend:
     <<: *backend
-    container_name: backend
+    container_name: tm-backend
     restart: always
     labels:
       - traefik.http.routers.backend.rule=(Host(`127.0.0.1`) || Host(`localhost`)) && PathPrefix(`/api/`)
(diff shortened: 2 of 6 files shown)


robsavoye:
There are two different docs on using docker with TM, but since docker wasn't working, I could update and merge the two docs.

spwoodcock:
Where is the second doc? I couldn't find it.

I found a guide for installing TM via docker for deployment, but no complete guide for running via docker during development.

robsavoye:
one is in docs-old, the other in in an old PR from Outreachy last year. I'm trying to get rid of docs-old by updating the docs, and moving them all to docs. Once I can get docker to work, I can merge and update a single doc.

robsavoye:
The old one I was planning to merge in is here: https://github.com/hotosm/tasking-manager/pull/5621. You're welcome to update a doc on docker usage for TM. :-)

spwoodcock:
Looks fine to me - that one is for deployment specifically though.

I know they are similar, but we need a doc for how to deploy, and a doc for how to develop using docker.

robsavoye:
For a doc on docker, I was thinking about a simple howto user guide. Basically an interested contributor should be able to clone the TM git repo, run docker-compose, and try TM in a self-hosted environment. I do this all the time when evaluating other projects, and if it doesn't work out of the box, that ends the evaluation most of the time.

spwoodcock:
To do that I think we should just update the `example.env` to use defaults for the docker based setup. 

Then change the port binded to traefik by default to any >1024.

After that a user could simply:

```bash
git clone https://github.com/hotosm/tasking-manager

cp example.env tasking-manager.env

docker compose up -d
```

robsavoye:
I like making it really easy for people to download and evaluate TM locally using docker.

spwoodcock:
I will make another PR updating as above & adding some info to the docs then 👍

PR #6247, opened 2024-02-15
Enhance : Active Projects API 

kshitijrajsharma:
## What does this PR do ? 

- This PR enhances the active projects API by : 
   - Fixing bug on UTC timestamp comparison while fetching projects 
   - if Project has a comment / conversation happening then that project is also considered as active 
  

## Consideration 

- I am aware of performance of the API , I will wait for the multi process approach that's going on before adding enhancement on this endpoint for the consistency 

## How to test ? 

- I have compared number of projects returned with following query in database directly and from the API endpoint : ```http://127.0.0.1:5000/api/v2/projects/queries/active/?interval=1```

```sql 
with t1 as (
select distinct  project_id 
from project_chat pc 
WHERE "time_stamp" > (CURRENT_TIMESTAMP AT TIME ZONE 'UTC') - INTERVAL '1 hours'
UNION
SELECT DISTINCT project_id 
FROM task_history 
WHERE action_date > (CURRENT_TIMESTAMP AT TIME ZONE 'UTC') - INTERVAL '1 hours'
)
select count(distinct  project_id)
from t1
```

### DIFF
--- a/backend/services/project_service.py
+++ b/backend/services/project_service.py
@@ -2,7 +2,7 @@ import threading
 from cachetools import TTLCache, cached
 from flask import current_app
 import geojson
-from datetime import datetime, timedelta, timezone
+from datetime import datetime, timedelta
 
 from backend.exceptions import NotFound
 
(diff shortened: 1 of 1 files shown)


ramyaragupathy (approved):
Thumbs up with the idea of including project Q&A section

PR #6256, opened 2024-02-25
Unlock tasks after validation using in parallel using python threads

### LINKED ISSUE #6154: [BUG] Unable to submit tasks after validation
**Describe the bug**

Several [users are reporting](https://hotosm.slack.com/archives/C319P09PB/p1700940729985989) delay in validation submission action. It is a huge blocker to the validation workflow. After hitting the, submit button there is a delay in submission and finally there is an error message indicated to the user.

image

image

image

image

image

image

### LINKED ISSUE #6244: Tasks validation using python threads
The validation API took a long time to validate and sometimes it fails as well.

Python threads module can be used to process multiple tasks at the same time and can reduce task validation time.

kaditya97:
- fix #6244 
- fix #6154 
- Uses Python multiprocessing pool to run validation tasks in parallel using available CPU cores
- Local session for thread implementation for validation-related tasks
- Test case-related validation modified to get task by taskId instead of index as threading may sometimes return task which do not follow order

### DIFF
--- a/backend/models/postgis/task.py
+++ b/backend/models/postgis/task.py
@@ -97,21 +97,29 @@ class TaskInvalidationHistory(db.Model):
         db.session.commit()
 
     @staticmethod
-    def get_open_for_task(project_id, task_id):
+    def get_open_for_task(project_id, task_id, local_session=None):
+        if local_session:
+            return local_session.query(TaskInvalidationHistory).filter_by(
+                task_id=task_id, project_id=project_id, is_closed=False
+            ).one_or_none()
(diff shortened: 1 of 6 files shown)


robsavoye:
Using SQLAlchemy sessions should fix the problem with thread contention when accessing the database. I can't run the backend yet, do the tests now pass ?

kaditya97:
> Using SQLAlchemy sessions should fix the problem with thread contention when accessing the database. I can't run the backend yet, do the tests now pass ?

I used new SQLAlchemy session in each thread and passed the session as local session and it worked. Also all test are passing now.

PR #6266, opened 2024-03-01
Deployment fixes and infrastructure maintenance

dakotabenjamin:
The major reason why the deployment failed last week was networking issues connecting via ipv4 to s3 resources for cloudwatch utilities and deployment signaling software. We changed over to using the dual-stack links for downloading those packages. I also took the liberty of fixing some long-standing problems with the infrastructure and deployment:
- Replaced the LaunchConfiguration with a [LaunchTemplate](https://docs.aws.amazon.com/autoscaling/ec2/userguide/launch-templates.html) 
- Made the frontend s3 bucket private and only accessible through cloudfront
- created a development environment in the CI config and removed some redundant builds to deployment branches

Huge thanks to @eternaltyro for helping with the NAT64 config on the VPC side as well. A lot of AWS resources still don't support dual-stack connectivity despite amazon's recent push away from ipv4. 

Some more improvements could be made but I don't want to hold up the deployment any longer:
- We should set proper response headers with a working content security policy
- Instead of using `deployment/*` branches, we should deploy to production directly from releases. 

For devs- now you can test new features on the dev server. All you need to do is replace one line in the [.circleci/config.yml](https://github.com/hotosm/tasking-manager/blob/fix/febdeployment/.circleci/config.yml#L455) file:
```
    - equal: [ {YOUR-BRANCH}, << pipeline.git.branch >> ] # change this to the branch you wish to test
```
just change {YOUR-BRANCH} to the branch you wish to test thedeployment on. It must be a branch within the hotosm tasking manager repository. Just note that whenever the `develop` branch is updated, those changes will be overridden. we may need to add some changes to make it clear what branch or version is the current one on the dev server? maybe under `release` in the [heartbeat api response](https://tasking-manager-dev-api.hotosm.org/api/v2/system/heartbeat/). 

Once merged, tasks-stage.hotosm.org will deploy off the `main` branch, which needs to be created. This will line up more closely with FMTM's deployment schema.


dakotabenjamin:
The deployment "failed" on CircleCI but it was actually successful. It looks like CircleCI will fail out after 10 minutes of no input, but the instance deployment took 11 minutes: 
image

After review, it would be fine to push despite the error. I will see if we can extend the timeout. @eternaltyro it is odd that now its taking longer to instantiate EC2s than before, i am not sure what is happening. Use to be around 6-8 minutes. Maybe not worth investigating though, we have other priorities for containerisation that will make it irrelevant.

mahesh-naxa:
I am fine with the branching strategy. Requesting @eternaltyro for the cloudformation checks.
cc: @dakotabenjamin

scripts/aws/cloudformation/tasking-manager.template.js
```diff
@@ -359,80 +362,155 @@ const Resources = {
...
+          'wget -6 https://s3.dualstack.us-east-1.amazonaws.com/cloudformation-examples/aws-cfn-bootstrap-py3-latest.tar.gz -O /tmp/aws-cfn-bootstrap-py3-latest.tar.gz',
+          'pip install /tmp/aws-cfn-bootstrap-py3-latest.tar.gz',
+          'pdm install',
+          'eval "$(pdm venv activate)"',
+          'echo fs.inotify.max_user_watches=524288 | sudo tee -a /etc/sysctl.conf',
+          'export LC_ALL=C',
+          'wget -6 https://s3.dualstack.us-east-1.amazonaws.com/amazoncloudwatch-agent/ubuntu/amd64/latest/amazon-cloudwatch-agent.deb -O /tmp/amazon-cloudwatch-agent.deb',
+          'dpkg -i /tmp/amazon-cloudwatch-agent.deb',
+          'echo "Exporting environment variables:"',
+          cf.sub('export NEW_RELIC_LICENSE=${NewRelicLicense}'),
```
eternaltyro, line 446: Invalid. Remove in favour on ENVVAR further along in line 478

scripts/aws/cloudformation/tasking-manager.template.js
```diff
@@ -359,80 +362,155 @@ const Resources = {
...
+          cf.join('', ['export POSTGRES_ENDPOINT=', cf.getAtt('TaskingManagerRDS','Endpoint.Address')]),
+          cf.sub('export POSTGRES_DB=${PostgresDB}'),
+          cf.sub('export POSTGRES_PASSWORD="${PostgresPassword}"'),
+          cf.sub('export POSTGRES_USER="${PostgresUser}"'),
+          cf.sub('export TM_APP_BASE_URL="${TaskingManagerAppBaseUrl}"'),
+          cf.sub('export TM_ENVIRONMENT="${AWS::StackName}"'),
+          cf.sub('export TM_CLIENT_ID="${TaskingManagerOAuthClientID}"'),
+          cf.sub('export TM_CLIENT_SECRET="${TaskingManagerOAuthClientSecret}"'),
+          cf.sub('export TM_REDIRECT_URI="${TaskingManagerAppBaseUrl}/authorized"'),
+          'export TM_SCOPE="read_prefs write_api"',
```
eternaltyro, line 456: Can be removed since the same value is set as default in the backend config file.

scripts/aws/cloudformation/tasking-manager.template.js
```diff
@@ -359,80 +362,155 @@ const Resources = {
...
+          cf.sub('export TM_SECRET="${TaskingManagerSecret}"'),
+          cf.sub('export TM_SMTP_HOST="${TaskingManagerSMTPHost}"'),
+          cf.sub('export TM_SMTP_PASSWORD="${TaskingManagerSMTPPassword}"'),
+          cf.sub('export TM_SMTP_PORT="${TaskingManagerSMTPPort}"'),
```
eternaltyro, line 460: Can be removed - as SMTP port is by default 587 as it is for AWS SES.

scripts/aws/cloudformation/tasking-manager.template.js
```diff
@@ -359,80 +362,155 @@ const Resources = {
...
+          cf.sub('export TM_SMTP_USER="${TaskingManagerSMTPUser}"'),
+          cf.sub('export TM_SMTP_USE_SSL="${TaskingManagerSMTPSSL}"'),
+          cf.sub('export TM_SMTP_USE_TLS="${TaskingManagerSMTPTLS}"'),
```
eternaltyro, line 463: Can be removed: SMTP TLS is enabled by default and SSL is disabled by default in the backend config as is the case for AWS SES.

scripts/aws/cloudformation/tasking-manager.template.js
```diff
@@ -359,80 +362,155 @@ const Resources = {
...
+          cf.sub('export TM_DEFAULT_CHANGESET_COMMENT="${TaskingManagerDefaultChangesetComment}"'),
+          cf.sub('export TM_EMAIL_FROM_ADDRESS="${TaskingManagerEmailFromAddress}"'),
```
eternaltyro, line 465: Can be removed: default/fall-back email address in backend config is same as the one set in CFN.

scripts/aws/cloudformation/tasking-manager.template.js
```diff
@@ -359,80 +362,155 @@ const Resources = {
...
+          cf.sub('export TM_EMAIL_CONTACT_ADDRESS="${TaskingManagerEmailContactAddress}"'),
```
eternaltyro, line 466: value in CFN set to an email address we don't monitor. So maybe we should let this fallback to defaults. Good to remove.

scripts/aws/cloudformation/tasking-manager.template.js
```diff
@@ -359,80 +362,155 @@ const Resources = {
...
+          cf.sub('export TM_IMAGE_UPLOAD_API_KEY="${TaskingManagerImageUploadAPIKey}"'),
+          'psql "host=$POSTGRES_ENDPOINT dbname=$POSTGRES_DB user=$POSTGRES_USER password=$POSTGRES_PASSWORD" -c "CREATE EXTENSION IF NOT EXISTS postgis"',
+          cf.if('DatabaseDumpFileGiven', cf.sub('aws s3 cp ${DatabaseDump} dump.sql; sudo -u postgres psql "postgresql://$POSTGRES_USER:$POSTGRES_PASSWORD@$POSTGRES_ENDPOINT/$POSTGRES_DB" < dump.sql'), ''),
+          'pdm run -vv flask db upgrade',
+          'echo "------------------------------------------------------------"',
+          cf.sub('export NEW_RELIC_LICENSE_KEY="${NewRelicLicense}"'),
+          cf.sub('export TM_SENTRY_BACKEND_DSN="${SentryBackendDSN}"'),
+          'export NEW_RELIC_ENVIRONMENT=$TM_ENVIRONMENT',
+          cf.sub('NEW_RELIC_CONFIG_FILE=./scripts/aws/cloudformation/newrelic.ini newrelic-admin run-program gunicorn -b 0.0.0.0:8000 --worker-class gevent --workers 5 --timeout 179 --access-logfile ${TaskingManagerLogDirectory}/gunicorn-access.log --access-logformat \'%(h)s %(l)s %(u)s %(t)s \"%(r)s\" %(s)s %(b)s %(T)s \"%(f)s\" \"%(a)s\"\' manage:application &'),
```
eternaltyro, line 481: Okay for now, but I think a systemd unit file is in order.

scripts/aws/cloudformation/tasking-manager.template.js
```diff
@@ -713,30 +792,67 @@ const Resources = {
...
               cf.join("/", [
-                  cf.getAtt("TaskingManagerReactBucket", "Arn"),
+                cf.getAtt("TaskingManagerReactBucket", "Arn"),
                   "*"
               ])
             ],
-            Sid: "PublicGetObject"
+            "Condition": {
+              "StringEquals": {
+                "AWS:SourceArn": cf.sub("arn:aws:cloudfront::${AWS::AccountId}:distribution/${TaskingManagerReactCloudfront}")
```
eternaltyro, line 809: <3

scripts/aws/cloudformation/tasking-manager.template.js
```diff
@@ -713,30 +792,67 @@ const Resources = {
...
           {
-            Action: [ "s3:ListBucket" ],
-            Effect: "Allow",
-            Principal: "*",
-            Resource: [
-              cf.getAtt("TaskingManagerReactBucket", "Arn")
+            "Sid": "s3allowbucketobjectaccess",
+            "Effect": "Allow",
+            "Principal": {
+                "AWS": cf.sub("arn:aws:iam::${AWS::AccountId}:role/CircleCI-OIDC-Connect")
```
eternaltyro, line 817: <3

eternaltyro (changes requested):
Just some deletions to reduce number of lines.

scripts/aws/cloudformation/tasking-manager.template.js
dakotabenjamin: what if it needs override in a different environment?

PR #6281, opened 2024-03-12
Pull "main" branch to develop for new branching strategy

dakotabenjamin:
@ramyaragupathy this is an example pull request that could be made to prepare a deployment. 

First all PRs should be made to "develop" which will be deployed to the dev servers

Then when preparing for a deployment, a PR should be made like this one, develop onto main. Merging this will deploy to staging and the PR can serve as a useful commit history for the release notes.


dakotabenjamin:
hm, there appear to be rebase conflicts? We may need do delete and recreate `main` or force push. @ramyaragupathy what do you think?

robsavoye:
Since there are many old PRs for the main branch, which nobody has used for quite a while,
I think recreating it from the current develop branch is the way to go. We'd then keep this stable main branch which would be used for deployments. All new PRs would go into the develop branch and tested on the dev server.

ramyaragupathy:
Agree with @robsavoye - we will create a new [main](https://github.com/hotosm/tasking-manager/tree/main) branch out of existing [develop](https://github.com/hotosm/tasking-manager/tree/develop).

PR #6289, opened 2024-03-18
Fix/ faster load project list with no map omit

varun2948:
## What does this PR do ? 

- Enhances Project Lists Page Project List API and overall execution time, when Map is not turned on. 

## Consideration 

- On previous approach , We were using project list api without omitMapResults true and even if the map was not open, It was still fetching all the geojson of projects.
- Current approach : We are using omitMapResults params so that when we don't need geojson we can only fetch the project related data and only fetch geojson when the map is turned on.

## How to test ? 

- Try using the projectlist page it's response time has decreased and projects card loads faster.

## Screenshot
image

### DIFF
--- a/frontend/src/components/projects/projectNav.js
+++ b/frontend/src/components/projects/projectNav.js
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useEffect } from 'react';
 import { Link, useLocation } from 'react-router-dom';
 import { FormattedMessage } from 'react-intl';
 import { useSelector, useDispatch } from 'react-redux';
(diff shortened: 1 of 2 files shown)


varun2948:
@ramyaragupathy we are having issue on github action any idea, what could it be ?

ramyaragupathy:
Following up w/ @eternaltyro @dakotabenjamin

dakotabenjamin:
we can safely ignore the failure. GitHub actions won't allow an image to be pushed from a forked repository.

royallsilwallz (approved):

PR #6326, opened 2024-04-10
Create a version of TM with sandboxed OSM for training data generation

spwoodcock:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Describe this PR

- I created a demo local setup for using a bundled version of OSM that is sandboxed.
  - We need something like this as part of a TM specifically for generating training data for fAIr.
- It adds a database and OpenStreetMapwebsite/API service to the stack.
- The version of OSM includes a pre-generated user 'admin@hotosm.org' with password 'Password1234'.
- There is also a pre-generated OAuth2 application, with the redirect URI set to that in the tasking-manager.env `REDIRECT_URI` variable.

Things still to do:
- Instead of creating an OAuth2 App setup and requiring login, we could possibly just generate an access token in the database, then use that as the `access_token` to hardcoded for iDEditor (in `editor.js`?).

- Instead of pushing to a sandboxed OSM, we could delete all existing geoms in live OSM (not ideal).
- There is also OSM dev servers we would request access to? https://wiki.openstreetmap.org/wiki/Using_the_dev_server

## Testing / Setup

1. First build the OSM container:

```bash
cd contrib/tm-training
docker compose build osm
```

2. Set the REDIRECT_URL correctly in tasking-manager.env:

3. Run all services together:

```bash
dc -f docker-compose.yml -f contrib/tm-training/docker-compose.yml up -d
```

4. Get the generated Client ID and Client Secret:

```bash
dc -f docker-compose.yml -f contrib/tm-training/docker-compose.yml logs osm
```

5. Update the env vars in tasking-manager.env.

6. Restart all services:

```bash
dc -f docker-compose.yml -f contrib/tm-training/docker-compose.yml up -d
```

## Extra resources

- OSM docker install: https://github.com/openstreetmap/openstreetmap-website/blob/master/DOCKER.md#initial-setup
- OSM OAuthToken model? https://github.com/openstreetmap/openstreetmap-website/blob/5dc9fb3dc550a7b81ee6df5d6ca97e050798aba9/app/models/oauth_token.rb#L38
- OSM OAuth tests: https://github.com/openstreetmap/openstreetmap-website/blob/5dc9fb3dc550a7b81ee6df5d6ca97e050798aba9/test/integration/oauth2_test.rb#L151 (for inspiration)


ramyaragupathy:
Would this fit the need for end to end mapping testing grounds @spwoodcock @dakotabenjamin ?

contrib/tm-training/rsa.private
dakotabenjamin, line 1: What is this file?

spwoodcock: It's a file required by the smtp service (https://github.com/ix-ai/smtp) - it's empty

spwoodcock:
I migrated most of the code here to https://github.com/hotosm/osm-sandbox/

The TM specific config can perhaps we made in another PR after discussion on the best approach.
As @dakotabenjamin says, the easiest way is probably just to specify which OSM backend to use (production OSM, or an instance on another domain, such as sandbox.hotosm.org).

spwoodcock:
Note for future ref:

- The sign up via TM doesn't work here if both 1) the instance is connected to sandbox osm 2) the username already exists in official osm but not the sandbox

Requires debugging and/or better docs how to solve that issue

PR #6334, opened 2024-04-11
Add ability to delete a user's information

### LINKED ISSUE #6002: Add function to delete user accounts
Related to https://github.com/hotosm/tasking-manager/issues/2526, we need to support user account deletion. When a user requests to have their account deleted, we do not have an effective way to do so short of direct database access. 

Users with admin-level access can have a select option under https://tasks.hotosm.org/manage/users/ to delete an account. There should be a dialogue to confirm deletion. 

image

On the backend, the solution should be designed to remove the users personal information without deleting core TM data/functionality like task history. Replace user ids with `[Deleted user]` in comments and such. We will need to decide how to effectively remove account information to comply with GDPR. 

**Additional context**
We have received requests for account deletion in the past, and have not come to a sufficient solution to comply with these requests.

tsmock:
- [x] 🍕 Feature

## Related Issue
Fixes #6002 (backend; frontend UI not done)

## Describe this PR

This adds an API call (specifically, a `DELETE` to `/users/<int:user_id>` (specific user) and `/users` (users no longer on OSM)).

As much user-supplied information as possible is redacted, with the following exceptions:
* Projects (specifically of note, project descriptions are not removed)
  * Projects are not reassigned either (maybe this should happen?)
* Tasks (I believe this is would only have task modification times)
* Interests (specifically, interests that the user created)
* Licenses (just in case a license was required by a project)
* User role (admins are downgraded to mappers, blocked users remain blocked)

## Screenshots

N/A

Deleting user from database completely. Rejected partially due to complexity of figuring out what to do with mapped/validated tasks and user blocks.

## Review Guide

For the code path that redacts users from the DB if the user deleted their account in OSM, user [535043](https://api.openstreetmap.org/api/0.6/user/535043) is known to no longer exist.

Sample psql command for user 535043:
```sql
INSERT INTO users VALUES (535043, 'some username', 0, 1, 0, 0, 0, null, 'foobar@example.com', '@meta', true, '@linkedin', '@twitter', NOW(), null, true, false, '@irc', 'London', 'United Kingdom', 'John Doe', '@skype', '@slack', 'ID', true, true, 'https://image.example/png.png', 0, 'a gender here', true, true, true);
```
```bash
$ curl -X DELETE -H 'Authorization: Token ${TOKEN}' localhost:5000/api/v2/users/
```

Further questions:
Is the OSM server admin team OK with us hitting `/user/:id` frequently?

EDIT: TomH pointed me at https://planet.openstreetmap.org/users_deleted/users_deleted.txt .


ramyaragupathy:
Implemented through https://github.com/hotosm/tasking-manager/pull/7070

PR #6431, opened 2024-05-14
v4.7.3 Release PR

dakotabenjamin:
## Describe this PR
We were originally going to make a hotfix only, but the other features are tested and fine to deploy, so we are just going to do a full release. 

@ramyaragupathy to follow with release notes.


dakotabenjamin:
@ramyaragupathy let me merge because i need to run a cloudformation script prior to deployment

Issue #6483, opened 2024-07-02
Partners: Image field character length should be extended on DB 


manjitapandey:
**Describe the bug**
I am unable to upload image link on partners page. Shows 500 error on network panel

**To Reproduce**
Make sure you have admin access to the platform
1. Go to 'Manage'
2. Click on 'partners'
3. Click on 'new
4. upload image link
5. See error

**Expected behavior**
The image should be uploaded and should be visible on partners statistics page

**Additional context**
The word "image" doesn't provide proper meaning since users will be uploading link of their logo. The terms "Partner's logo" would fit more. 
cc: @ramyaragupathy , @royallsilwallz , @prabinoid

PR #6487, opened 2024-07-04
Partners table atrributes lengths adjusted

### LINKED ISSUE #6464: Partners: secondary hashtag field size is too small in the DB
**Describe the bug**
The limit for the secondary hashtag column is 50 characters, which is too small.

**To Reproduce**

1. Go to create a partner
2. Try with several hashtags (+50 characters) in the secondary hashtag field
3. Click "Save"
4. It will fail because the API returns Error 500

**Expected behavior**

To accept more than 50 characters for the secondary hashtag.

**Additional info**

Check this migration file:

https://github.com/hotosm/tasking-manager/blob/develop/migrations/versions/e8ffa33a9c18_.py#L26C1-L26C23

prabinoid:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [x] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Fixes #6464 and #6483

## Describe this PR

Changed the max length of the attributes in the Partners table.

Wanted to remove the length contraints on urls as it can be longer than anticipated but had to downgrade the alembic version in the develop too as Alembic sometimes does not automatically detect all changes in the model, especially when the changes involve removing length constraints or modifying columns with minimal schema impact. Nevertheless, defining the length in attributes is generally a good practice. So, increased the length and made migrations.


prabinoid:
Rebased develop and handled flake8 and black formatting tests.

PR #6488, opened 2024-07-04
Add bind mounts to docker compose dev setup + migration command override

spwoodcock:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [x] 🤖 Build or CI

## Related Issue

https://github.com/hotosm/tasking-manager/pull/6487

## Describe this PR

- Bind mounts
  - Currently there are no bind-mounts to local files inside dev containers.
  - Meaning we have to build every time to run tests or migrations.
  - Mounting the files makes it easier for: hot reload, running tests, running migrations.
- I also removed the `version` param from docker-compose.yml files, as it's now obsolete.
- Moved the `upgrade` command out of the tm-migration entrypoint. Allow for override via `command: xxx` in docker-compose.

## Review Guide

Notes for the reviewer. How to test this change?

### DIFF
--- a/docker-compose.override.sample.yml
+++ b/docker-compose.override.sample.yml
@@ -1,5 +1,3 @@
-version: "3"
-
 name: tasking-manager-main
 
 services:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,5 +1,3 @@
-version: "3"
-
 name: tasking-manager-main
 
 volumes:
@@ -38,6 +36,10 @@ services:
         condition: service_completed_successfully
     env_file:
       - tasking-manager.env
+    volumes:
+      - ./pyproject.toml:/usr/src/app/pyproject.toml:ro
+      - ./backend:/usr/src/app/backend:ro
+      - ./tests:/usr/src/app/tests:ro
     restart: unless-stopped
     healthcheck:
       test: curl --fail http://localhost:5000 || exit 1
@@ -64,14 +66,18 @@ services:
     image: ghcr.io/hotosm/tasking-manager/backend:main
     build:
       context: .
-    entrypoint: ["python", "manage.py", "db", "upgrade"]
+    entrypoint: ["python", "manage.py", "db"]
+    command: "upgrade"
     depends_on:
       tm-db:
         condition: service_healthy
     env_file:
       - tasking-manager.env
+    volumes:
+      - ./pyproject.toml:/usr/src/app/pyproject.toml:ro
+      - ./backend:/usr/src/app/backend:ro
+      - ./migrations:/usr/src/app/migrations:ro
     deploy:
-      replicas: ${API_REPLICAS:-1}
       resources:
         limits:
           cpus: "1"
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -17,7 +17,7 @@ services:
       - tasking-manager.env
     restart: unless-stopped
     healthcheck:
-      test: pg_isready -U ${POSTGRES_USER} -d ${POSTGRES_DB}
+      test: pg_isready -U ${POSTGRES_USER:-tm} -d ${POSTGRES_DB:-tasking-manager}
       start_period: 35s
       interval: 10s
       timeout: 5s


mahesh-naxa (approved):

mahesh-naxa:
> I just rebased to `develop`.
> 
> Also added default values for `POSTGRES_USER` and `POSTGRES_DB` for the healthcheck in docker-compose.yml.
> 
> One of the annoying things about using a .env file named `tasking-manager.env` instead of `.env` is that we either need to specify with `docker-compose --env-file tasking-manager.env ...` or specify default for a standard `docker-compose ...`.
> 
> Values in `.env` are automatically injected into docker-compose.yml, while vars in custom named .env files are not.

Yeah @spwoodcock i recently had to dig into this, As, Bash from techwd, had issues trying to set up tasking-manager on his server.
He got this.
> -----------
> Despite setting the variables in the tasking-manager.env file. We are getting the following on running 'docker compose up --detach':
> 
>     WARN[0000] The "POSTGRES_USER" variable is not set. Defaulting to a blank string.                                  
>     WARN[0000] The "POSTGRES_DB" variable is not set. Defaulting to a blank string.
> 
> -----------

The warning was due to the vars not setting up in compose file. Although not an issue as `pg_isready -U "anything" -d "anything"` would give the exact exit code, even if its an empty string. Basically it would just check if postgres server is up or not.
For those setting up `tasking-manager` for the first time, this caused migration to fail as the database wasn't initialized on time.
We have `start_period` in health-checks , I thought it acted like a `grace-period`(that doesn't perform health check for this period of time), but this time flag is just for discarding health-checks results before that time. But Our's one succeeds anyway, so it will pass at the first second it gets.

This caused migration container to fail, resulting in `backend` service to be `Created` but not `Started`.

I think we should replace this with something that does actual query, i have seen examples like this one: https://github.com/docker-library/healthcheck/blob/master/postgres/docker-healthcheck

Your thoughts?

spwoodcock:
You can use `depends_on` in docker compose for this purpose 😄

With various conditions like `service_healthy`.
Check the FMTM docker compose files for examples 👍

Adding of the default vars in this PR will also fix the error the user above was having

prabinoid:
Related Issue #6498

mahesh-naxa:
> You can use `depends_on` in docker compose for this purpose 😄
> 
> With various conditions like `service_healthy`. Check the FMTM docker compose files for examples 👍
> 
> Adding of the default vars in this PR will also fix the error the user above was having

Agreed `depends_on` is the way it should avoid these situation, and we have it in `tasking-manager` docker setup. But, here `db` service is presumed `healthy` due to the health checks that we have, its similar to fmtm right now. [here](https://github.com/hotosm/fmtm/blob/f603be6b5fb37726da5676acd8cd5857ea6ec32b/docker-compose.yml#L250). 
But when user sets up the project for the 1st time, The process flows like this:
1. `db` service is started.
2. `migration` container is created but not started, since it `depends_on` `db` container to be healthy. 
3. `db` container is healthy, as `pg_isready -U ${FMTM_DB_USER:-fmtm} -d ${FMTM_DB_NAME:-fmtm}` exits 0. Even if the database is not there, or user isn't created yet & so on.
4. `migration` container now starts since `db` is presumed healthy (meanwhile db is now creating extensions, creating database, granting access to user etc).
5. `migration` container exits 1, as `db` is yet to configure itself with the user/pass/db, so migration fails.
6. `backend` container never starts as `migration` exited non-zero.

If the user performs `docker compose up -d` again, the containers would rerun thus now fixing this issue (since `db`'s initialization presumably is now complete),

spwoodcock:
Thanks for clarifying! I understand the issue now. 

I'm not sure why the db and user would not exist yet, before the first healtcheck runs.

But if this is the case, and tweaking the `interval` also doesn't help, then like you say a custom entrypoint is possible.

The backend container could have a migrate-entrypoint.sh, which includes a check on the db first before running the migration code. 

FMTM has something similar with bundled entrypoints in the backend container than can be selected based on the context

dakotabenjamin:
@spwoodcock can you rebase this? The backend tests should run properly after.

spwoodcock:
Done 😄 Let's see!

Daryus2023:
@image

PR #6514, opened 2024-07-30
Fix ssm permissions to pull the AMI

dakotabenjamin:
## What type of PR is this? (check all applicable)

- [ ] 🍕 Feature
- [x ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [x ] 🤖 Build or CI
- [ ] ❓ Other (please specify)

## Related Issue

Example: Fixes #6512 regression

## Describe this PR

I had tested the PR using a role with admin level permissions, so I didn't catch the lack of permissions needed to GetParameters:

```
19:41:35Z *********: UPDATE_FAILED TaskingManagerLaunchTemplate: Resource handler returned message: "User: arn:aws:sts::************:assumed-role/*********************/backend-deploy is not authorized to perform: ssm:GetParameters on resource: arn:aws:ssm:*********:************:parameter/ami-01e5ff16fd6e8c542 because no identity-based policy allows the ssm:GetParameters action ...
```

## Screenshots

Please provide screenshots of the change.

## Alternative Approaches Considered

Did you attempt any other approaches that are not documented in code?

## Review Guide

Notes for the reviewer. How to test this change?

## Checklist before requesting a review

- 📖 Read the Tasking Manager Contributing Guide: <https://github.com/hotosm/tasking-manager/blob/develop/docs/developers/contributing.md>
- 📖 Read the HOT Code of Conduct: <https://docs.hotosm.org/code-of-conduct>
- 👷‍♀️ Create small PRs. In most cases, this will be possible.
- ✅ Provide tests for your changes.
- 📝 Use descriptive commit messages.
- 📗 Update any related documentation and include any relevant screenshots.
- 🔠 Does this PR introduce or change any environment variables? If so, make sure to specify this change in the description.

PR #6515, opened 2024-07-30
Revert "Fix ssm permissions to pull the AMI"

dakotabenjamin:
Reverts hotosm/tasking-manager#6514

I misread the error- It was actually a CircleCI OIDC permissions issue rather than one with the LauchTemplate itself. @tsmock thats why you didn't experience it!


tsmock:
No worries; I just spent a lot of time updating the tasks.mapwith.ai TM instance from v4.5.0 to v4.8.0, and that still isn't done. Most of that time was troubleshooting configuration issues. And fiddling with terraform configurations.

PR #6517, opened 2024-07-31
Implement query execution using encode/databases for organisations li…

prabinoid:
…st, updated configs/settings and user dependency function

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [x] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Describe this PR
Installed and implemented [databases](https://www.encode.io/databases/database_queries/) and core expressions for database queries.
Updated configs.
Pydantic field validators for db url validation.
User dependency function.
Organisation list api with core expressions.

### DIFF
--- a/backend/db.py
+++ b/backend/db.py
@@ -10,55 +10,57 @@ from sqlalchemy.ext.asyncio import (
 )
 from sqlalchemy.orm import declarative_base
 
+from databases import Database
+from sqlalchemy import create_engine
+from sqlalchemy.orm import declarative_base, sessionmaker
+
 Base = declarative_base()
+class DatabaseConnection:
+    """Manages database connection (sqlalchemy & encode databases)"""
 
-# Heavily inspired by https://praciano.com.br/fastapi-and-async-sqlalchemy-20-with-pytest-done-right.html
-# https://medium.com/@tclaitken/setting-up-a-fastapi-app-with-async-sqlalchemy-2-0-pydantic-v2-e6c540be4308
+    def __init__(self):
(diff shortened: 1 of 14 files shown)


backend/db.py
```diff
@@ -10,55 +10,57 @@
...
-    @contextlib.asynccontextmanager
-    async def connect(self) -> AsyncIterator[AsyncConnection]:
-        if self._engine is None:
-            raise Exception("DatabaseSessionManager is not initialized")
+    def __init__(self):
+        self.database = Database(
+            settings.SQLALCHEMY_DATABASE_URI.unicode_string(), min_size=5, max_size=20
+        )
+        # self.database = Database(settings.DTM_DB_URL.unicode_string())
+        self.engine = create_engine(
```
spwoodcock, line 26: Do we need to create two database connections for some reason? One with encode/databases, the other with SQLAlchemy?

This was also done for DroneTM, but I removed it. The only reason I see for the SQLAlchemy engine is if we need to use alembic too, but this shouldn't be part of the endpoint dependencies and should probably be done elsewhere

spwoodcock (approved):
Nice use of SQL 👍

backend/db.py
prabinoid: No Sam, Previously our setup was different and I used get_session() in the migrated endpoints. Now we are using databases so once it is implemented in all those migrated endpoints, I will remove it as removing now will throw error so.

Issue #5780, opened 2023-05-10
iD editor version upgrade 


ramyaragupathy:
We are still using a somewhat outdated version of iD ([v2.21.1](https://github.com/hotosm/tasking-manager/blob/develop/frontend/package.json#L11) – the most recent version of iD is v2.25.2) in the tasking manager. 

cc @eternaltyro @emi420 @Aadesh-Baral @HelNershingThapa

ramyaragupathy:
This should also cover the maxar key update - which was applied as a hotfix through a temporary solution.

cyrossignol:
Hello, I updated iD to version 2.26.2 for our use at [tasks.opensidewalks.com](https://tasks.opensidewalks.com/) here: OpenSidewalks@786c5108b03713463ed77194be13638ef03fd4a8

I reworked the integration by avoiding the direct bundling of iD code with the tasking manager JS. This gets around an issue with the output of iD's newer esbuild bundling system and significantly improves the tasking manager build time. I'm inclined to apply the same treatment for RapiD to remove the need for the special upstream releases.

Let me know if this is worth a PR or if HOT is pursuing a different upgrade strategy.

ramyaragupathy:
@cyrossignol - thanks for checking on this. Your idea sounds good, could you please go ahead making a PR for our team to go through this in detail?

cc @HelNershingThapa @tsmock

tsmock:
I'll have to check and see how well it works with Rapid. Of specific note, I don't know if waiting for Rapid to finish "loading" (via `script.onload`) will work -- Rapid does a lot of setup async now (current `main`, should be released sometime soon).

I haven't touched the iD code, but iD also doesn't have as much backend changes going on.

tsmock:
I've been playing around with the `script` loading stuff since Rapid v2.1 was released.

Some comments on how it improved the build experience and could (potentially) make it much easier to update out-of-band:
* We can use a CDN (e.g. `https://cdn.jsdelivr.net/npm/@rapideditor/rapid@~2.1.0/dist/rapid.js`)
 * Pro: Reuse of rapid download between TM instances
 * Con: Dependent upon free CDN
* It _significantly_ reduces the build time for me -- as in 18 minutes to 3 minutes for me.

If anyone wants to play with that (for Rapid), I've got a branch here: https://github.com/facebook/OSM-HOT-Tasking-Manager/tree/chore/update-rapid

Notes:
[...]

PR #6551, opened 2024-09-03
iD editor upgrade to v2.30.2

emi420:
## What type of PR is this?

- [X] ❓ Other (dependency upgrade)

## Related Issue

Fixes https://github.com/hotosm/tasking-manager/issues/5780

## Describe this PR

A new version (2.30.2) of the iD Editor was published as a NPM package on [hotosm/id](https://www.npmjs.com/package/@hotosm/id)

This PR upgrades the package and also adds a configuration line for Webpack to ignore and not parse the `iD.min.js` file, as it was raising some warnings:

```sh
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted
Critical dependency: the request of a dependency is an expression

Search for the keywords to learn more about each warning. 
To ignore, add // eslint-disable-next-line to the line before.

WARNING in ./node_modules/@hotosm/id/dist/iD.min.js 14:40-47
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted

WARNING in ./node_modules/@hotosm/id/dist/iD.min.js 17:37-44
Critical dependency: require function is used in a way in which dependencies cannot be statically extracted

WARNING in ./node_modules/@hotosm/id/dist/iD.min.js 40022:67-76
Critical dependency: the request of a dependency is an expression
```

It should be ok as this is a packed and minified file.

### Copy static files and imagery patch

I've also divided the commands in package.json for copying static files for both iD and Rapid and removed the patch for the iD imagery as isn't being used anymore.


spwoodcock (approved):

royallsilwallz:
hey @emi420 , we have another issue related to iD upgrade. When I toggle the editor to RapiD and again back to iD, an error occurs. Below is the screenshot - 

image

**Other findings**
I have also tried it on different branches (`develop`). It still shows the same error.
So, it is not specific to this (`upgrade/iD-editor-2.30.2`) branch.
Older version (`v2.29.0`) did not have this error.

**Steps to reproduce**
- Go to any project
- Map a task
- Change the editor to RapiD
- Again change it back to iD
- See the error

cc - @ramyaragupathy @manjitapandey

PR #6570, opened 2024-09-23
v4.8.1 Latest changes to staging

ramyaragupathy:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [x] 🤖 Build or CI


dakotabenjamin:
There is a docker build issue: 
```
 ERROR: tag is needed when pushing to registry
Error: buildx failed with: ERROR: tag is needed when pushing to registry
```

and then some critical CVEs in the docker scan, which we can choose to ignore until we get the containerization finished (but need to fix before then!!)

There are some changes in the deployment with the new AMI config from Taylor, so we should coordinate staging deployment so I can make sure those changes go through successfully.

PR #6577, opened 2024-09-26
Fix: Speed up CSV generation by minimizing DB probes

bshankar:

- [x] 🐛 Bug Fix

## Describe this PR

Download CSV feature was unreasonably slow when the number of projects was high - 8 minutes for 2700 projects.
I moved the expensive DB calls outside the loop into just one query. It should be much faster now.

## Review Guide

Use an DB with multiple users and lots of projects and task history. Preferably 2000+ projects in the search results. Press the download CSV button. Look for the right contributor count and the CSV file getting downloaded without a gateway timeout.

### DIFF
--- a/backend/models/postgis/project.py
+++ b/backend/models/postgis/project.py
@@ -206,13 +206,13 @@ class Project(db.Model):
     def percent_mapped(self):
         return (
             (self.tasks_mapped + self.tasks_validated)
-            / (self.total_tasks - self.tasks_bad_imagery)
             * 100
+            // (self.total_tasks - self.tasks_bad_imagery)
         )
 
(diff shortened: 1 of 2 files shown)


ramyaragupathy:
cc @mahesh-naxa @prabinoid @manjitapandey @royallsilwallz

mahesh-naxa:
I just tested this on my local setup (2700ish projects), the download was success for all projects and just below timeout. For prod this should be even lower.
Thanks @bshankar

mahesh-naxa:
@ramyaragupathy can be merged.

prabinoid (approved):

PR #6583, opened 2024-10-01
Set AMI to debian for staging and prod deployment

dakotabenjamin:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Describe this PR

This change sets the deployment backend server environment to use debian AMI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -294,7 +294,7 @@ workflows:
           name: Deploy backend production
           gitsha: $CIRCLE_SHA1
           stack_name: "tm4-production"
-          host_ami: "/aws/service/canonical/ubuntu/server/20.04/stable/current/amd64/hvm/ebs-gp2/ami-id"
+          host_ami: "/aws/service/debian/release/11/20240813-1838/amd64"
           backend_instance_type: c6a.large
           pg_version: "13.10"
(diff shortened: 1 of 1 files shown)


mahesh-naxa (approved):

.circleci/config.yml
```diff
@@ -357,7 +357,7 @@ workflows:
           name: Deploy backend production
           gitsha: $CIRCLE_SHA1
           stack_name: "tm4-production"
-          host_ami: "/aws/service/canonical/ubuntu/server/20.04/stable/current/amd64/hvm/ebs-gp2/ami-id"
+          host_ami: "/aws/service/debian/release/11/20240813-1838/amd64"
```
tsmock, line 360: Why not `/aws/service/debian/release/11/latest/amd64` (currently set to the 2024-10-07 Debian image)?

dakotabenjamin: My logic is that using latest could introduce a breaking or unintended change. Even at a low chance I would rather avoid it, and this code is going to be depreciated soon anyways so no long-term planning needed.

Issue #6631, opened 2024-11-14
Teams detail page should only be accessible to logged in users


manjitapandey:
**Describe the bug**

-  Username list of team's manager and members are currently accessible for public without needing to log in.  
- Also my teams option is available for users without logging in. And since there isn't user information available, clicking on my teams shows error. 
- Clicking on my team when not logged in leads you to view the list of all teams name and logo.

**To Reproduce**
1. Go to 'teams page through URL https://tasks.hotosm.org/teams/13/membership/'
2. Make sure you aren't logged into the system.
3. Click on 'my teams'

**Expected behavior**
User shouldn't be able to see the teams page without logging in. 

**Screenshots**
![image](https://github.com/user-attachments/assets/8bdbdbee-50fc-4b00-b89f-0f280eb23963)

**Additional context**
Related to #6541

ramyaragupathy:
In agreement with you @manjitapandey - user list for a team, list of Tasking Manager Teams should be accessible only for logged-in users.

Issue #6629, opened 2024-11-13
Draft projects are currently accessible to the public through teams page


manjitapandey:
**Describe the bug**
Related to #6541 , Draft projects are currently accessible to the public through teams page. 

**To Reproduce**
1. Go to 'teams page'
2. click on a team to go to teams details page
3. view the draft and archived projects too.

**Expected behavior**
Draft projects should not be accessible to public users. The project cards for draft projects should be hidden and onnly active projects should be shown instead. 

**Screenshots**
![image](https://github.com/user-attachments/assets/9aeb95a3-9344-40ef-adb9-1da027a44e85)

**Additional context**

SColchester:
Well spotted!

PR #6641, opened 2024-11-20
Fix Team Members page issues

### LINKED ISSUE #6541: Team members to see active projects instead of archived projects
**Describe the bug**
When opening a team page, team members see a list of projects that is assigned to their team.  This list however shows the archived projects at first. 

**To Reproduce & Expected behavior**
This is for example the link to the MSF validators team: https://tasks.hotosm.org/teams/13/membership/
On the right you see several projects, but they are all archived (see green circles on screenshot). Ideally they would see on this page instead of the archived projects, the active projects that they can map on at the moment. 

**Screenshots**
image

royallsilwallz:
## What type of PR is this?

- [x] 🐛 Bug Fix

## Related Issue

- Resolves  #6541, #6629,  #6631

## Describe this PR

This PR fixes the above mentioned issues.

**Changes**
- Fetch only `published projects` in `Team Members` page
- Do not allow user to access `Teams` page without login

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -30,9 +30,10 @@ jobs:
             - env
       - run:
           name: Run yarn test
+          no_output_timeout: 20m
           command: |
             cd ${CIRCLE_WORKING_DIRECTORY}/frontend/
-            CI=true yarn test -w 3 --silent
+            CI=true yarn test -w 1 --silent
             CI=true GENERATE_SOURCEMAP=false yarn build
 
   backend-code-check-PEP8:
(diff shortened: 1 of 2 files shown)


mahesh-naxa:
@dakotabenjamin the frontend tests was freezing due to no obvious reason, probably `-w` worker concurrency flag was too much for the VM, but still not sure how it was working previously. 
setting this to `-w 1` fixes it, but the tests take a minute long, i am testing this with github actions and see if things work there. 
cc: @mahesh-naxa

PR #6649, opened 2024-12-02
Tasking manager fastapi

prabinoid:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [x] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue
Related discussions:

https://github.com/hotosm/tasking-manager/discussions/6264

https://community.openstreetmap.org/t/migrating-tasking-manager-from-flask-to-fastapi-and-psycopg2-to-asyncpg/116543

## Describe this PR

This PR includes the TM fastapi migration and covers the commits from the initial phases of the migration.

This PR is made for the code review of the migration and is NOT be merged currently.

### DIFF
--- a/backend/api/annotations/resources.py
+++ b/backend/api/annotations/resources.py
@@ -1,14 +1,15 @@
+from databases import Database
+from fastapi import APIRouter, Depends, Request
+from loguru import logger
+from starlette.authentication import requires
+
+from backend.db import get_db
+from backend.models.dtos.user_dto import AuthUserDTO
 from backend.models.postgis.task import Task
 from backend.models.postgis.task_annotation import TaskAnnotation
 from backend.services.project_service import ProjectService
 from backend.services.task_annotations_service import TaskAnnotationsService
-from fastapi import APIRouter, Depends, Request
-from starlette.authentication import requires
-from loguru import logger
-from backend.db import get_db
-from databases import Database
 from backend.services.users.authentication_service import login_required
-from backend.models.dtos.user_dto import AuthUserDTO
 
 router = APIRouter(
     prefix="/projects",
--- a/backend/api/issues/resources.py
+++ b/backend/api/issues/resources.py
@@ -5,13 +5,12 @@ from loguru import logger
 
 from backend.models.dtos.mapping_issues_dto import MappingIssueCategoryDTO
 from backend.services.mapping_issues_service import MappingIssueCategoryService
-from backend.db import get_session
 from backend.db import get_db
 
 router = APIRouter(
     prefix="/tasks",
     tags=["issues"],
-    dependencies=[Depends(get_session)],
+    dependencies=[Depends(get_db)],
     responses={404: {"description": "Not found"}},
 )
 
--- a/backend/api/issues/resources.py
+++ b/backend/api/issues/resources.py
@@ -5,9 +5,7 @@ from loguru import logger
 
 from backend.models.dtos.mapping_issues_dto import MappingIssueCategoryDTO
 from backend.services.mapping_issues_service import MappingIssueCategoryService
-from backend.services.users.authentication_service import tm
 from backend.db import get_session
-from starlette.authentication import requires
 from backend.db import get_db
 
 router = APIRouter(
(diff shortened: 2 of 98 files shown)


backend/api/annotations/resources.py
```diff
@@ -1,153 +1,169 @@
...
 from backend.models.postgis.task_annotation import TaskAnnotation
 from backend.services.project_service import ProjectService
 from backend.services.task_annotations_service import TaskAnnotationsService
-from backend.services.application_service import ApplicationService
+from backend.services.users.authentication_service import login_required
 
+router = APIRouter(
+    prefix="/projects",
+    tags=["projects"],
+    dependencies=[Depends(get_db)],
```
spwoodcock, line 17: I'm not sure this is needed here. Some endpoints may not need a db connection and this would create one anyway. I'm not sure how you would access the database connection variable like this anyway

backend/api/annotations/resources.py
```diff
@@ -1,153 +1,169 @@
...
+    request: Request,
+    project_id: int,
+    annotation_type: str = None,
+    user: AuthUserDTO = Depends(login_required),
+    db: Database = Depends(get_db),
+):
+    """
+    Get all task annotations for a project
+    ---
+    tags:
```
spwoodcock, line 34: Entirely optional, but you could always delete this docstrings. Fastapi generates the OpenAPI docs using typing anyway, so it's a bit redundant. It also means you need to remember to update the docstring if you update variables

backend/api/issues/resources.py
```diff
@@ -1,219 +1,256 @@
...
+            description: Mapping-issue category found
+        404:
+            description: Mapping-issue category not found
+        500:
+            description: Internal Server Error
+    """
+    category_dto = await MappingIssueCategoryService.get_mapping_issue_category_as_dto(
+        category_id, db
+    )
+    return category_dto.model_dump(by_alias=True)
```
spwoodcock, line 50: This will work, but the preferred approach to a manual model dump is probably using `response_model` on the endpoint decorator. This will handle serialisation and dumping

backend/api/notifications/resources.py
```diff
@@ -1,241 +1,268 @@
...
+        500:
+            description: Internal Server Error
+    """
+    preferred_locale = request.headers.get("accept-language")
+    page = request.query_params.get("page", 1)
+    page_size = request.query_params.get("pageSize", 10)
+    sort_by = request.query_params.get("sortBy", "date")
+    sort_direction = request.query_params.get("sortDirection", "desc")
+    message_type = request.query_params.get("messageType", None)
+    from_username = request.query_params.get("from")
```
spwoodcock, line 192: This is an interesting approach for endpoint query params. It should work, but typically you would define query params as function arguments and they would be parsed as such. If there is confusion, it's possible to explicitly define arguments as fastapi.Query

backend/api/organisations/campaigns.py
```diff
@@ -1,155 +1,199 @@
...
+            )
+            return JSONResponse(
+                content={"Error": message, "SubCode": "CampaignAlreadyAssigned"},
+                status_code=400,
+            )
+        async with db.transaction():
+            await CampaignService.create_campaign_organisation(
+                organisation_id, campaign_id, db
+            )
             message = "campaign with id {} assigned for organisation with id {}".format(
```
spwoodcock, line 83: F-strings are a bit nicer to read, but not a big deal 👍

backend/api/system/applications.py
```diff
@@ -1,132 +1,155 @@
...
+      302:
+        description: User is not authorized to create a key
+      500:
+        description: A problem occurred
+    """
+    token = await ApplicationService.create_token(user.id, db)
+    return token.model_dump(by_alias=True)
+
+
+@router.patch("/authentication/applications/{application_key}/")
```
spwoodcock, line 82: Using a PATCH method here seems a bit out of place, but maybe I am missing something

spwoodcock:
That's a lot of code changed! Well done on this - lots of great work 🙌

Just a few minor comments

PR #6716, opened 2025-01-25
Typescript support for react-redux selector and dispatch

royallsilwallz:
## What type of PR is this?

- [x] 🧑‍💻 Refactor

## Describe this PR
The code changes in the PR will remove the requirement to specify type for redux state in each selector.
This results in better typescript code.

**Changes**
- Setup Typescript support for `useSelector` & `useDispatch` hooks
- Configure import path aliases for folders

### DIFF
--- a/frontend/vite.config.ts
+++ b/frontend/vite.config.ts
@@ -1,26 +1,12 @@
 // vite.config.js
 import { defineConfig } from 'vite';
 import react from '@vitejs/plugin-react-swc';
+import tsconfigPaths from 'vite-tsconfig-paths'
 
 export default defineConfig({
-  plugins: [react()],
+  plugins: [react(), tsconfigPaths()],
   envPrefix: 'REACT_APP_',
   server: {
     port: 3000
-  },
-  resolve: {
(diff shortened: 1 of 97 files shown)


frontend/vite.config.ts
JoltCode, line 1: Can we use https://www.npmjs.com/package/vite-tsconfig-paths instead?

frontend/vitest.config.ts
JoltCode, line 1: Same as the vite config comment (Can we use https://www.npmjs.com/package/vite-tsconfig-paths instead?)

JoltCode (changes requested):
Awesome work! - Just one thing, perhaps we should use a vite plugin instead of declaring the paths manually to ensure our paths are a single point of truth?

royallsilwallz:
> Can we use https://www.npmjs.com/package/vite-tsconfig-paths instead?

Yes, sure! I didn't know of this.. I'll do it. Thanks! :rocket:

JoltCode (approved):
LGTM! Great work 👍

PR #6802, opened 2025-04-09
infra: Add bastion ec2

nischalstha9:

- [X] 🍕 Feature
- [X] 🏗️ Infra

## Related Issue
- Resolves #6802 

## Describe this PR

- Adds ec2 resources as Bastion Host to connect to AWS resources


dakotabenjamin (changes requested):
We need a couple more tags (`project` and `tool`) and probably a way to automate authenticated access. I've sent a message on slack about possibly installing JumpCloud agent in the UserData.

nischalstha9:
The tags as project, application, environment are auto added from default tags with terragrunt. Also, I have integrated user-data feature to auto add ec2 into jump cloud. We will need to add an secret variable `JUMPCLOUD_CONNECT_KEY` and run `terragrunt apply` for `extras` module once before running apply in EC2 module.

PR #6915, opened 2025-06-30
Migrate `mapbox-gl` to `maplibre-gl`

### LINKED ISSUE #6719: Replace `mapbox-gl` with `maplibre-gl` on the frontend

- Currently we use `mapbox-gl@1.13.3`, the final version before they switched from BSD3 to their proprietary license. This version is 5yrs out of date by now.
- We also use a few mapbox plugins: draw, geocoder, language (plus an ancient plugin mapbox-gl-draw-rectangle-mode that is probably no longer needed).

A migration plan from mapbox --> maplibre, taking us into the future:
- Replace mapbox-gl with maplibre-gl. Hopefully this part is easy and it's mostly a drop-in replacement.
- Replace `@mapbox/mapbox-gl-geocoder` with `@maplibre/maplibre-gl-geocoder`.
- `@mapbox/mapbox-gl-draw` is still open and probably fine to use, but we could consider a migration to TerraDraw if needed.
- Uncertain about `@mapbox/mapbox-gl-language` for now - I feel this is probably simply built into the maplibre style spec by now - but could potentially keep using this too if needed as it's open BSD-3 license.

- There aren't other great alternatives with a clear migration pathway.
- As our map usage is pretty limited, we could consider leaflet, but would lose the WebGL performance gains.

**Additional context**

- Requirement for approval as a Digital Public Good.

royallsilwallz:
## What type of PR is this?

- [x] 🍕 Feature
- [x] 🧑‍💻 Refactor

## Related Issue

- Resolves #6719 

## Describe this PR
This PR migrates the mapping library `mapbox-gl` to `maplibre-gl`.

**Changes**
- Replace `mapbox-gl` with `maplibre-gl`
- Replace `@mapbox/mapbox-gl-geocoder` with `@maplibre/maplibre-gl-geocoder`
- Update `MAPBOX_RTL_PLUGIN_URL`
- Address test cases failure


frontend/src/utils/isWebglSupported.js
spwoodcock, line 1: Do you think the original check for web-gl was a concern about browser support, or for if users have it disabled?

I don't think we need to worry about **browser support**, as it's pretty ubiquitous, even on older browser: https://caniuse.com/?search=webgl

However, some weird people like me do sometimes use a browser with hardened security and things like **web-gl disabled**. Do you think this is a major concern we should add the extra check for though?

spwoodcock (approved):
Excellent work! This looks good to me, except if you want to refactor out the web-gl check 👍 

Are the broken CI tests are fixable too?

spwoodcock (approved):

frontend/src/utils/isWebglSupported.js
royallsilwallz: Hi @spwoodcock, thanks for the review! 
For the webgl check, it was already there so I just modified it for maplibre gl. I'll have a detailed look and see if it would be required.

royallsilwallz: Hey @spwoodcock, I removed the WebGL check, but the test cases were failing - it was related to test setup not being able to find Map methods (Map.addControl.. etc). Every test containing map was failing and I was unable to resolve them for now. So, I kept it whatsoever. :shrug:

PR #6938, opened 2025-07-08
Make auth callback URL relative to base API_URL

jake-low:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Describe this PR

This is a tweak to the changes made in #6933. That PR works correctly AFAIK for tasks.hotosm.org's use case, but I discovered when backporting the fix to [osmus/tasking-manager](https://github.com/osmus/tasking-manager) (which powers tasks.openstreetmap.us and tasks.teachosm.org) that it does not work if the backend is deployed to a nonstandard URL path.

This change makes the callback URL string a relative path (removing the hardcoded `/api/v2` prefix). The appropriate prefix will then be [prepended in `fetchLocalJSONAPI()`](https://github.com/hotosm/tasking-manager/blob/develop/frontend/src/network/genericJSONRequest.js#L27) when the string is converted to a `URL` object. This prefix may be `/api/v2` or it may be another value depending on the configuration environment variables provided at runtime.

This change is running in production on tasks.openstreetmap.us and tasks.teachosm.org already, so I am fairly confident that it works.

### DIFF
--- a/frontend/src/views/authorized.js
+++ b/frontend/src/views/authorized.js
@@ -10,7 +10,7 @@ import { AnimatedLoadingIcon } from '../components/button';
 export function Authorized(props) {
 
   const authComplete = (authCode, state) => {
-      let callback_url = `/api/v2/system/authentication/callback/?redirect_uri=${OSM_REDIRECT_URI}&code=${authCode}`;
+      let callback_url = `system/authentication/callback/?redirect_uri=${OSM_REDIRECT_URI}&code=${authCode}`;
       const emailAddress = getItem('email_address');
       if (emailAddress !== null) {
         callback_url += `&email_address=${emailAddress}`;


ramyaragupathy:
@suzit-10 can you please review this PR?

suzit-10:
Thank you @jake-low for raising this 🙏
Removing the hardcoded `/api/v2` is definitely a good improvement, and it’s very helpful for instances with non-standard backend URL (e.g., without /api/v2). 
Besides the authorization URL, there are a few other places with the same prefix — could you update those too? I think it will also need a small tweak to `fetchLocalJSONAPI()`, since it’s used across the codebase, and simply prepending may not work in all cases (e.g., where `api/v2` isn’t expected).

Really appreciate you taking the time to look into this 🙌

jake-low:
Hi @suzit-10, thanks for reviewing this. I unfortunately don't have time to revisit this PR and modify or retest it. I believe it fixes a genuine bug, although it may not fix all occurrences of that bug since I was focused only on fixing the code introduced in #6933.

Feel free to either merge this PR or close it at your discretion.

suzit-10:
Hi @jake-low, thanks for the update! I appreciate you taking the time to address this. I'll proceed with the required changes across the code and merge the PR.

jake-low:
This was superseded by #7046 which has been merged and released, so I'll close this one.

PR #6942, opened 2025-07-10
Replace requests with httpx client for non blocking execution

prabinoid:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [x] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Describe this PR

Using requests (a sync lib) in an async context might block the thread, degrading performance. 
Replace requests with httpx client for non blocking execution.

### DIFF
--- a/backend/api/users/statistics.py
+++ b/backend/api/users/statistics.py
@@ -1,6 +1,6 @@
 from datetime import date, timedelta
 
-import requests
+import httpx
 from databases import Database
 from fastapi import APIRouter, Depends, Request, Query
 from fastapi.responses import JSONResponse
@@ -13,6 +13,7 @@ from backend.services.interests_service import InterestService
 from backend.services.stats_service import StatsService
 from backend.services.users.authentication_service import login_required
 from backend.services.users.user_service import UserService
+
 
 router = APIRouter(
     prefix="/users",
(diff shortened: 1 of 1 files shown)


backend/api/users/statistics.py
```diff
@@ -175,39 +176,38 @@ async def get_ohsome_stats(
...
         )
+
+    headers = {"Authorization": f"Basic {settings.OHSOME_STATS_TOKEN}"}
+
     try:
-        headers = {"Authorization": f"Basic {settings.OHSOME_STATS_TOKEN}"}
-        # Make the GET request with headers
-        response = requests.get(url, headers=headers)
+        async with httpx.AsyncClient() as client:
+            response = await client.get(url, headers=headers, timeout=10.0)
```
spwoodcock, line 191: This change is great, but it flagged an error - rightly so too.

The user could pass in the URL to a malicious script here easily.

Can we pass in something like the user id here and construct the URL on the backend?

(and on that note, do we pass a URL to any other endpoints in TM?)

spwoodcock (changes requested):

backend/api/users/statistics.py
prabinoid: Thank you Sam. We allow redirect uri to be passed in login. Although, by default, it is being fetched from config but can be overridden.

prabinoid:
Closing as  its captured in [PR 6943](https://github.com/hotosm/tasking-manager/pull/6943)  where the updated ohsome apis are implemented.

PR #7003, opened 2025-08-14
Add option to immediately exit cron_jobs script

jacobwhall:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Describe this PR

Adds a `--immediate-exit` option to the `backend/cron_job.py` script, which exits immediately upon completing the jobs. I additionally pulled the sleep statement out of the try block, so that the database connection wouldn't stay open for 10 minutes.

I would prefer for the script to exit immediately by default, but didn't want to break anything upstream.

## Review Guide

My intention was for this to be entirely backwards compatible. I hope that the `backend/cron_jobs.py` script runs the same as before, unless the user opts-in to the immediate exit. I'm not entirely sure how Tasking Manager runs in AWS, so I want to be sure I don't break that functionality.

### DIFF
--- a/backend/cron_jobs.py
+++ b/backend/cron_jobs.py
@@ -157,7 +157,17 @@ async def main():
         await db_connection.database.connect()
         logger.info("Database connection established.")
 
-        await setup_cron_jobs()
+        if args.immediate_exit:
+            # Run each job once, and that's it
+            await auto_unlock_tasks()
+            await update_all_project_stats()
+            await update_recent_updated_project_stats()
+        else:
+            # Set up a scheduler and run it indefinitely
+            await setup_cron_jobs()
+            # Keeping the process alive.
+            while True:
+                await asyncio.sleep(3600)
 
     except (KeyboardInterrupt, SystemExit):
         logger.info("Shutting down...")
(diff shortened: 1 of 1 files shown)


jacobwhall:
I am happy to provide a test for this if you feel it's appropriate.

prabinoid:
Hi @jacobwhall, Thank you for the contribution! I like the idea of an --immediate-exit option and not holding the DB connection open unnecessarily. However, as this PR stands it seems it introduces a lifecycle bug that breaks scheduled jobs.

The refactor calls await db_connection.database.connect() then await setup_cron_jobs() inside the try, but the finally: immediately calls await db_connection.database.disconnect() before the script goes into the long while True: await asyncio.sleep(...) that keeps the process alive to run the jobs periodically.

Because disconnect() runs immediately, the scheduler is still active but the DB connection pool is closed. When a job runs it attempts to open a connection and fails with DatabaseBackend is not running. If this is handled, adding the immediate exit sounds good 👍.

jacobwhall:
**edit: I must be tired. more coherent comment incoming.**

jacobwhall:
Hey @prabinoid,

Yep that makes sense. I read this code a bit too quickly and missed some important details. I just pushed another commit (and force-pushed to keep whitespace consistent with upstream) which will now leave the scheduler alone when `--immediate-exit` is not used.

prabinoid:
Thanks @jacobwhall 🙌🏻 ! Looks good to me 👍 cc: @nischalstha9

nischalstha9 (approved):

PR #7013, opened 2025-08-19
Restore three requested functionality related to super mapper

categulario:
* [ ] allow updating the beginner level without setting required badges (and stop showing badges ui for that level)
* [ ] prevent an overlapping in the level select for project permissions
* [ ] allow clearing the level filter in the users table without clearing all of the filters
* [ ] fix projects table failing when listing projects to map or validate


emi420 (approved):
Looks good @categulario ! let's test it on the dev sever directly.

PR #7014, opened 2025-08-22
fix problem when deleting level with votes

categulario:
Uses CASCADE for relationships that come from voting related tables so that they don't interfere when deleting a level with no users.


emi420 (approved):
👍 all good @categulario , approved for testing

PR #7018, opened 2025-08-29
Fix/script formatting

prabinoid:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [x] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Describe this PR
Previously, we had a script to check and update mapper level, which is no longer relevant due to the super mapper related Mapping level enum changes. Updated the script to save the current user stats and check for mapper level upgrades. 

The script fetches the user stats but allows to omit the badges assignment and mapper level upgrades on exception.

Also, updated files to fix pre-commit errors.


scripts/commands/refresh_mapper_level.py
```diff
@@ -0,0 +1,98 @@
+import asyncio
```
categulario, line 1: I was just wondering why wasn't it better to call the `check_and_update_mapping_level` function an option. Having this code duplicated creates the possible problem in the future that a change in the logic hast to be made twice...

categulario (commented):
Just the one comment about the duplicated code

scripts/commands/refresh_mapper_level.py
prabinoid: When i tested initially, the advanced mappers were not promoted when i used that function. But it seems the correct way to do was create one more mapping level higher than advanced mapper and then use the same script. Will update to use the same function. Thanks!

PR #7064, opened 2025-11-03
Refactor pm_only dependency to admin_only

### LINKED ISSUE #6990: Dependency pm_only should be named admin_only
While reviewing a PR, I see a dependency called `pm_only`, defined here:

https://github.com/hotosm/tasking-manager/blob/71a753178a4be66a9f8ef865aeee3d95a7c4896c/backend/services/users/authentication_service.py#L286

Its name, and a comment in [other part of the code](https://github.com/hotosm/tasking-manager/blob/71a753178a4be66a9f8ef865aeee3d95a7c4896c/backend/api/utils.py#L13) suggest that this dependency is used for restrict actions to users with the role "Project Manager", but actually the restriction is for users with the role "Admin".

The name should be changed to `admin_only` and the code where it is used changed accordingly.

prabinoid:

- [ ] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [x] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Fixes #6990

## Describe this PR

pm only dependency is refactored to admin only as it validates admin role only.

### DIFF
--- a/backend/api/issues/resources.py
+++ b/backend/api/issues/resources.py
@@ -7,7 +7,7 @@ from backend.db import get_db
 from backend.models.dtos.mapping_issues_dto import MappingIssueCategoryDTO
 from backend.models.dtos.user_dto import AuthUserDTO
 from backend.services.mapping_issues_service import MappingIssueCategoryService
-from backend.services.users.authentication_service import pm_only
+from backend.services.users.authentication_service import admin_only
 
 router = APIRouter(
     prefix="/tasks",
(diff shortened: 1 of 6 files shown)


ramyaragupathy:
Since it is a naming change convention, good to have one testing done on admin functionalities across. Tagging @manjitapandey for review.

PR #7070, opened 2025-11-06
PR 6334 User deletion

### LINKED ISSUE #6002: Add function to delete user accounts

prabinoid:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Fixes #6002

## Describe this PR
Add ability to delete user information.

Refactored from [pr-6334](https://github.com/hotosm/tasking-manager/pull/6334)
Rebased the pr against current develop and refactored it.

Adds an api which allows user to delete their own account and admins to delete any user.
The function that automatically deletes users that are gone from osm is kept but not implemented as an api.

### DIFF
--- a/frontend/src/components/deleteModal/index.js
+++ b/frontend/src/components/deleteModal/index.js
@@ -128,7 +128,11 @@ export function DeleteModal({
                     <FormattedMessage {...messages.cancel} />
                   </Button>
                   <Button className="bg-red white" onClick={() => deleteEntity()}>
-                    <FormattedMessage {...message} />
+                    {type === 'users' ? (
+                      <FormattedMessage {...messages.deleteUser} />
+                    ) : (
+                      <FormattedMessage {...message} />
+                    )}
                   </Button>
                 </div>
               </>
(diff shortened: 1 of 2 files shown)


ramyaragupathy:
image


This is the message end user receives upon selecting delete user action.


As discussed, here are my suggestions:


Are you sure you want to permanently delete this user?

_All user information, history, and associated data will be deleted forever.
This action cannot be reversed._

CTA:
- Cancel
- Yes, delete permanently

@prabinoid @manjitapandey @suzit-10  - let me know what do you think.

suzit-10:
This is the new confirmation dialog after updating the message:

image



cc: @ramyaragupathy , @prabinoid , @manjitapandey

PR #7122, opened 2025-12-31
Add plain text alternative to email messages

### LINKED ISSUE #5595: [BUG] Validation mail has no text part
**Describe the bug**
Mail with subject such like: "Task XXXX mapped by you in Project ...has been validated" has no text part, only html part.
So if a mail client shows the text part by default, no content is displayed (In my case, I use mutt).

**To Reproduce**
The original mail I got is attached (utf-8 encoded)

**Expected behavior**
Some description should be in text part of mail, because `Content-Type: multipart/mixed` is in header.

**Screenshots**
Screenshot_2023-02-19-20-12-03_1920x1080

heathdutton:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Fixes #5595

## Describe this PR

Emails sent by Tasking Manager only include an HTML body. Email clients that display plain text by default show empty messages.

This PR adds a plain text alternative to all outgoing emails by:
- Using FastAPI-Mail's `alternative_body` parameter with `multipart_subtype=alternative`
- Auto-generating plain text from HTML when not explicitly provided

## Review Guide

The change is in `backend/services/messaging/smtp_service.py`. The `_send_message` method now sends multipart/alternative emails with both HTML and plain text parts.

### DIFF
--- a/backend/services/messaging/smtp_service.py
+++ b/backend/services/messaging/smtp_service.py
@@ -207,9 +207,9 @@ class SMTPService:
         msg = MessageSchema(
             recipients=[to_address],
             subject=subject,
-            body=html_message,
-            alternative_body=text_message,
-            subtype=MessageType.html,
+            body=text_message,
+            alternative_body=html_message,
+            subtype=MessageType.plain,
             multipart_subtype=MultipartSubtypeEnum.alternative,
         )
         logger.debug(f"Sending email via SMTP {to_address}")


prabinoid:
Thank you @heathdutton  for this. But, it seems like the emails received are on plain text even when email clients supports rich texts like html. Is it the case? I think we might want to keep the html by default for sent messages as most clients do support formatted/Rich text by default.

heathdutton:
@prabinoid The original issue #5595 was that plain-text email clients (like mutt) showed empty messages because emails only had HTML, no plain text fallback.

The fix should send both HTML and plain text in a multipart/alternative structure, where the client picks its preferred format. HTML clients should still see HTML.

prabinoid:
Hi @heathdutton , I tested out the PR and it seems gmail (supports html by default) shows only the plain text part and doesn't render out html which should be otherwise as shown in other image: 
Screenshot 2026-02-02 at 11 46 27
Screenshot 2026-02-02 at 11 51 22

From Gmail → Show original, it looks like the multipart/alternative is malformed — the HTML part appears before the plain text, and the multipart container itself seems to be base64-encoded. Maybe that’s causing Gmail to fall back to the plain text? Would ensuring text/plain comes first, text/html second, and that the multipart container isn’t base64-encoded solve the issue?

heathdutton:
Sorry for stalling, just a bit confused because FastAPI-Mail attaches HTML then plaintext, but RFC 2046 says that the preferred format should be last. I think maybe we should swap body and alternative_body for now rather than try to fix fastapi-mail. Standby.

https://www.rfc-editor.org/rfc/rfc2046.html

PR #7123, opened 2026-01-01
Extend markdown attributes to support various markdown features

### LINKED ISSUE #6685: [BUG] The Tasking Manager does not actually properly/fully support Markdown
**Describe the bug**

The text highlighted in the following image would have you believe the TM fully supports markdown formatting.

Image

### IT LIES! 

**To Reproduce**
1. Go to '[Markdown Guide](https://www.markdownguide.org/)'
2. Go through that guide and (one by one) try to use each feature of markdown in a comment.
3. See error

Alt.

find a project with posted comments and see that the author and manager hashtags have been incorrectly formatted as headings. I see this very offten and I'm not sure but it's possible that formatting the hashtags as headings means that project authors and managers may not recieve notifications regarding comments on their projects and in tasks.

Image
hashtags incorrectly formatted as a heading (even with text following the hashtag).

**Expected behavior**

All markdown formatting is supported.

In the case of hashtags and headings: `#character` results in "#character"; and only `# character` results in 
### character. 
(refer to the headings section of the Markdown guide)

**Screenshots**
This is an example of posting a comment with a table in the TM.

Image
Generating markdown table code
Image
Checking the preview in an IDE
Image
In the comments box on the TM
Image
Preview in the TM
Image
Posted on the TM

 - OS: [Windows 10]
 - Browser [Firefox]
 - Version [134]

**Additional context**
There are other formatting functions that do not work like adding annotations to images and probably others. Adding annotations is desirable becasue It makes for better image comparison as they are side by side rather than adding a description between the two. 

I recommend that when working on this issue the Markdown guide is followed so that all markdown formatting functionallity can be systematically addressed. 

I suggest that the first thing to be fixed is the incorrect formatting of hashtags as headers, becasue I see it all the time and it may affect project project management and co-ordination.

Website used to generate [Markdown table code](https://www.tablesgenerator.com/markdown_tables#).

The TM preview is not always accurate as in the case of the table. I believe that when I tried to add annotations they were present in the preview but did not appear in the posted comment.

Adding this functionallity will also help project creators to better format instructions and the like.

prabinoid:

- [x] 🍕 Feature
- [ ] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Fixes #6685

## Describe this PR

Additional markdown tags and support in project comments.

### DIFF
--- a/frontend/src/components/comments/commentInput.js
+++ b/frontend/src/components/comments/commentInput.js
@@ -145,6 +145,8 @@ function CommentInputField({
     sessionStorage.setItem(sessionkey, comment);
   }, [comment, sessionkey]);
 
+  console.log(comment, 'comment');
+
   return (
     <div {...getRootProps()}>
       {isShowTabNavs && (
@@ -193,12 +195,12 @@ function CommentInputField({
       <div className={`bg-white`} data-color-mode="light">
         {/* <div className={`${isShowPreview ? 'dn' : ''} bg-white`} data-color-mode="light">*/}
 
-        <div className={`${isShowPreview ? '' : 'dn'}`}>
+        <div className={`${isShowPreview && comment ? '' : 'dn'}`}>
           <MDEditor
             preview={'preview'}
             commands={Object.keys(iconConfig).map((key) => iconConfig[key])}
             extraCommands={[]}
-            height={500}
+            height={200}
             value={formatUserNamesToLink(comment)}
             onChange={setComment}
             textareaProps={{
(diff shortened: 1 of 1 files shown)


ramyaragupathy:
Task list seems to be working in the preview, but does not render as expected after psoting the comment.
We can track this in a separate issue.

One other item for improvement could be sub items in the bullet. Again, that can be an enhancement in a later release.

cc @prabinoid @suzit-10

PR #7124, opened 2026-01-02
Fix task list page jumping to wrong page on selection with sort

### LINKED ISSUE #6850: [BUG] Erroneous Pages are Displayed in the Task Menu in the LHS Contribute Page Panel when Selecting Untouched Tasks and Sorting by Least Recently Updated
**Describe the bug**
When sorting by least recently updated the page goes to one where the selected task is not visible making it tedious to use the task functions like view history, zoom to task and link to task.

**To Reproduce**
1. Go to a project with untouched tasks available
2. Click on Tasks
3. Filter for available for mapping and sort by least recently updated
4. Select an untouched task either via the map or task menu
5. See error
**Expected behavior**
The selected task should be visible on both the map and in the task menu.

**Screenshots**

Image

 - OS: [Windows 10]
 - Browser [Firefox]
 - Version [138.0.1]

**Additional context**
It is possible to work around this problem by sorting by most recently updated and working backwards, or by filtering for the task ID which will only result in one possible page. 

I'm unsure if this is always the case, but searching for the task by going through the pages manually does not allow you to find the task in the menu (it may just flash for a brief moment.

heathdutton:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Fixes #6850

## Describe this PR

When selecting a task while sorting by "least recently updated", the task list would jump to the wrong page, making the selected task not visible in the list.

The issue was caused by a stale ref pattern in the `PaginatedList` component. The page calculation used `latestItems.current` which could be out of sync with the actual sorted items array due to React's effect ordering.

The fix uses `items` directly in the useEffect and adds it to the dependency array, ensuring the page calculation always uses the current sorted list.

## Review Guide

1. Go to any project with tasks
2. Click on Tasks tab
3. Filter for "Available for mapping" and sort by "Least recently updated"
4. Select a task from the map or list
5. Verify the task list page navigates to show the selected task

### DIFF
--- a/frontend/src/components/taskSelection/taskList.js
+++ b/frontend/src/components/taskSelection/taskList.js
@@ -1,4 +1,4 @@
-import { useState, useEffect, useCallback } from 'react';
+import { useState, useEffect, useRef, useCallback } from 'react';
 import { useLocation } from 'react-router-dom';
 import Popup from 'reactjs-popup';
 import { useQueryParam, NumberParam, StringParam } from 'use-query-params';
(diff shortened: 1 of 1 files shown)


suzit-10:
Hi! @heathdutton, thanks for the PR! While testing the functionality, it looks like the pagination behavior has regressed — clicking on a page number no longer loads that page as it did before. Also, some FE tests are failing. Could you please take a look at restoring the pagination logic and fixing the test failures? Thanks!

heathdutton:
Fixed. The ref is now updated synchronously during render instead of in an effect, keeping items fresh while only triggering on selection changes.

suzit-10:
Thanks @heathdutton, for the fix! It’s working as expected now, and the regression is resolved. The FE test failure I mentioned earlier turned out to be an environment/permission issue on our side, not related to this PR.

Everything looks good from my end now. Thanks for the quick update.

Issue #7188, opened 2026-03-09
[HOTFIX] Prevent login failure when external OSM/OhSome stats fetch returns 429


ramyaragupathy:
Current implementation of user stats fetching during login raises an exception if the external OSM or OhSome API returns a 429 Too Many Requests error. This is currently blocking users from logging into the platform.

Proposed Solution: Merge [PR #7187](https://www.google.com/search?q=https://github.com/hotosm/tasking-manager/pull/7187) which:

  *  Returns an empty stats object instead of crashing.

   * Increases log verbosity to capture the error details.

prabinoid:
The stats update calls https://www.openstreetmap.org/api/0.6/user/{user_id}.json which is currently being blocked with 429 Too many requests which needs be passed in order to fetch and update the stats and upgrade the badges and mapping level.

ramyaragupathy:
Fixed through https://github.com/hotosm/tasking-manager/pull/7187 and https://github.com/hotosm/tasking-manager/releases/tag/v5.5.1

PR #7191, opened 2026-03-09
fix: Replace httpx async client with non blocking sync request

prabinoid:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [x] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

#7188

## Describe this PR

The request made to osm user detail api ([example](https://www.openstreetmap.org/api/0.6/user/21754844.json)) returns 429 Too many requests when called from httpx AsyncClient. 

Made the request using requests library as non blocking to prevent blocking of the event loop during execution.

Tried to delay the request with 3 seconds sleep between the requests to check if rate limiting caused this issue, the issue still persisted.

Tried User-Agent in httpx async client, still the issue remained same.

Need to figure out the root of the issue, whether it is rate limiting or something else.

### DIFF
--- a/backend/services/users/user_service.py
+++ b/backend/services/users/user_service.py
@@ -5,7 +5,8 @@ from typing import List, Optional
 from databases import Database
 from loguru import logger
 from sqlalchemy import and_, desc, distinct, func, insert, select
-from httpx import AsyncClient
+import requests
+import asyncio
 
 from backend.config import Settings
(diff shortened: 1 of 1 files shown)


prabinoid:
Fixed through #7194.

PR #7208, opened 2026-03-31
Fixed Translation Bug On Task Completeness Question

### LINKED ISSUE #7108: [BUG] rare translation issue on task completness question
**Describe the bug**
On a mapathon one of the users had this bizzare issue. See the photo below.... They locked a task for mapping, they had their HOT TM website set to **Czech language** but when they locked a task it switched the task completness page to english and on the question "Is this task completely mapped?" they had two options "Again" and "Yes" (instead of "Ano" and "Ne" in english "Yes" nad "No"). After clicking on "Again" worked as if they click on "Yes" and when they clicked on "Yes" it worked as if they select "No". So more likely just translation strings got messed up for some reason. 
Also the grey text under "COMPLETION" and above "Is this task completey mapped?" was strangely "JOB STATUS" which in regular english version is "TASK STATUS". 
**There were many other mappers using the HOT TM in czech language and for everyone else it worked as intended.** 

**To Reproduce**
**I tried to reproduce the issue but I was not able**. I was on the mapathon in person, I am the one who took the picture, I did everything the mapper told me they did but I was not able to reproduce the issue. And I never seen it before or after or with another user so I believe it was a very rare occasion. 
1. Set your HOT TM language to Czech
2. Lock a task for mapping

**Expected behavior**
Have the page in czech language.

**Screenshots**
Photo of the issue from mapathon:
Image

Here is side by side how it usually looks like. Left is HOT TM set to czech language, right is english:

Image

 - OS: Windows or Linux  
 - Browser:  no idea
 - HOT TM Version: v5.3.1

**Additional context**
This happened 20.11.2025. I am opening this ticket 2 weeks after it happened because I tried to contact the user for more informations but they didn't replied to me. I can speak czech, I took the photo of their screen. They were mapping in JOSM.

PiRaNg-1256:
## What type of PR is this?

Bug Fix

## Related Issue

Fixes #7108 

## Describe this PR

This PR fixes a reported "bizarre" translation issue where Czech users saw incorrect terms like "JOB STATUS" (instead of TASK STATUS), "Again" (instead of Yes), and "Yes" (instead of No).

Root Cause: The application was missing the lang attribute on the <html> element. This caused browsers (specifically Chrome) to attempt auto-detection of the page language. In some cases, the browser incorrectly guessed the language or attempted to auto-translate short strings (like "Ano"/"Ne") into English using incorrect mappings, leading to hallucinations like "Again" and "Yes".

Solution:

index.html: Added a default lang="en" to the <html> tag.
App.js: Added a useEffect hook to dynamically update document.documentElement.lang whenever the application locale changes. This informs the browser exactly what language is being displayed, preventing inappropriate auto-translation triggers.

## Review Guide

Open the application and change the language to Czech (or any other language).
Inspect the page source/DOM using Developer Tools.
Verify that the <html> tag has the correct lang attribute (e.g., <html lang="cs">).
Switch back to English and verify it updates to <html lang="en">.

### DIFF
--- a/frontend/public/index.html
+++ b/frontend/public/index.html
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html>
+<html lang="en">
   <head>
     <meta charset="utf-8" />
     <link rel="shortcut icon" href="%PUBLIC_URL%/static/favicon.ico" />
--- a/frontend/src/App.js
+++ b/frontend/src/App.js
@@ -48,6 +48,10 @@ const App = () => {
     store.dispatch(getUserDetails(store.getState()));
   }, []);
 
+  useEffect(() => {
+    document.documentElement.lang = locale;
+  }, [locale]);
+
   return (
     <ErrorBoundary fallback={<FallbackComponent />}>
       {isLoading ? (
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,6 @@
+{
+  "name": "tasking-manager",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {}
+}


sumitdahal7 (changes requested):
Thanks for the fix. This is worth doing since setting the document `lang` helps prevent browser auto-translation issues and improves accessibility.

A couple of changes before merge:

1. Please normalize the locale before assigning it to `document.documentElement.lang`. Some supported
locales use underscores internally, e.g. `nl_NL` and `zh_TW`, but HTML `lang` expects BCP 47 tags like
`nl-NL` and `zh-TW`.

```js
document.documentElement.lang = locale ? locale.replace(/_/g, '-') : 'en';
```

2. App.js already has a root wrapper with lang={locale}. Please either apply the same normalized value
there too, or remove the div-level lang if <html lang> is now the source of truth.
3. Please remove the package-lock.json change. This repo uses Yarn (frontend/yarn.lock), and this fix
should not require dependency changes.

PR #7044, opened 2025-10-06
Fix Missing task extension duration in action_text.

Aadesh-Baral:
## What type of PR is this? (check all applicable)

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI
- [ ] ❓ Other (please specify)

## Related Issue

Example: Fixes #6014

## Describe this PR

Refactor of #6046

PR #7214, opened 2026-04-03
fix: Lock duration in extension task history action text and test cas…

### LINKED ISSUE #6014: [BUG] Missing Task Lock Extension Duration in Action Text Column

prabinoid:
…es refactor/fix

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Fixes #6014

## Describe this PR

Refactor of #7044

The action text was missing in EXTEND_FOR_MAPPING and EXTEND_FOR_VALIDATION actions which were later introduced.

Previously, all the Task actions and status were same so the based on the current task status the actions were being patched in the action text accordingly. However, in case of extending mapping or validation, the equivalent status weren't there due to which no tasks were being fetched and hence the timestamps was not being patched.

The task is always relocked when the session is expired for mapping or validation. This causes redundant entries for extension in the history table. These redundant entries are recursively deleted and only one entry is patched with the total extended time.
This recursive deletion function is legacy code which was introduced due to race condition creating multiple entries. This handles the multiple redundant entries for extension currently though this relock mechanism needs to be handled.

Ideal scenario for multiple extension of task duration would be, everytime the task is relocked, the extended session is patched and then again the task is locked and extended for the next phase of extension.

The test cases are refactored too as previously, the test cases didn't check the last entry of the task history but instead just checked the task status. Now when it tried to fetch the last entry, the tests that directly updated the status with sql, now didn't find the history entry and hence were refactored accordingly.

### DIFF
--- a/backend/models/postgis/task.py
+++ b/backend/models/postgis/task.py
@@ -383,15 +383,22 @@ class TaskHistory(Base):
         try:
             # Fetch the last locked task history entry with raw SQL
             query = """
-                SELECT id, action_date
-                FROM task_history
-                WHERE task_id = :task_id
-                AND project_id = :project_id
-                AND action = :action
-                AND action_text IS NULL
-                AND user_id = :user_id
-                ORDER BY action_date DESC
-                LIMIT 1
+                WITH locked_rows AS (
+                    SELECT
+                        id,
+                        action_date,
+                        COUNT(*) OVER() AS total_rows
+                    FROM task_history
+                    WHERE task_id = :task_id
+                    AND project_id = :project_id
+                    AND action = :action
+                    AND action_text IS NULL
+                    AND user_id = :user_id
+                    ORDER BY action_date DESC, id DESC
+                    FETCH FIRST 1 ROW ONLY
+                )
+                SELECT id, action_date, total_rows
+                FROM locked_rows
             """
             values = {
                 "task_id": task_id,
@@ -409,6 +416,9 @@ class TaskHistory(Base):
                 # rather than showing the user an error that they can't fix.
                 # No record found, possibly a race condition or auto-unlock scenario.
                 return
+
+            if last_locked["total_rows"] > 1:
+                raise MultipleResultsFound()
 
             # Calculate the duration the task was locked for
             duration_task_locked = (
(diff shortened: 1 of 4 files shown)


ramyaragupathy:
Test by @manjitapandey in Naxa' dev setup. Merging it to develop branch for further testing

PR #7222, opened 2026-04-13
Security: Fix DoS risks, patch XXE vulnerability, and add security he…

walle5eva:
This pull request addresses several security vulnerabilities identified during a security audit using Bandit (SAST) and OWASP ZAP (DAST).

1. Mitigated XML External Entity (XXE) Vulnerabilities (Bandit B314/B405)

Replaced the standard, insecure xml.etree.ElementTree parser with the secure defusedxml equivalent in the mapping services and test helpers to prevent potential XML attacks.

Added defusedxml to requirements.txt.

2. Resolved Denial of Service (DoS) Risks (Bandit B113)

Added a standard timeout=30 parameter to multiple requests.get() and requests.post() calls across the backend (general.py, osm_service.py, project.py, etc.). This prevents the server from hanging indefinitely if external APIs (like GitHub or OSM) fail to respond.

3. Added Global Security Headers (ZAP)

Implemented a FastAPI middleware in the main application file to automatically append essential security headers to all responses.

Added X-Content-Type-Options: nosniff to prevent MIME-sniffing.

Added X-Frame-Options: SAMEORIGIN to mitigate clickjacking.

Added a basic Content-Security-Policy.

These changes were made as part of an open-source security contribution for a university computer security course.


prabinoid:
Thanks you @walle5eva for the work on this PR and for the security hardening here, really appreciate it. The timeout, headers, and XML-related changes are useful.
One important thing to note: because the Dockerfile/runtime setup does not seem to pick up new Python packages directly from requirements.txt, we likely need to add the dependency with uv add <package_name>. That updates both pyproject.toml and uv.lock, which seems necessary for the dependency to be installed correctly in the backend image. Without that, the backend can become unhealthy at runtime due to import errors.
Also, could we please remove the accidental desktop.ini file, and double-check the pyproject.toml change that excludes tests from Bandit? Since this is a security-focused PR, I think we should avoid reducing scanner coverage unless there is a strong reason.
Finally, for the defusedxml change, it would be great to point to the exact XML parsing path that is being hardened, just to make sure the XXE fix is addressing a real untrusted-XML sink. Thanks again!

Issue #6704, opened 2025-01-21
Allow Projects to be 'Featured' - for easy access on 'Explore Projects' page


SColchester:
**Problems**

- The first page of the Explore Projects page of the Tasking Manager is [dominated by old irrelevant projects](https://github.com/hotosm/tasking-manager/issues/5138). This is not true of 'Urgent' projects (because there are so few), but there are hundreds of 'High' priority projects and they are sorted by default from oldest first. This means that all the 'High' priority projects on the first page of the Explore Projects page are at least a year old at the moment and they are the most likely projects in this priority level to have data needs which have since moved on (i.e. contributing to these projects is likely to have relatively low impact).
- Over time, as OSM has become more complete it has become harder and harder for beginner level users of Tasking Manager to find projects which are appropriate for them to contribute to (simple buildings only projects are much rarer). In the past, the first page of Explore Projects had several projects appropriate for beginners, nowadays this is rarely the case and likely makes a difficult/unattractive onboarding process for new mappers and may lead to new mappers making bad contributions to projects that are too complex for them.

**Solution**

Enable projects with any priority level to be manually marked by a Tasking Manager admin (via [API](https://tasks.hotosm.org/api-docs/)) as a 'Featured project'. Any featured project would then appear directly after the 'Urgent' priority projects on the first page of the Explore Projects page in the default view. This would be used to highlight very relevant (impactful) non-disaster (disaster projects are set to 'Urgent') related projects that are well designed with good instructions and descriptions. This could also be used ensure there is at least one appropriate and impactful beginner level project on the first page of Explore Projects for easy access. Several projects could be featured at the same time, see mock-up below:

![Image](https://github.com/user-attachments/assets/28c3aa1c-7664-4da4-b22b-7d58e5645474)

Candidates for featured projects would be decided on by members of the Activation Working Group (which includes staff and volunteers), to ensure fairness, for wider perspectives and to avoid HOT/Hub projects being consistently featured by HOT staff. 

By featuring high quality, impactful projects this would incentivise the organizations and communities on Tasking Manager to ensure that the projects they are creating are well designed --> If they create a high quality project they have a higher chance the project being featured and given extra prominence for faster completion. 

To summarize, the default ordering of projects on the Explore Projects would then be:

`Urgent > Featured > High > Medium > Low`

**Alternatives**

- Change project sorting to sort by most recently created projects first in each priority category 
[...]

PR #7230, opened 2026-04-25
Add featured projects to default sort on Explore Projects page

wtomeo46:
## What

Implements issue #6704 by adding "featured" projects to the default sort order on the Explore Projects page. When no explicit `order_by` parameter is supplied, projects are now sorted in the following order: 
**Urgent → Featured → High → Medium → Low**

## Why

As described in #6704, the Explore Projects page currently defaults to sorting by priority, with older projects appearing first within each priority group. This means the front page is often dominated by outdated projects and beginner-friendly projects are hard to find. The "featured" flag had no effect on the default sort. This PR makes featured projects discoverable in the default view by placing them in their own prioritized tier.

## How

Two small changes in `backend/services/project_search_service.py`:

1. **SELECT**: Added `p.featured` to the project search query to be available for ordering.
2. **ORDER BY**: Replaced the previously empty default `order_by_clause` with a `CASE` expression that orders the five tiers. The existing logic that overrides `order_by_clause` when `search_dto.order_by` is provided is preserved.

## Testing

Added `test_default_sort_order_places_featured_between_urgent_and_high` in `tests/api/unit/services/test_project_search_service.py`. The test initializes 4 projects covering each tier (urgent, featured-high, plain-high, medium) and asserts that search service returns them in the expected order when no `order_by` is supplied.

To run:
'uv run python -m pytest tests/api/unit/services/test_project_search_service.py -v'

All 4 tests in that file pass locally.

### DIFF
--- a/backend/services/project_search_service.py
+++ b/backend/services/project_search_service.py
@@ -120,7 +120,8 @@ class ProjectSearchService:
                 u.name AS author_name,
                 u.username AS author_username,
                 o.name AS organisation_name,
-                o.logo AS organisation_logo
+                o.logo AS organisation_logo,
+                p.featured
             FROM projects p
             LEFT JOIN organisations o ON o.id = p.organisation_id
             LEFT JOIN users u ON u.id = p.author_id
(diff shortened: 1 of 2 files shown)


SColchester:
Great, how would the featured attribute be added to projects? Just by Tasking Manager admins?

wtomeo46:
Yes, as already set, only admin users can set the featured attribute through the project edit page as access is gated by UserRole.ADMIN.

SColchester:
Fantastic @wtomeo46! Great

PR #7283, opened 2026-07-02
Fix duplicate react-intl message ids breaking yarn build-locales

### LINKED ISSUE #7284: [BUG] "yarn build-locales" fails on develop: duplicate react-intl message ids
## Problem

`yarn build-locales` (frontend/) fails on `develop`:

```
Error: Duplicate message id: management.button.cancel
```

`combine-react-intl-messages` throws on any message id defined in more than one `src/**/messages.js` file ‚Äî even when the texts are identical. As a result `frontend/src/locales/en.json` cannot be regenerated. It was last regenerated in June 2025, so no UI string added since then has reached Transifex.

The reported id is only the first duplicate the tool encounters. A full audit finds 11 duplicated ids:

**Identical texts** (defined in `teamsAndOrgs/messages.js` since 2019, re-declared in `levels/messages.js` and `badges/messages.js` in 2025): `management.button.cancel`, `management.button.save`, `management.fields.name`, `management.fields.description`, `management.fields.managers.add`, `management.fields.organisation.image`, `management.link.manage`, `management.badges`, `management.levels`

**Conflicting texts** (same id, different English strings):
- `management.projects.create.review.database` + `.osm` + `.sandbox` ‚Äî `projectCreate/messages.js` ("Where should edits be saved?" / "OpenStreetMap (Live)" / "Training Sandbox (Practice)") vs `projectEdit/messages.js` ("Database" / "OSM" / "Sandbox")
- `project.detail.sandbox` ‚Äî `projectDetail/messages.js` ("Sandbox Mode") vs `taskSelection/messages.js` ("Sandbox(Practice)") vs `header/messages.js` (unused since #7278)

## Why it went unnoticed

CI never runs `build-locales` ‚Äî only the manual `make refresh-translatables` target invokes it, although the contributing guidelines require regenerating `en.json` with any string change.

jaroslav-kubicek:
Closes https://github.com/hotosm/tasking-manager/issues/7284

## What

`yarn build-locales` fails on `develop` with `Error: Duplicate message id: management.button.cancel`, so `frontend/src/locales/en.json` could not be regenerated since June 2025 — no string added since then has reached Transifex. The reported id is only the first duplicate the tool hits; a full audit found 11 duplicated ids. This PR deduplicates all of them, regenerates `en.json`, and adds a CI step so this class of breakage fails PRs in the future.

## Changes

**Identical duplicates (translations preserved).** `levels/messages.js` and `badges/messages.js` re-declared nine ids that `teamsAndOrgs/messages.js` has defined since 2019 (`management.button.cancel`, `.save`, `management.fields.name`, `.description`, `.managers.add`, `.organisation.image`, `management.link.manage`, `management.badges`, `management.levels`). They now import and re-export the canonical definitions — same runtime keys, zero component changes, existing Transifex ids and translations untouched.

**Conflicting duplicates (new ids, no translation loss).** Two clusters shared an id with *different* English texts, so one text silently won depending on extraction order:

- `projectEdit/messages.js` vs `projectCreate/messages.js`: `management.projects.create.review.database` + `.osm` + `.sandbox` ("Database"/"OSM"/"Sandbox" vs "Where should edits be saved?"/"OpenStreetMap (Live)"/"Training Sandbox (Practice)"). The projectEdit entries now use `management.projects.edit.database*`.
- `taskSelection/messages.js` vs `projectDetail/messages.js`: `project.detail.sandbox` ("Sandbox(Practice)" vs "Sandbox Mode"). The taskSelection entry now uses `project.detail.sandbox.info`.

All four renamed ids were added after the last successful en.json regeneration, so they never reached Transifex — no translations are lost. The unused `sandbox` message in `header/messages.js` (dead since #7278 moved the badge to the project detail header) is removed.

**Regenerated `en.json`.** ~140 additions are a year of accumulated new messages; the few removals/rewordings all correspond to messages actually deleted or changed in source since June 2025. After merge, a maintainer needs to run `tx push -s` so these strings finally reach translators.

**CI guard.** `pr_test_frontend.yml` now runs `yarn build-locales` after install; the tool exits non-zero on any duplicate id, so future duplicates fail the PR instead of breaking locale regeneration unnoticed. (The contributing guidelines already require regenerating en.json with string changes — this was just unenforced.)


sumitdahal7:
Nice fix for the yarn build-locales failure  reusing the canonical descriptors, renaming the conflicts, and regenerating the catalog all makes sense.

Two things before I approve:
1. CI still won't catch a stale catalog, it only fails if generation errors, not if someone forgets to commit the regenerated en.json. Mind adding a diff check after the build step?
```
yarn build-locales
git diff --exit-code -- src/locales/en.json
```
2. Minor: Description says 11 duplicate IDs, but the list adds up to 13. Not blocking, just fix if you get a chance.

Also, no issue at all with using AI to help write this, that's normal. Just noticed a couple commits have a Claude co-author trailer; could you drop that? Just want authorship on the repo to stay clean regardless of what tools were used.

PR #7333, opened 2026-09-15
Fix permanent white screen after an interrupted OSM login

sumitdahal7:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [ ] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue
N/A

## Describe this PR

If an OSM login gets interrupted, the user closes the tab or the network stalls while the callback is still in flight and the app ends up permanently blank on every later visit, with no way out short of clearing site data.

The chain: `/authorized/` lands back with no credentials, we call `setAuthDetails(null, null, null)` anyway, `btoa(null)` quietly stores the string `'bnVsbA=='` as a token, and `setUserDetails` flips the global loader on for a request for user `"null"`. There's a `finally` that clears the loader, but if the page dies before it runs, redux-persist has already written `loader.isLoading: true` to disk. From then on `App.js` renders `<Preloader />` and nothing else, so the flag can never be cleared.

Two halves:
- **`views/authorized.js`** — stop creating new stuck users. `authComplete` now returns its promise and catches a rejection (including a state mismatch) by sending the user to `/login` instead of leaving them on "Redirecting …". And if the callback comes back with no `username`/`session_token`, bail to `/login` before dispatching anything, so no junk token is ever stored. Also fixes a stray `class` → `className` on the way through.
- **`store/index.js`** — stop the stuck flag from persisting. Swapped the persist `blacklist` for a `whitelist` of `preferences` and `auth`, so transient state stays off disk by default instead of every new reducer being persisted until someone remembers to opt out. Users already carrying the bad flag see one more blank load, then the first write drops the slice and they're back.

No new environment variables. No new strings (nothing in `messages.js`), so no Transifex sync needed.

- **A redux-persist migration** (`version: 1` + `createMigrate`) to drop the stale `loader` slice on first load, so stuck users recover one reload sooner. Rejected: the obvious implementation rebuilds a fixed object (`({ _persist, preferences, auth }) => ({ _persist, preferences, auth })`), which sets `auth: undefined` for anyone whose stored blob lacks that key. `autoMergeLevel1` hard-sets that over the reducer's initial state and every `state.auth.token` selector throws a white screen of a different kind. Recovering a load earlier isn't worth owning a version contract.
- **Just adding `loader` to the blacklist.** Fixes this bug but leaves the default pointing the wrong way; the next transient reducer someone adds gets persisted again. The whitelist makes the safe thing the default.

## Review Guide

Reproducing needs the request to stall, otherwise the `finally` clears the flag and it looks fine:

1. Log out. In DevTools → Network, add a request-blocking pattern for `*/users/*` (Chrome: right-click a request → Block request URL, or the Network request blocking panel).
2. Visit `/authorized/` with no query params.
3. In Application → Local Storage, look at `persist:root` — on `develop` it now contains `"loader":"{\"isLoading\":true}"`.
4. Remove the block and reload. On `develop`: blank page with a spinner, on every reload and every URL. On this branch: one blank load, then the next reload boots normally and `persist:root` is down to `_persist`, `auth`, `preferences`.

Also worth checking:

- `/authorized/?code=bad&state=x` and plain `/authorized/` both land on `/login` with nothing in `localStorage.token`.
- A normal login still works and `auth` + `preferences` still survive a reload.
- Five slices stop being persisted: `loader`, `notifications`, `lockedTasks`, `project`, `user`. I went through the consumers `project` and `user` are dead (their actions are never dispatched), `lockedTasks` is rebuilt from the `activities` response on mount in `taskSelection/index.js`, and `notifications.unreadCount` just starts `null` until the first fetch instead of showing a cached number. Sandbox tokens live in `auth`, so they're unaffected.


frontend/src/store/index.js
```diff
@@ -7,10 +7,25 @@ import storage from 'redux-persist/lib/storage';
 import { setItem } from '../utils/safe_storage';
 import reducers from './reducers';
 
+// Only genuinely durable state belongs on disk: user preferences and the
+// session. Everything else is either transient UI state or derived from the
+// server, and must be recomputed on boot.
+//
```
ramyaragupathy: Please trim down here

frontend/src/store/index.js
```diff
@@ -7,10 +7,25 @@ import storage from 'redux-persist/lib/storage';
...
+// on a permanent spinner after an interrupted login.
 const persistConfig = {
   key: 'root',
   storage,
-  blacklist: ['editor', 'orgBarVisibility'],
+  // Applied on write only -- redux-persist reads back whatever is already on
+  // disk. So a user carrying a stale `loader.isLoading: true` still sees one
+  // blank load, and is then rescued by the first write here dropping the slice.
+  // No migration: recovering them a load earlier is not worth owning a version
+  // contract, and rebuilding a fixed object in one would set `auth: undefined`
```
ramyaragupathy: Same, let's keep the comment short and maintanable.

ramyaragupathy:
@sumitdahal7 - thanks for looking into the issue. Wanted to understand the scope here: the actual bug was just one thing (loader) getting saved to the browser when it shouldn't have been. I understand this PR changes the approcah,  so almost nothing is saved except login/session info, going forward. 

Was that intentional, so a future feature doesn't accidentally cause the same kind of bug? If so that makes sense to me, just want to confirm.

Two things on the user-facing side:

* Locked tasks: I understand this gets refreshed from the server as soon as the page loads anyway, so a user shouldn't actually lose their locked task. Can you confirm that's right?
* Notifications: the unread count will briefly show zero right after a page load instead of showing the last known number, until it checks the server again a moment later. Just flagging so it's a known tradeoff.

ramyaragupathy (changes requested):

sumitdahal7:
Yes, the whitelist was intentional. With a blacklist, every new reducer gets persisted by default until someone remembers to opt it out, which is exactly how `loader` ended up on disk. Flipping it to a whitelist makes the safe thing the default, so a future feature can't reintroduce this class of bug by accident.

To be concrete about scope: we still persist `auth` (token, userDetails, session, osm, organisations, teams, sandbox token) and `preferences` (locale, mapShown, action). What stops being persisted is `loader`, `notifications`, `lockedTasks`, `project` and `user`. Of those, `project` and `user` are dead slices (nothing dispatches their actions), and the rest are either transient or server-derived.

- On locked tasks: correct, nothing is lost. The lock lives on the server and the redux slice is only a cache. It gets rebuilt from the project `activities` response on mount in `taskSelection/index.js`, and the `/map` and `/validate` routes don't read redux at all, they fetch `users/queries/tasks/locked/details/` through react-query.
- On notifications: good flag, and it's worth noting it's narrower than it looks. The red dot on the bell never used the persisted count, it comes from `useUnreadNotificationsCountQuery`. The only reader of `notifications.unreadCount` is the "N unread" link inside the bell popout. That popout is mounted but hidden with a `dn` class, so the empty count isn't visible, and the bell's query fires on mount and dispatches `SET_UNREAD_COUNT` well before a user can click to open it. So there shouldn't be a visible flash of zero. @ramyaragupathy

PR #7334, opened 2026-09-18
fix: load the project's imagery in the embedded iD editor on first load

### LINKED ISSUE #7255: [BUG] Wrong default imagery in iD editor
**Describe the bug**
During mapathon some iD editor users reported that the imagery changed for them. I just tried it myself and looks like iD editor first load Bing imagery and then automatically switch to the one that is set by the project. It doesn't happen always, when I try to lock the same task now it correctly load immediately the proper imagery but to my colleague it happened at least twice on the same project.

**To Reproduce**
1. Go to project [50430](https://tasks.hotosm.org/projects/50430/tasks/)
2. Lock a task for mapping with iD editor
3. See that it loads Bing imagery even tho it is set to map with Esri in the project

I don't have enough informations but looks like **after you click** it switch to the correct imagery

**Expected behavior**
Load the correct imagery according to project by default.

**Screenshots**

Image

 - OS Windows 11 Version 24H2 (Build 26100.8390)
 - Browser Vivaldi version 8.0.4033.34 (Official Build) (64-bit)
 - TM Version v5.5.4

sumitdahal7:

- [ ] 🍕 Feature
- [x] 🐛 Bug Fix
- [ ] 📝 Documentation
- [ ] 🧑‍💻 Refactor
- [x] ✅ Test
- [ ] 🤖 Build or CI

## Related Issue

Fixes #7255

## Describe this PR

On the first load of the embedded iD editor, the configured project imagery could be unavailable while iD was selecting its background. This caused iD to fall back to Bing until an unrelated interaction triggered a React render.

This PR sets the `background` hash parameter before initializing iD so the configured imagery is selected immediately. It also:

- Supports built-in and custom imagery sources
- Defaults to Bing when no project imagery is configured
- Replaces imagery left over from a previously opened project
- Preserves other existing hash parameters
- Adds unit tests for the new behavior

This change does not introduce environment variables, database changes, or API changes.

## Screenshots

Not applicable. This fixes background selection behavior without changing the interface.

Applying the imagery after iD initialization was considered. This is unreliable on a cold start because the imagery index loads asynchronously and does not trigger a React render when it becomes available.

Setting the background in the hash before initialization follows iD's existing background selection behavior.

## Review Guide

1. Open a mapping task from a project with custom imagery configured.
2. Launch the embedded iD editor.
3. Verify that the project imagery is selected immediately without requiring another interaction.
4. Repeat with a built-in imagery source.
5. Open a project without configured imagery and verify that Bing is selected.
6. Confirm that switching between projects does not retain the previous project's imagery.

Automated checks completed:

- `CI=true yarn craco test --env=jsdom --watchAll=false src/utils/tests/idEditorContext.test.js`
  - 9 tests passed
- `yarn eslint src/components/editor.js src/utils/idEditorContext.js src/utils/tests/idEditorContext.test.js`
  - Passed

The full frontend test suite and manual browser checks were not run.

## AI Tool Usage

How much of this PR was AI-assisted? (check one)

- [ ] **0** - No AI at any point
- [ ] **1** - AI helped me think, but I wrote all the code myself
- [x] **2** - I planned the change and decided the approach; AI helped write the code
- [ ] **3** - AI planned and wrote it; I checked and approved each step as it went
- [ ] **4** - I set the AI going and left it to it; I reviewed the finished result
- [ ] **5** - I set the AI going and left it to it; nobody has read the result - no review, or AI review only

**If level 1 or above:**

- Tool(s) used: OpenAI Codex
- What was generated: PR description and, if applicable, implementation or test suggestions
- What you reviewed and changed: Reviewed the generated content against the implementation, tests, and observed behavior

## [optional] What gif best describes this PR or how it makes you feel?

Not applicable.


frontend/src/utils/idEditorContext.js
```diff
@@ -59,3 +59,29 @@ export function resolveIdEditorContext(existingContext, editorType, buildContext
...
+// only sees the right imagery after some unrelated click happens to re-render
+// (#7255).
+export function setBackgroundHashParam(imagery) {
+  const params = new URLSearchParams(window.location.hash.replace(/^#/, ''));
+  // `custom:<template>` is how iD spells a custom source in the hash. With no
+  // project imagery, ask for Bing explicitly so a background left over from a
+  // previous project doesn't carry across.
+  params.set(
+    'background',
+    !imagery ? 'Bing' : imagery.startsWith('http') ? `custom:${imagery}` : imagery,
```
ramyaragupathy: Can you please fix the sonarcloud warnings?

ramyaragupathy: I think splitting it into an if/else or a small named variable would be a one-line cleanup. Please add your thoughts.

frontend/src/utils/idEditorContext.js
```diff
@@ -59,3 +59,29 @@ export function resolveIdEditorContext(existingContext, editorType, buildContext
   context.__idEditorType = editorType;
   return context;
 }
+
+// iD chooses its background exactly once, after its UI has finished loading,
```
ramyaragupathy: One general note: a few of the comments in this PR and it has become a pattern in the codebase. Could you please trim comments to a line or two? Easier to scan and maintain as the code changes.

ramyaragupathy (changes requested):
Edge case to double check:  If imagery changes while the editor is already open and not just on first load, does that reset wipe out any in-progress edits? 

Worth a manual test on dev setup. let's ensure we switch between different imagery while mid-edit and confirm nothing's lost.

sumitdahal7:
> Edge case to double check: If imagery changes while the editor is already open and not just on first load, does that reset wipe out any in-progress edits?
> 
> Worth a manual test on dev setup. let's ensure we switch between different imagery while mid-edit and confirm nothing's lost.




 Checked this. You're right that the reset is destructive: the warm branch calls `iDContext.reset()`, which resets iD's history and would drop unsaved edits.

 But `imagery` can't change while the editor is open today. It comes from `useProjectSummaryQuery`, which sets no `staleTime` or `refetchInterval`, `refetchOnWindowFocus` is off globally, and `['project-summary', projectId]` is never invalidated anywhere in the frontend. It's fetched once per mount and stays fixed for the session, so nothing was at risk in practice.

 I've removed the dependency anyway, by reading `imagery` through a ref. It never did anything useful, since `setBackgroundHashParam` only runs on the cold branch, so a reset was its only possible effect. If someone later adds a refetch or an invalidation for the project summary, this would have started dropping edits silently.

 One note on the manual test: switching background inside iD doesn't change the `imagery` prop, which comes from the project summary API. So that test won't exercise the path. Forcing the prop to change from outside iD is the only way to trigger it, and with the ref in place that no longer resets anything.

frontend/src/utils/idEditorContext.js
sumitdahal7: Agreed, made the changes. Split into an if/else with a named variable, which also clears the SonarCloud issue since it was the same finding.

frontend/src/utils/idEditorContext.js
sumitdahal7: Agreed, made the changes. Trimmed the comments in this PR to a line or two.

